PR #10, opened 2017-10-01
Jump Search through Java

navendu29:


code/search/interpolation_search/Interpolation.java
```diff
@@ -0,0 +1,45 @@
+package arrays;
```
AdiChat: This line must be removed as an end user may not be using the package name and thus, will result in error.

Removing it makes the code perfect.

code/search/jump_search/JumpSearch.java
```diff
@@ -0,0 +1,55 @@
+package arrays;
```
AdiChat: This line must be removed as an end user may not be using the package name and thus, will result in error.

Removing it makes the code perfect.

navendu29:
Removed the unwanted line. You can merge the PR now. Thanks

AdiChat:
Looks good 👍 

PR #22, opened 2017-10-01
Added Bubble Sort in Java

n0zg:


code/sorting/bubble_sort/BubbleSort.java
```diff
@@ -0,0 +1,31 @@
+package bubblesort;
```
AdiChat: Remove this line. This will cause error for many as they may not have the same package.

n0zg:
Done! I think, This is the first time I've modified a file in a pull request.

AdiChat:
Looks good 👍 

PR #24, opened 2017-10-01
Added file for bubble sorting in c language

jainkeval:
This code is for bubble sorting an array.
It first asks for size of array, then it asks for it's elements.
And then prints the sorted array.


AdiChat:
Looks good 👍 

jainkeval:
Yeah ,will be glad to work on further issues.

Issue #11, opened 2017-10-01
Add bubble sort


AdiChat:
Add the code for `bubble sort` in any language ( `C`, `C++`, `Java`, `Go`, `Python` or any other)

The code should be placed at [code/sorting/bubble sort](https://github.com/OpenGenus/cosmos/tree/master/code/sorting/bubble_sort)

**Note**: multiple contributors can work on this issue as it has multiple parts (languages)

Your pull request will be reviewed by maintainers instantly.

For contribution guidelines, see [this](https://github.com/OpenGenus/cosmos/wiki/contribute)

If you need any help, let us know.

gautam2705:
I want to :)

AdiChat:
Sure. Go ahead. 👍

jainkeval:
I will work on it for c language

AdiChat:
Sure. 👍

shashank-sharma:
I would like to work on python if possible

n0zg:
I will do the java version

ahhmarr:
#18 
for JavaScript

choudhary001:
Hi I am trying to learn Rust language. Will that be accepted if I submit Rust versions of these algorithms?
Thanks!

n0zg:
I've already added the code for Bubble Sort in Java :)

AdiChat:
Yes, @choudhary001 go for it.

Sanyam96:
I want to ;)

AdiChat:
Go ahead 👍

charul97:
I want to do it for C++

AdiChat:
@charul97 Bubble sort in C++ has been pushed. Kindly take a look at other requirements or make a pull request to make improvements to the existing code. ([see](https://github.com/OpenGenus/cosmos/blob/master/code/sorting/bubble_sort/bubble_sort.cpp))

endiliey:
#35 

I created a pull request for an improvement to the existing code. Instead of doing it in procedural way, I've implemented it such that it's implemented as a function. Using it is as simple as `bubble_sort(our_vector);`

trshimpi:
i wanna add the code in c++

leodevo:
#106

I made this pullrequest to also improve the existing code. 
bubble_sort.cpp was not compiling, and I then added another commit to improve the coding style

Guilhermeslucas:
I'll right it on golang.

BenjaminFarley:
[...]

PR #35, opened 2017-10-01
Refactor bubblesort implementation

endiliey:
Previous implementation was done in a very procedural way. Implementing it as a function would make it easy to be used as a personal algorithm library. If we want to copy the bubble sort implementation, we just need to copy line 6 to line 21 to our c++ files and call bubble_sort(our_vector);

I believe this is called better practice


endiliey:
Issue: #11

endiliey:
#11

endiliey:
#11

AdiChat:
Looks good 👍 

PR #129, opened 2017-10-02
Added bit division in Go

druzmieres:
I couldn't find a better way to handle the division by zero error, but it works.


druzmieres:
Added binomial coefficient in Go too

AdiChat:
Looks good 👍 

PR #137, opened 2017-10-02
add NQueen implementation for backtracking

hwchiu:
Add a example about how to use backtracking so solve N-Queen problem and speed up via bit manipulation


AdiChat:
Looks good 👍 

PR #182, opened 2017-10-02
Add Bit manipulation Division in C++

Eleanor1702:
Added Bit manipulation devision in C++.


AdiChat:
Bit division exists in `C`, Kindly add code for other task or refactor the existing code ([see](https://github.com/OpenGenus/cosmos/blob/master/code/bit-manipulation/bitDivision.c))

PR #197, opened 2017-10-02
Added Stack in Python3

AlexanderDavid:
Added a stack class implementation with arrays in python3, has both the stack class and a main function with some functionality.


AdiChat:
Kindly remove the submodule commit `Submodule cosmos added at f342e4`

It will be good to add then. 👍

Issue #96, opened 2017-10-01
Add Sieve of Eratosthenes


vstrimaitis:
Add the code for the sieve of Eratosthenes algorithm.
The code should be placed under /code/mathematical-algorithms/sieve_of_eratosthenes.

vstrimaitis:
Added the first implementation in C++: #91

ArunaMaurya221B:
Can I work on the java version for this??

AdiChat:
Sure. Go ahead.

[Contribution guidelines](https://github.com/OpenGenus/cosmos/wiki/contribute)

philipyoo:
Added a Javascript implementation here: https://github.com/OpenGenus/cosmos/pull/117

Jecoms:
I'll add a C# version.

druzmieres:
I can work on the Go version of this.

druzmieres:
#124

ArunaMaurya221B:
added the java version #153

setupminimal:
I'd like to add a Haskell version when I get home tonight.

AdiChat:
Sure @setupminimal 👍

setupminimal:
Added in #193.

qoo2p5:
Added a C++ implementation of a _linear_ sieve: #414.

abhijeet010304:
I would like to work on it in C++  and pyhton please assign it to me as a part of GSSoC 20

PR #213, opened 2017-10-02
Added an implementation of the Sieve of Eratosthenes in Haskell.

setupminimal:
I accidentally put too much in my previous pull request - but I think I've got the hang of it now. This is for #96.


AdiChat:
Looks good 👍 

PR #218, opened 2017-10-02
Add delannoy number algorithm

danielvieiravega:
Hi!
Adding the delannoy number algorithm written in c++. 
More information about it is available at wikipedia: https://en.m.wikipedia.org/wiki/Delannoy_number
Thanks!


AdiChat:
Looks good 👍 

PR #227, opened 2017-10-02
Added python stack && c++ queue

AlexanderDavid:
I added a python stack and c++ queue with arrays


AdiChat:
Looks good 👍 

PR #236, opened 2017-10-03
Java SingleLinkedList: Java convention cleanup and Generics

Liamjen:
The SingleLinkedList java example was limited to ints, so I added the ability to use Generics.

Other methods (like the display function) didn't really follow Java conventions, so I added a toString method for displaying. The InsertNth method was also cleaned up.


AdiChat:
Looks good 👍 

PR #238, opened 2017-10-03
Java Queues: Allow Java Generics

Liamjen:
I also made remove() throw an exception if there are no elements as opposed to printing a message.


AdiChat:
Looks good 👍 

PR #240, opened 2017-10-03
Add shuffle_an_array in C++

LRFLEW:
I saw this folder was empty, and since it's a fairly simple algorithm, I decided to just push an implementation. I made this work in the style of `std::shuffle`, so it uses iterators and C++11 random number generators.


AdiChat:
Looks good 👍 

PR #309, opened 2017-10-03
AI TSP added with test cases

AkshGautam:
TSP using simulated annealing and Tabu search added with test cases


AdiChat:
Looks great 🎉  
We would love to see such great contributions 👍 

Issue #300, opened 2017-10-03
Add Catalan Number


ZoroarkDarkrai:
I'll do it in c++ again :D.

manan904:
I'll do it in Python.

mkfeuhrer:
I'll do it in C

bamsarts:
i'll do it in java

manan904:
Should I create a new folder?

Ilozuluchris:
Can I do this in Scala

mkfeuhrer:
#307

AdiChat:
Sure @Ilozuluchris Go ahead 👍

ravilkashyap:
Can I do it in C++?

AdiChat:
@ravilkashyap We have a pull request for catalan number in `C++`. Kindly do it in some other language or try out some other task 👍

bhaveshAn:
@AdiChat please let me know why the issue is still not closed.

AdiChat:
@bhaveshAn We can add code in other languages as well 👍

rattandeep1998:
I've added Catalan Number in Ruby. @AdiChat

abhijeet010304:
@arnavb I can contribute to it in C++ please assign it to me

arnavb:
Already [here](https://github.com/OpenGenus/cosmos/tree/master/code/mathematical_algorithms/src/catalan_number) @abhijeet010304

PR #324, opened 2017-10-03
Implementation of catlan number in scala

Ilozuluchris:
Added a function that gets Catlan number in Scala #300


AdiChat:
Looks good 👍 

PR #325, opened 2017-10-03
Queue.c: Fixes several GCC warnings, typos and style issues.

dysoco:
Good evening.

I checked the code for the Queue implementation in C and it had a lot of GCC warnings when compiling, as well as some obvious C bad practices (no specified return type for functions, was missing an include for <stdlib.h>, etc.) apart from a minor typo on the actual program.

This commit attempts to fix those issues in order to provide a better example (imho) to learners; I've however maintained the original algorithm implementation (since I didn't write this) and the program runs exactly the same.

There could be some arguments made about replacing scanf calls with fgets (for security purposes) and some other changes but I've decided to not work on this further as to adhere to the KISS principle.

Anyways this is my first pull request in years so I hope I have done everything right; I'm not sure if I'm allowed to work on other people's implementations but I sent the PR regardless.

Cheers!

**p.s.** I accidentally added another commit to this PR looking on how to revert it.


AdiChat:
Looks good 👍 

Looks like the commit is fine now 👍 

Issue #335, opened 2017-10-03
Add Code for 2 unique numbers


sirjan13:
Find 2 unique Numbers in an array where every number occours twice except 2 numbers.find those two Numbers using bit manupilation

Lawih:
Does these numbers have a range?

PR #339, opened 2017-10-03
Added the code for 2 Unique Numbers in an Array

sirjan13:
Given an array of 2*n+2 elements where every elemnt occours in pairs accept 2 element.Task is to find those two elements efficiently using Bit masking.
Added C++ Code Resolving #335


AdiChat:
Looks good 👍 

Issue #331, opened 2017-10-03
Add documentation in README.md inside each algorithm.


dysoco:
So I noticed there's a `README.md` file inside each folder that actually provides no useful information.

What about we change each `README.md` file, for example [cosmos/code/search/ternary_search/README.md](https://github.com/OpenGenus/cosmos/blob/master/code/search/ternary_search/README.md) to explain a bit about the Ternary Search algorithm?

AdiChat:
This is important and needs to be done for each folder level 👍 

Go ahead and add some details 😄

leodevo:
#360

Added it for the Bubble sort algorithm

levifranca:
Can I add for the Minimum Coin algorithm?

AdiChat:
Sure. Go for it 👍

bsphair:
Can I add a README.md for the insertion sort?

AdiChat:
Sure @bsphair Go ahead 👍

levifranca:
Hi @AdiChat, sorry for the delay, I had some setbacks that prevented me from update the README before.

Here is the PR with the change #798 

Thank you

AdiChat:
Hello @levifranca 😄 No problem. Hope you are good now 👍

leodevo:
#1717 

Added it for the quicksort algorithm

PR #345, opened 2017-10-03
Add linear search info to README.md

fbontin:
Added information about linear search as sought after in #331. I only added a small description and information about performance.


AdiChat:
Looks good 👍 

PR #349, opened 2017-10-03
#1 height of binary tree in java

codefeeder:
Implementation of Binary Tree in java with basic functions like size, isempty , display etc. and a function to calculate the height of the tree.

### DIFF
--- a/code/data_structures/Tree/heightofbt1.java
+++ b/code/data_structures/Tree/heightofbt1.java
@@ -1,0 +1,110 @@
+package practiceds;
+
+import java.util.LinkedList;
+import java.util.Queue;
+import java.util.Scanner;
+import java.util.Stack;
+
+public class binarytree {
+private class node {
+	int data;
+	node left;
+	node right;
+	
+	public node(int data,node right,node left){
+		this.data=data;
+		this.left = left;
+		this.right = right;	
+	}
+}
+private node root;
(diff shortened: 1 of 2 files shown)


code/heightofbt.java
```diff
@@ -0,0 +1,110 @@
+package practiceds;
```
AdiChat: This code shall be removed from the code directory 👍 

The above file is the same one 👍

code/data_structures/Tree/heightofbt1.java
```diff
@@ -0,0 +1,110 @@
+package practiceds;
```
AdiChat: Kindly remove this package as another user may not have the same package name 👍

code/unclassified/fibonacci.cpp
```diff
@@ -0,0 +1,22 @@
+#include <bits/stdc++.h>
```
AdiChat: This code must be in [this code](https://github.com/OpenGenus/cosmos/tree/master/code/mathematical-algorithms/fibonacci_number)

Kindly make corresponding change 👍

arnavb:
@codefeeder There are many style issues with this code; check our C and Java style guides. Also, move this file to another location. If you want to avoid the hassle, I would just open a new PR and close this one.

Issue #342, opened 2017-10-03
Amicable numbers


ravilkashyap:
Find the sum of all the amicable numbers under 10000.

vaibhavsingh97:
Can i take up this issue?

AdiChat:
Sure. Go for it 👍

manan904:
I'll do it in python :)

PR #357, opened 2017-10-03
Amicable numbers.cpp

ravilkashyap:
Program to Calculate Amicable Numbers under 10000 and find sum.
An amicable pair of numbers consists of two different integers where the sum of the divisors of the first integer is equal to the second integer, and the sum of the divisors of the second integer is equal to the first integer.
Resolved issue  #342
@AdiChat kindly merge my PR  :)


AdiChat:
Looks good 👍 

PR #377, opened 2017-10-04
Java BinaryTree: Added generics and fixed class name to match file name

Liamjen:
I did not change any of the actual logic of the code. Methods that either returned or took a Node as a parameter were privatized, as the Node class shouldn't be touched by consumers of the code.


AdiChat:
Looks good 👍 

PR #379, opened 2017-10-04
Adding pascal_triangle.cpp and horner_polynomial_evaluation.cpp

tiagomontalvao:
Adding both Pascal Triangle and Horner Polynomial Evaluation.

Pascal Triangle runs in O(n^2), where n is the dimension of the triangle (linear time considering the matrix size).
Horner Polynomial Evaluation runs in O(n) time, where n is the degree of polynomial.


AdiChat:
Looks good 👍 

PR #404, opened 2017-10-04
Add SegTree with spaces

livace:
Now SegmentTree in Java is in Tree directory, but there is SegmentTree directory.

Added segment tree with spaces -- allows to find sum in O(log size) and uses O(n log size) memory


AdiChat:
Looks good 👍 

Issue #403, opened 2017-10-04
Print mirror Image(of the nodes) in the Binary tree (taking mirror as the root node )


shaikkamran:
Print mirror Image(of the nodes) in the Binary tree (taking mirror as the root node ) #392

aakash-sopho:
I will do it.

aakash-sopho:
Implemented in C++ and created PR

Arunraj97:
Is this thing open can I send a Pr

AdiChat:
Yes @Arunraj97 Go ahead 👍

rstalekar:
Implementing in C.

rstalekar:
Done writing in C, I am adding it in Tree subdirectory.

rstalekar:
I have pushed it in the git repository am creating a pull request

rstalekar:
I have pushed it in git repository, where can I see the my code being pushed it is mirror_image.c

AdiChat:
You can create the pull request from [here](https://github.com/OpenGenus/cosmos/compare/master...rstalekar:master) 👍 
It exists in your local workspace ([see](https://github.com/rstalekar/cosmos/commit/3d63e6a69c53b5c9721303c1b190a3dd5d60cdd5))

I will review and merge it soon 👍

PR #407, opened 2017-10-04
MirrorElementsBst.py

shaikkamran:
#403  resolved
input should be like this
10 8
1 2 R
1 3 L
2 4 R
2 5 L
3 6 R
3 7 L
5 8 R
5 9 L
7 10 R
2
5
3
6
1
10
9
4

First line of input is N and Q.
Next N-1 line consists of two integers and one character first of whose is parent node , second is child node and character "L" representing Left child and "R" representing right child.
Next Q lines represents qi.
Output:
For each qi printing  the mirror node if it exists else printing -1.


AdiChat:
Looks good 👍 

PR #413, opened 2017-10-04
Breadth first search implemented in Ruby

IAmBullsaw:
I built a node class which connects to other nodes, creating a graph.
I then implemented a function, bfs(), that performs a breadth first search for the nodes.


AdiChat:
Looks good 👍 

PR #421, opened 2017-10-04
Depth first search implemented in Ruby

IAmBullsaw:
I built a class GraphNode which could connect to other GraphNodes.
I implemented a depth first search, dfs(), which accepts a start node and traverses the graph from there.
I added a small test for a small graph (tree).

This is practically a small modification of the breadth first search, bfs.rb.


AdiChat:
Looks good 👍 

PR #422, opened 2017-10-04
Add persistent segTree

livace:
Persistent segTree allows to add and get sum in any version of tree


AdiChat:
Looks good 👍 

PR #432, opened 2017-10-04
add shortest_common_supersequence.cpp

aakash-sopho:
Shortest Common Supersequence using DP.


AdiChat:
Looks good 👍 

PR #434, opened 2017-10-04
Salesman c++

livace:
Maybe it should be in bits cos it's O(n * 2 ^ n) realisation


AdiChat:
Looks good 👍 

PR #451, opened 2017-10-04
Computes the height of the binary search tree

rstalekar:
Created node class and member functions to create a tree and then calculate the height of the tree and eventually cleanup of the tree.
One can even having a void* value instead of int value, but let me know if you think that can help.


AdiChat:
Looks good 👍 

PR #456, opened 2017-10-04
Added count of ways length n

livace:
Works in O( size * size * log length )


AdiChat:
Looks good 👍 

Issue #436, opened 2017-10-04
Check if number is a perfect number


imfaizz:

imfaizz:
[perfectnumber.pdf](https://github.com/OpenGenus/cosmos/files/1356393/perfectnumber.pdf)

AdiChat:
Sure. Go for it 👍

AlexanderDavid:
I'd like to do this in C++, I can do it in python too. Pretty much whichever language @manan904 doesn't do

devAmoghS:
@AdiChat Implementing this in Python

devAmoghS:
@imfaizz Where should i make the PR?

AdiChat:
Create a folder `perfect_number` under [mathematical_algorithms](https://github.com/OpenGenus/cosmos/tree/master/code/mathematical-algorithms) and add code to this folder

frazjp65:
We now have three directories for this and duplicates in Python and C++. 

https://github.com/OpenGenus/cosmos/tree/master/code/mathematical-algorithms/perfect_numbers

https://github.com/OpenGenus/cosmos/tree/master/code/mathematical-algorithms/perfect_number

https://github.com/OpenGenus/cosmos/tree/master/code/mathematical-algorithms/perfect-number

AdiChat:
Major issue. I will correct it immediately 👍 

To avoid this in future, folder name should be decided before pull requests if it does not exist 👍

frazjp65:
I was going by: 

> Create a folder perfect_number under mathematical_algorithms and add code to this folder

I made a PR for the python one, but I don't know which of the three C++ ones you want to keep.

AdiChat:
Resolved 👍 

I will set a naming convention to avoid such issues and make corresponding change soon 👍

manan904:
@AlexanderDavid What do you mean by manan doesn't do???

edwardrichtofen115:
``
[PerfectNumber.txt](https://github.com/OpenGenus/cosmos/files/1358108/PerfectNumber.txt)

C++ file for PerfectNumber

Jay9596:
Perfect number checker in Rust #544

AlexanderDavid:
@manan904 You said you'd do it in python /or/ C++. So I picked the one you wouldn't do it in

weirdGuy:
Will implement in ruby :)

AdiChat:
Sure. Go ahead 👍

[...]

PR #467, opened 2017-10-04
PerfectNo.py

shaikkamran:
#436 solved
This code helps in printing the perfect nos That is(No's having the sum of its factors=num itself )


AdiChat:
Looks good 👍 

PR #472, opened 2017-10-04
Perfect Number

ravilkashyap:
A perfect number is a positive integer that is equal to the sum of its proper positive divisors, that is, the sum of its positive divisors excluding the number itself.
#436 Solved in c++


AdiChat:
Looks good 👍 

PR #484, opened 2017-10-04
Added Haskell Implementation of Largest Contiguous Sum in Array.

setupminimal:
Implementation for #444.


AdiChat:
Looks good 👍 

Issue #477, opened 2017-10-04
Implement geometry


livace:
#473 - distance between points
#470 - area of triangle
#475 - area of polygon

drcabral:
#498

drcabral:
#498 Java solution for area of triangle, area of polygon and distance between points.

mor11235:
Can I implement this is C++?

AdiChat:
Sure. Go ahead 👍 

Be sure to check existing implementations 👍

PR #487, opened 2017-10-04
Python program for Miller_rabin_primality_test added

codeSG:
Added miller rabin primality test in python language.

### DIFF
--- a/code/mathematical-algorithms/primality_tests/Miller_Rabin_Primality_Test.py
+++ b/code/mathematical-algorithms/primality_tests/Miller_Rabin_Primality_Test.py
@@ -1,0 +1,76 @@
+"""
+    Program to implement Miller-rabin Primality Test
+    Part of cosmos by OpenGenus Foundation 
+"""
+import random
+
+#main method to check primality of a number
+def isprime(number):
+    iterations=4 #no. of iterations to run test
+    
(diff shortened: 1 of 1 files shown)


codeSG:
distance and area_of _triangle for  issue #477

FrankKair:
@codeSG You should pay attention to [PEP8](https://www.python.org/dev/peps/pep-0008/) when writing Python code for formatting.

For example, it states that:

```
Yes:

i = i + 1
submitted += 1
x = x*2 - 1
hypot2 = x*x + y*y
c = (a+b) * (a-b)

No:

i=i+1
submitted +=1
x = x * 2 - 1
hypot2 = x * x + y * y
c = (a + b) * (a - b)
```

So I think it would be better if we all maintained the coding styles determined by each language.

Other than that, Miller-Rabin is a very welcoming addition!

AdiChat:
@codeSG Kindly make corresponding changes as implied by PEP8 👍 

It will be good to merge then 👍

codeSG:
Thanks for the advice!!
I have updated the file accordingly.

arnavb:
@AdiChat please merge this after:

1. Moving the file to the right location
2. Making the files all lowercase separated by underscores.

PR #494, opened 2017-10-04
Minor cleanups on the c solution for fibonacci

Jaco-Pretorius:
The current solution generates a compiler warning (due to the result of the `for` loop not being used) and doesn't prompt the user for input.


AdiChat:
Looks good 👍 

PR #501, opened 2017-10-05
Java Stacks: Stack can now use Java generics and Stack2 was removed

Liamjen:
Stack now uses an ArrayList to hold data to allow use of Java generics, and class Stack2 was removed as it became redundant and had less methods.


AdiChat:
Looks good 👍 

PR #508, opened 2017-10-05
Implemented the mirror image of the tree in C, can be verified using …

rstalekar:
Implemented the mirror_image function to swap the left and right nodes from the root,, so that the entire left subtree and right subtree are swapped and to test print_mirror_image function is used to do a pre-order traversal and print.


AdiChat:
Looks good 👍 

PR #512, opened 2017-10-05
Create mediansort.cpp

codyalgorith:
very fast implementation in cpp , the first choice of competitive programmers


AdiChat:
Looks good 👍 

PR #532, opened 2017-10-05
Newton Raphson Method in C

imfaizz:
Newton-Raphson method using c


AdiChat:
Looks good 👍 

PR #541, opened 2017-10-05
C program for armstrong number

imfaizz:
c program that checks if a number is an Armstrong number


AdiChat:
Looks good 👍 

PR #543, opened 2017-10-05
Multiply using russian_peasant

imfaizz:
C/C++ program for multiplication using Russian peasant method


AdiChat:
Looks good 👍 

Issue #542, opened 2017-10-05
Add square root using babylonian method


phoenix157:

phoenix157:
Can i work on this?

AdiChat:
Sure. Go for it 👍

imfaizz:
can I work on this?

AdiChat:
Sure. Go for it 👍

imfaizz:
C program for sq root using Babylonian method
[babylonian.txt](https://github.com/OpenGenus/cosmos/files/1358878/babylonian.txt)

Inlaik:
Can I do this in Java?

AdiChat:
Sure. Go for it 👍

Arunraj97:
I'm doing this in C++ if it is fine

PR #548, opened 2017-10-05
C program for sq root using babylonian method

imfaizz:
#542 
[babylonian.txt](https://github.com/OpenGenus/cosmos/files/1358895/babylonian.txt)


AdiChat:
Looks good 👍 

PR #551, opened 2017-10-05
Shortest common supersequence in Python

amitXsarkar:
gives length of Shortest common supersequence


AdiChat:
Looks good 👍 

PR #553, opened 2017-10-05
Armstrong number in Go

Jay9596:
Implemented a Armstrong number generator from 0 to 999 in GO


AdiChat:
Looks good 👍 

PR #555, opened 2017-10-05
Correcting php function names to follow the coding standards.

isfonzar:
Corrected some function names for PHP, following php.net coding standards, which can be found here: http://git.php.net/?p=php-src.git;a=blob_plain;f=CODING_STANDARDS;hb=HEAD

> 1.  Function names for user-level functions should be enclosed with in
>     the PHP_FUNCTION() macro. **They should be in lowercase, with words
>     underscore delimited**, with care taken to minimize the letter count.


AdiChat:
Looks good 👍 

PR #573, opened 2017-10-05
Fix the memoized version to use the cache for n > 1

Jaco-Pretorius:
Currently the memoized version declares a new cache on every single step, so the cache is only used for values 0 and 1.  I've illustrated this with a RubyFiddle - http://rubyfiddle.com/riddles/60df0/1 - you can see the app never reports 'cache hit'.  With this updated version - which you can see with another RubyFiddle, http://rubyfiddle.com/riddles/9d409/2 - the cache is used properly.

It's also more semantic to use `Hash#key?` instead of `Hash#has_key?` in Ruby 1.9 and onwards.

### DIFF
--- a/code/mathematical-algorithms/fibonacci_number/fibonacci.rb
+++ b/code/mathematical-algorithms/fibonacci_number/fibonacci.rb
@@ -6,13 +6,11 @@ def fibonacci(n)
   end
 end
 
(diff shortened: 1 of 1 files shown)


AdiChat:
Looks great 🎉  

Hope you are enjoying your journey with **OpenGenus** 👍

Issue #580, opened 2017-10-05
Given inorder and preorder traversal of a tree gets its postorder traversal 


Arunraj97:

007akk:
hi @Arunraj97  i am doing this

rstalekar:
@007akk are you doing this in C?
If so I will do this in C++

arora-72:
@Arunraj97  algo implemented in java 
Please review
PR #602 
Thanks:)

rstalekar:
Hi Aditya,
Can you let me know how I can add myself in Contributors list? My name is only shown in members list

AdiChat:
You need to send a pull request 👍

ranjan-abhi:
@AdiChat i have implemented in c.

rstalekar:
I have already mentioned that I will be doing in C

PR #589, opened 2017-10-05
preorder and inorder tranversal to postorder traversal

rohit1576:
In response to #580 , this is the C++ code to print the postorder traversal of a binary tree when inorder and preorder traversals are given.


AdiChat:
Looks good 👍 

PR #593, opened 2017-10-05
Adding string anagram search in ruby

Jaco-Pretorius:
Adding ruby implementation for anagram search


AdiChat:
Looks great 🎉  

Hope you are enjoying your journey with **OpenGenus** 👍

Issue #603, opened 2017-10-05
Replace all ‘0’ with ‘5’ in an input Integer without using arrays for individual digits


imfaizz:

007akk:
@AdiChat i m doing this

akritikts:
#include<iostream.h>
int main()
{
int n;
cin>>n;
while(num!= 0)
{
a = num%10;
If(!a)
{
num = num*10 + 5;
}
else
{
num = num*10 + a;
}
return 0;
}

007akk:
where should i place the pull request

007akk:
#include <bits/stdc++.h>

using namespace std;

int main(){

	int n; cin >> n;

	int a = 1;

	while( n / a >= 1){

		a = a* 10;

		int temp = n % a - n % (a / 10);

		temp = temp / (a/10);

		int temp2;

		if(temp == 0){

			temp2 = 5 * (a/ 10);

			n = n - temp + temp2;
		}
		
	}

	cout << n << endl;
}

akritikts:
I am totally new to this hackathon, I didn't know where to answer that's
why I commented.

AdiChat:
This should go to the `unclassified` section 👍

PR #604, opened 2017-10-05
C program to replace all ‘0’ with ‘5’ in an input Integer withour using arrays

imfaizz:
issue #603


AdiChat:
Looks good 👍 

PR #607, opened 2017-10-05
NQueen Updated

sachin-kmr:
Earlier Implementation was NQueen using Bitmasking, so updated that to backtracking. Also it prints the valid positions of N-Queens on board.


AdiChat:
No need to delete the bitmap implementation. Kindly update your pull request accordingly 👍 

It will be good to merge then 👍

PR #609, opened 2017-10-05
Create Replace all ‘0’ with ‘5’ in an input Integer without using arr…

007akk:
…ays for individual digits


AdiChat:
Looks good 👍 

PR #617, opened 2017-10-05
Create lonelyinteger.cpp

TanviKumar:
Program to find the only number without a pair using bit manipulation. XOR operator used.


AdiChat:
`C++` implementation exists. Kindly take a look at other scopes of contribution or refactor existing code 👍

PR #654, opened 2017-10-05
Flip bits: C and Python implementations

sransara:
Following #621


AdiChat:
Looks good 👍 

PR #668, opened 2017-10-06
Update count_set_bits

junaid1460:
We have an inbuilt function to count number of set bits in Integers


AdiChat:
Looks good 👍 

Issue #714, opened 2017-10-06
Next larger number with same digits 


Arunraj97:
The program must take a number as input and give the next larger number with same digits as output
For eg :- If input is 697 output is 769
               If input is 5674 output is 5746

imfaizz:
I'd like to take this up

AdiChat:
Sure. Go for it 👍

mukul96:
I would like to implement it in python.

AdiChat:
Sure. Go for it 👍

mukul96:
In which Folder I have to put in the code.

mukul96:
please review #727

Rishabh570:
can i do this in c ??

AdiChat:
Sure 👍

PR #717, opened 2017-10-06
Find next larger number using java

imfaizz:
issue #714 using java


AdiChat:
Looks good 👍 

InfiniteCoder:
@imfaizz Hey, I have noticed that somehow some stray characters have gotten added to every program you have written. It probably has to do something with either space or tab you use, it's not what it's supposed to be 😉   Do have a look

Issue #703, opened 2017-10-06
Segment Tree with Lazy Propagation


shubhamdhama:
I would like to work on its implementation in C++.

AdiChat:
Sure. Go for it 👍

PR #721, opened 2017-10-06
Segment Tree with Lazy Propagation #703

Jagrati619:
Implementation of Segment Tree with Lazy Propagation in Java


AdiChat:
Looks good 👍 

PR #731, opened 2017-10-06
Added tree sort example in python

Norbinsh:
Added tree sort example in python


AdiChat:
Looks good 👍 

PR #744, opened 2017-10-06
Top view of binary tree

anku580:
This is the top view of a binary search tree in c language.

### DIFF
--- a/code/data_structures/Tree/top view.c
+++ b/code/data_structures/Tree/top view.c
@@ -1,0 +1,51 @@
+#include<stdio.h>
+#include<malloc.h>
+//make node of tree
+struct node {
+	int data;
+	struct node* right;
+	struct node* left;
+};
+//it creates the binary tree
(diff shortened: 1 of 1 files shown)


code/data_structures/Tree/right view of tree.cpp
```diff
@@ -0,0 +1,56 @@
...
+ 
+struct node* create()
+{
+    struct node *p;
+    int x;
+    printf("Enter data(-1 for no data):");
+    cin>>x;
+    
+    if(x==-1)
+        return NULL;
```
iattempt: use nullptr instead NULL in c++

code/data_structures/Tree/right view of tree.cpp
```diff
@@ -0,0 +1,56 @@
+#include<iostream>
+using namespace std;
```
iattempt: recommend  do not define std directive in global namespace

PR #750, opened 2017-10-06
Create prime_factors

mor11235:


code/unclassified/prime_factors
```diff
@@ -0,0 +1,18 @@
+void find_prime_factor(int n){
```
AdiChat: Kindly write complete code with header and main function 👍

mor11235: thank you very much :)

Issue #739, opened 2017-10-06
Convert number to binary and vice versa


mor11235:
Can I implement in C++?

AdiChat:
Sure. Go for it 👍

dgarvit:
Can I implement this in Java?

AdiChat:
Sure @dgarvit Go for it 👍

mor11235:
#745

ezeq3:
Can I implement this in Python?

AdiChat:
Sure @ezeq3 Go for it 👍

sarthak268:
In case this is still open I can do this in python or C++

AdiChat:
Go ahead @sarthak268 👍

darsh7807:
@sarthak268  you are still doing it?   I would like to do this in c++,c!

AdiChat:
Go for it :+1:

darsh7807:
@AdiChat  my pull request in c++ #842

Josh5231:
Just submitted javascript version. #898

nehal6657:
can i add in python

PR #751, opened 2017-10-06
Convert number to binary and vice versa in Java. #739

dgarvit:
Simple java class with 2 methods which can convert a given number to its binary number string, and vice versa.
#739


AdiChat:
Looks good 👍 

PR #753, opened 2017-10-06
Updated from raw array to use vector

arnavb:
This reduces a line of code, and raw arrays are error prone.


AdiChat:
Looks good 👍 

PR #758, opened 2017-10-06
Create sum_vs_xor.c

TanviKumar:


code/bit-manipulation/sum_vs_xor/sum_vs_xor.c
```diff
@@ -0,0 +1,25 @@
+#include <math.h>
```
AdiChat: Which problem are you trying to solve? `sum_vs_xor` is not giving enough information to review the flow of the code 👍

TanviKumar: Counts the number of values between 0 and a given number that satisfy the condition x+n = x^n (0<=x<=n). Common problem associated with bit manipulation. Logic behind it is that xor simulates binary addition without actually doing the carry over so the algorithm is simply to find the number of 0s in the given number and take 2^no.of zeros. I suppose I can change code to return only number of zeros thereby allowing its use in other algorithms.

AdiChat:
Looks good 👍 

PR #761, opened 2017-10-06
Sum vs Xor

TanviKumar:


AdiChat:
What this code is trying to solve? `sum_vs_xor` does not provide enough information 👍

AdiChat:
Looks good 👍 

PR #782, opened 2017-10-06
Update with similar as #753, fixed indentation

arnavb:


AdiChat:
Looks good 👍 

Issue #442, opened 2017-10-04
C++ transitive closure


qoo2p5:
I've implemented it: #440.

PR #783, opened 2017-10-06
Python prog for transitive closure #442

Jagrati619:


AdiChat:
Looks good 👍 

Issue #779, opened 2017-10-06
Add quick sort to divide_conquer


AkshayCHD:

AkshayCHD:
I would like to work on this issue

AdiChat:
Sure, go ahead 👍

bamsarts:
check my PR #796

Raghav17:
I am doing it in C.

PR #796, opened 2017-10-06
add quicksort to divide qonquer in python

bamsarts:
#779 added quicksort to divide qonquer in python


AdiChat:
Looks good 👍 

PR #801, opened 2017-10-07
Added basic linear regression example

codemastermd:
Added a basic example of linear regression with one variable


AdiChat:
Looks good 👍 

PR #813, opened 2017-10-07
Fix nullptr not defined errors and refactor code

adeen-s:
Replace nullptr with NULL from the stddef.h header file 
Add spaces to improve readability and format code according to Google styleguide


AdiChat:
Looks good 👍 

PR #823, opened 2017-10-07
Adding fractional knapsack code in python.

amartyaamp:
Fractional knapsack implementation in python. Comments are kept same as the cpp implementation.


AdiChat:
Looks good 👍 

PR #827, opened 2017-10-07
average

win72:


codeunclassified/average
```diff
@@ -0,0 +1,16 @@
+#include <iostream>
+#define N;//length of the array
+using namespace std;
+
+int average(int []arr){
+    int sum=0;
+    for (i=1;i<N;i++){
+        sum+=arr[i];
+    }
+    return sum/N;
```
AdiChat: This may throw a warning

codeunclassified/average
```diff
@@ -0,0 +1,16 @@
...
+}
+
+int main() {
+	// your code goes here
```
AdiChat: Kindly write sample code

PR #829, opened 2017-10-07
coprimes

win72:


coprimes
```diff
@@ -0,0 +1,34 @@
+#include <iostream>
```
AdiChat: Kindly place the code in proper location

coprimes
```diff
@@ -0,0 +1,34 @@
...
+using namespace std;
```
AdiChat: No extension of your file? It should be `.cpp`

PR #838, opened 2017-10-07
Greatest digit in a integer c 

darsh7807:
C implementation of greatest digit in integer


AdiChat:
Looks good 👍 

Issue #844, opened 2017-10-07
Add Sleep Sort


ahhmarr:
Add the code for Sleep sort in any language ( C, C++, Java, Go, Python or any other)

The code should be placed at [code/sorting/Sleep_sort](https://github.com/OpenGenus/cosmos/tree/master/code/sorting/Sleep_Sort)

Note: multiple contributors can work on this issue as it has multiple parts (languages)

Your pull request will be reviewed by maintainers instantly.

For contribution guidelines, see [this](https://github.com/OpenGenus/cosmos/wiki/contribute)

If you need any help, let us know.

ahhmarr:
#845 
for JavaScript

amitsin6h:
#1018 Sleep Sort using php @AdiChat

IAmBullsaw:
Implemented a sleep sort in ruby, please revise #1266 :smirk_cat:

PR #849, opened 2017-10-07
Add C++ Sleep Sort (#844)

cxong:


code/sorting/Sleep_Sort/Sleep_sort.cpp
```diff
@@ -0,0 +1,17 @@
+#include <unistd.h>
```
AdiChat: Kindly consult this [C code](https://github.com/OpenGenus/cosmos/blob/master/code/sorting/Sleep_Sort/Sleep_Sort.c) and make necessary changes 👍

cxong: what's wrong with using unistd?

cxong: Nevermind; I updated to use standard C++11 functions 😄

PR #858, opened 2017-10-07
Median in Rust

Jay9596:
Implemented median calculator from array in Rust.


AdiChat:
Looks good 👍 

PR #865, opened 2017-10-07
Cycle Sort in Go

amirkarimi:
Just translated Python sample code from [Wikipedia page](https://en.wikipedia.org/wiki/Cycle_sort) to Go.


AdiChat:
Looks good 👍 

PR #872, opened 2017-10-07
replace 0 with 5 in an integer program in c++

tanmai-h:
this solves issue #603


AdiChat:
The `C++` code already exists [see](https://github.com/OpenGenus/cosmos/blob/master/code/mathematical-algorithms/replace_0_with_5/replace.cpp) Kindly browse through the code base before doing a task 👍

arnavb:
@tanmai-h The code for this PR already exists [here](https://github.com/OpenGenus/cosmos/blob/master/code/mathematical_algorithms/src/replace_0_with_5/replace_0_with_5.cpp). As such, I would recommend you close this and instead work on another algorithm not yet implemented!

PR #873, opened 2017-10-07
Fixed some Armstrong Number errors

arnavb:
.Moved python version to right folder, and added header comment.


AdiChat:
Looks good 👍 

PR #879, opened 2017-10-07
CoPrime Integers

ArjitJ:
This Program takes 2 integer inputs from user and tells whether the two integers are coprime or not.


AdiChat:
The `C++` code already exists [see](https://github.com/OpenGenus/cosmos/blob/master/code/mathematical-algorithms/coprime_numbers/coprime_numbers.cpp). Kindly browse through the code base before working on a task 👍

ArjitJ:
Yeah but he has used iterator method and i have used Recursion..
I thought that the users should be acquainted with both these techniques.

AdiChat:
I see. You should mention it in the filename and pull request comment 👍 
Looks good 👍 

Issue #694, opened 2017-10-06
Minimum height of binary tree


piyush6348:
I would like to implement this in c++

AdiChat:
Sure. Go for it 👍

anku580:
i would like to implement in c++

AdiChat:
Sure @anku580 Go for it 👍 

Be sure to check existing implementations 👍

piyush6348:
please review #715

rstalekar:
I am implementing this in C

rstalekar:
Created a pull request for the minimum height of tree, please review

Abishek10:
I have implemented this in Python. Please review #881 .

priyalaggarwal:
Can I proceed to implement this for Java?

AdiChat:
Sure 👍

priyalaggarwal:
#1103 for Java

PR #881, opened 2017-10-07
Add min height of binary tree Python implementation

Abishek10:
Python implementation of #694


AdiChat:
Looks good 👍 

PR #882, opened 2017-10-07
find the average of array of n elements under 100

invincibel:
this is code to find the average of n elements in an array.


AdiChat:
Kindly place the code at the correct location 👍

PR #887, opened 2017-10-07
Update from original

manishsundriyal:


AdiChat:
The `C++` code already exists [see](https://github.com/OpenGenus/cosmos/blob/master/code/unclassified/average/average.cpp) 👍

PR #888, opened 2017-10-07
To count the number of digits

090max:
To count the number of digits using functions


hwchiu:
1. Please move this file to corresponding category. 
2. Don't forget the OpenGenus comment!

PR #900, opened 2017-10-07
Fix bubble sort example in Go

DiSiqueira:
There is no package called `algoutils` in standard libraries, also removed swap function since there are multiple assignments in GO.


AdiChat:
Looks good 👍 

Issue #808, opened 2017-10-07
Find the median of array of numbers


win72:
Can I implement in C++?

minijainsahab:
implemented in java. check #820

AdiChat:
Sure. Go for it 👍

win72:
Done. PR #830 (In C++).

Raghav17:
Where to place the code?

minijainsahab:
@Raghav17 add your code here(code/dynamic_programming/Array_Median)

Jay9596:
Rust implementation #858

amp180:
Python implementation in pr #906.

c650:
This should really be done with [Quickselect](https://en.wikipedia.org/wiki/Quickselect). It'll be faster in the average-case [O(n)] than sorting the array [O(nlogn)] and _then_ finding the median.

akhil117:
Can I implement in c??

akhil117:
pull for https://github.com/OpenGenus/cosmos/pull/1082 in c

PR #906, opened 2017-10-08
Added python median.

amp180:
Hi,
I added a function to compute the median value of a list as described in issue #808.
I'll remove the tests if they're not needed.
Thanks :)


AdiChat:
Looks good 👍 

PR #915, opened 2017-10-08
Added code to take User Input

HarendraSingh22:
Earlier the code only check in a particular predefined array and searched for a predefined number


AdiChat:
Looks good 👍 

Issue #392, opened 2017-10-04
Add B Tree


AdiChat:
Add the code for `B Tree` in any language ( `C`, `C++`, `Java`, `Go`, `Python` or any other)

The code should be placed under folder `b_tree` located at [code/data_structure](https://github.com/OpenGenus/cosmos/tree/master/code/data_structures)

**Note**: multiple contributors can work on this issue as it has multiple parts (languages)

Your pull request will be reviewed by maintainers instantly.

For contribution guidelines, see [this](https://github.com/OpenGenus/cosmos/wiki/contribute)

If you need any help, let us know.

dublevsky:
Hi! I'd like to work on that.

shashwatjolly:
Hi! How can I submit the code for this?

AdiChat:
Sure @dublevsky Go for it 👍

AdiChat:
@shashwatjolly You need to fork this repository and then, commit code in your local working copy. Thereby, you can submit a pull request which we will review and merge 👍

prateekiiest:
PR #395

shaikkamran:
heyy I would like to do this in Python

AdiChat:
Sure. Go ahead 👍

Abhishek2019:
Let me implement b_tree in python3?

PR #927, opened 2017-10-08
Create BTree.go

Mujazoji:
Binary Tree implementation in go
#392


AdiChat:
Please note that B Tree is different from Binary Tree.

PR #941, opened 2017-10-08
Added sleep sort in go

the-ethan-hunt:
Solved Issue #844


AdiChat:
Sleep sort in `Go` already exists [see](https://github.com/OpenGenus/cosmos/blob/master/code/sorting/Sleep_Sort/Sleep_Sort.go) 👍

PR #943, opened 2017-10-08
python: added bead sort and fixed bubble sort

sirex:
Bubble sort inmplementation was ineficient.


AdiChat:
Is your `bead sort` implementation in `Python` different from existing one? If so, kindly reflect it in the filename or else delete it 👍 
Additionally, delete the `.gitingnore` file 👍

sirex:
Removed `.gitignore` and rename `bead sort` file by adding `numpy` suffix. Basically this is axactly the same algorithm, only numpy version is a bit more readable.

AdiChat:
Looks good 👍 

PR #952, opened 2017-10-08
use StringBuilder so appending is not O(n)

DwordPtr:
Appending Strings in java is O(n) while using a StringBuilder is O(1).
I know these are demo implementations for an idea but its a good idea to keep this in mind especially for people preparing for interviews.


AdiChat:
Looks good 👍 

DwordPtr:
Yeah this idea is quite cool!
I hope this helps some people out.

AdiChat:
Glad you liked it 😄

PR #962, opened 2017-10-08
Remodeled to be more C++ Friendly

arnavb:
Use vectors instead of arrays, since they are less error prone. Also don't have to keep track of size.


AdiChat:
Looks good 👍 

Issue #933, opened 2017-10-08
Add a pangram checker


palash25:
A good first timers issue.

ron-popov:
I would like to implement this in python , where should i place this file ?

AdiChat:
Under a folder named `pangram checker` in `string algorithms` 👍

PR #976, opened 2017-10-08
(Issue #933) Adding javascript implementation of a pangram validator.

ryanml:
Adding a simple JS class that will validate a string to check whether it's a pangram or not.


AdiChat:
Looks good 👍 

PR #980, opened 2017-10-09
convert int to binary in python

bradleyJohnson55:


AdiChat:
Looks good 👍 

PR #983, opened 2017-10-09
Create single_number.cpp

karanchawla:


arnavb:
Don't forget the OpenGenus comment!

karanchawla:
Updated. Thanks

code/unclassified/singleNumber/single_number.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+/* Function to find the unique number in an array */
+
+int singleNumber(vector<int>& A) 
+{
+    int result = A[0];
+    
+    for(int i = 1 ; i < A.size() ; i++)
+    {
+        result = result^A[i];
```
AdiChat: For your input, it will give 4 as answer. Did you intend to keep one unique number in the array?

AdiChat:
If so, then the code belongs to `code/bit-manipulation/twice_unique_number` and the `C++` code exists so try to commit the `C` code 👍

code/unclassified/singleNumber/single_number.cpp
karanchawla: Yes, that's the idea. To detect the one unique number in the given array.

AdiChat: According to your input, the output should be 2 and 6 👍

karanchawla: Yep, you're correct. Edited the test case to include 2 in the `nums` array. Now, the only unique number is 6, which is the output of the program.

PR #993, opened 2017-10-09
Karger's min cut algo (C++)

gauravarora011:


gauravarora011:
Added KRUSKAL's algo as well

AdiChat:
Looks good 👍 

Vitorvgc:
About your Kruskal algorithm ([here](https://github.com/OpenGenus/cosmos/blob/master/code/greedy-algorithms/kruskal_minimum_spanning_tree/kruskal.cpp)), I think it is actually written in C. I believe that it would be better to move this to C language, as it wouldn't need any modifications, and open space to a C++ implementation that uses more of the language's features.

AdiChat:
Yes, this is infact written in `C`. Changed in #1077 👍

PR #1004, opened 2017-10-09
Added Prime factorisation in python

likecs:
Added prime factorisation code in python. It is compatible with both python 2 and python 3. The method used for factorisation is [Pollard-Rho](https://en.wikipedia.org/wiki/Pollard%27s_rho_algorithm) with Brent's optimisation which helps detect cycle's faster. Also, a pre-computation is done so that prime checking and finding prime factor for small numbers is done in constant time.


AdiChat:
This code has already been merged in your previous pull request 👍

PR #1007, opened 2017-10-09
Did some updates with binary search

arnavb:
Updated to use non-recursive functions as well, and also fixed some minor errors. C++ified the C++ binary search, since currently it was a C style interface.


AdiChat:
Looks good 👍 

Issue #805, opened 2017-10-07
Find the average of array of number


win72:
Can I implement in C++?

AdiChat:
Sure. Go for it 👍

win72:
Done. PR #827

Gupta-shifali:
Can I implement this in java?

AdiChat:
Sure. Go for it 👍

Jay9596:
Rust implementation #856

foobar98:
C++ and Python 3.5 implementation #867 
Kindly review @AdiChat 
Thanks. :)

PR #1009, opened 2017-10-09
average calculation in Ruby

LesnyRumcajs:
#805


AdiChat:
Looks good 👍 

PR #1013, opened 2017-10-09
Add 2sum in Rust

twify93:
If anybody has a better solution I would love to see it! Pretty new to Rust, I am just glad I made it work.


AdiChat:
Looks good 👍 

PR #1016, opened 2017-10-09
Add Polynomial multiplication using FFT

likecs:
Added implementation of Polynomial Multiplication using Fast Fourier Transforms. The implementation in C++ currently uses templates so that it can be easily extended. Also, the codes have been tested for efficiency on online judges like codechef, codeforces, spoj etc.

### DIFF
--- a/code/mathematical-algorithms/multiply_polynomial/FFT.cpp
+++ b/code/mathematical-algorithms/multiply_polynomial/FFT.cpp
@@ -1,0 +1,142 @@
+// Templated Implementation of FFT.
+// Used to multiply 2 polynomials of size N in O(N logN).
+// Polynomials are passed as vectors with index of "i" is
+// coefficient of x^i in polynomial.
+
+#include <bits/stdc++.h>
+
+using namespace std;
+
+const int MAX = 131100;		//2^ceil(log2(N)) where N = Maximum size of Polynomial
+
+//Complex class: Quite faster than in-built C++ library as it uses
+// only functions required
+template<typename T> class cmplx {
+private:
+	T x, y;
+public:
+	cmplx () : x(0.0), y(0.0) {}
+	cmplx (T a) : x(a), y(0.0) {}
+	cmplx (T a, T b) : x(a), y(b) {}
+	T get_real() { return this->x; }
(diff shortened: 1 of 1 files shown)


code/mathematical-algorithms/multiply_polynomial/FFT.cpp
```diff
@@ -0,0 +1,142 @@
...
+		for (int i = 0; i < n; ++i) cc.push_back(aa[i] * bb[i]);
+		DFT(cc, true);
+		vector<T> ans(sc);
+		for (int i = 0; i < sc; ++i) ans[i] = cc[i].get_real() + 0.5;
+		return ans;
+	}
+};
+
+int main() {
+	vector<int> poly_1(4);		//1 + 3x^2 + 2x^3
```
arnavb: Why not just initialize the vector on one line? It is pretty redundant to assign every element on a separate line.

code/mathematical-algorithms/multiply_polynomial/FFT.cpp
```diff
@@ -0,0 +1,142 @@
...
+	poly_1[0] = 1;
+	poly_1[1] = 0;
+	poly_1[2] = 3;
+	poly_1[3] = 2;
+
+	vector<int> poly_2(3);		//2 + x + x^2
```
arnavb: Same here.

likecs: HI, using vector<int> poly_1(4) = {1, 0, 3, 2}; will only work for C++11 onwards. It doesn't work for previous versions. Hence, I used the above method.

likecs: https://stackoverflow.com/a/38739297/5161197

arnavb: Ok, that works.

AdiChat: I see. The comment in the code makes it clear 👍

AdiChat:
Looks good 👍 

PR #1019, opened 2017-10-09
Added easy exponentation power using gnu ext/numeric. Added modulo power

neckbosov:
this algorithm always implemented in gnu library

### DIFF
--- a/code/mathematical-algorithms/exponentiation_power/exponentation_power_gnu_easy.cpp
+++ b/code/mathematical-algorithms/exponentiation_power/exponentation_power_gnu_easy.cpp
@@ -1,0 +1,10 @@
+#include <ext/numeric>
+#include <iostream>
+using namespace std;
+using namespace __gnu_cxx;
+int main() {
+	long long base, exp;
+	cin >> base >> exp;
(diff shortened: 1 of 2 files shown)


arnavb:
I don't think this will get merged because of the dependency. Is it possible to get the same effect with the `pow()` function from `<cmath>`?

neckbosov:
No, because of floating point errors. I specially write that it is gnu implementation of this algorithm and many programmists use gnu c++ compiler.

arnavb:
Ok, but it is still a dependency, which may prevent it from getting merged.

neckbosov:
@arnavb , ext/numeric is included in gnu c++ standart library, like ext/pb_ds ans bits/stdc++.h. Code with  ext/pb_ds is also merged in cosmos.

arnavb:
Ok, then forget everything I said :/

AdiChat:
Yes, there are 3 other codes that currently use GNU specific dependencies. Merging this in the sense that GNU is pretty much common/ standard even though this will increase a step in our set-up guide 👍

Looks good 👍 

PR #1020, opened 2017-10-09
aadded cycle sort

agarwalsarthak129:


code/sorting/cycle_sort/cycle_sort_by_SKAL.cpp
```diff
@@ -0,0 +1,83 @@
+// C++ program to impleament cycle sort
+#include <iostream>
+using namespace std;
+ 
+// Function sort the array using Cycle sort
+void cycleSort (int arr[], int n)
```
arnavb: Since this is C++, you should probably use `vector`s, since they are less error prone than normal arrays, and don't require the size to be passed as a parameter.

AdiChat:
@agarwalsarthak129 The `C++` code for cycle sort exists and is same as your implementation ([see](https://github.com/OpenGenus/cosmos/blob/master/code/sorting/cycle_sort/cycle_sort.cpp)). Kindly make changes as suggested by @arnavb in the existing file to get this merged 👍

PR #1036, opened 2017-10-09
added palindrome checker

### LINKED ISSUE #1033: Add algorithms to check if strings are palindromes?
Can I do this in Python and C++? I will place it in a new folder called `palindrome` in the string algorithms section. I plan on doing it in two ways, either by checking equality with the reverse of a string, by recursion, or by a loop.

anu0012:
Fixes #1033


AdiChat:
Looks good 👍 

PR #1039, opened 2017-10-09
Fixed gcd

neckbosov:
Prevous version was slow and made infinite loop on tests like "0 6". Also added gnu implementation.


AdiChat:
Looks good 👍 

PR #1044, opened 2017-10-09
Added cpp stl implementation of string permutations

neckbosov:


AdiChat:
Looks good 👍 

PR #1070, opened 2017-10-10
Add permutation/mouse in maze problem.

hwchiu:
1. Modify current Binary Search Tree to support `height()` function.
2. Implement partition_of_number in golang
3. add missing comment


AdiChat:
Looks good 👍 

PR #1081, opened 2017-10-10
Add Painter Problem

ivorynoise:


AdiChat:
As this is not a well known problem, kindly create a new category named `painter_problem` in the `unclassified_section` and create a `readme` with the description and add your code there. It will be good to merge then 👍

aggarwaldeepak:
@AdiChat I have updated the solution and put it as you guided.

AdiChat:
The problem is more of an application than it brings new approach and moreover, it is just a name suggested by a person and not adopted globally or locally. Hence, this code cannot be accepted in the current scenario.

Moreover, `readme` should contain description not links which is against the offline mission 👍 

Kindly take a look at other scopes of contribution 👍

ivorynoise:
Okay. This makes sense 👍

PR #1082, opened 2017-10-10
added the median code

akhil117:


code/dynamic_programming/Array_Median/med.c
```diff
@@ -0,0 +1,44 @@
...
+   {                                                           
+       int i,j,n;                                              
+       float median,a[N],t;                                    
+                                                               
+       printf("Enter the number of items\n");                  
+       scanf("%d", &n);                                        
+    /* Reading items into array a  */
+ 
+       printf("Input %d values \n",n);                         
+       for (i = 1; i <= n ; i++)
```
arnavb: This code will not work if the use enters a value for `n` that is greater than 10, since that is what N is set to. You should create a dynamic array using `malloc`, in which the size is determined at runtime, so this works well.

code/dynamic_programming/Array_Median/med.c
```diff
@@ -0,0 +1,44 @@
...
+#define N 10                                                
+                                                               
+  int main( )                                                     
+   {                                                           
+       int i,j,n;                                              
+       float median,a[N],t;                                    
+                                                               
+       printf("Enter the number of items\n");                  
+       scanf("%d", &n);                                        
+    /* Reading items into array a  */
```
arnavb: This code will not work if the use enters a value for `n` that is greater than 10, since that is what N is set to. You should create a dynamic array using `malloc`, in which the size is determined at runtime, so this works well.

code/dynamic_programming/Array_Median/median.c
```diff
@@ -0,0 +1,53 @@
...
+      }
+   }
+}
+
+int main() {
+	int i;
+	int n;
+	printf("please enter the size\n");
+	scanf("%d",&n);
+	float a[n];
```
arnavb: While Variable length arrays (VLAs) are **mostly** supported in C99, and definitely after it, some compilers still could define the macro ` _STDC_NO_VLA_` ([Source](https://stackoverflow.com/questions/16672322/malloced-array-vs-variable-length-array#comment23988958_16672322)) and not implement variable length arrays. So, just to be safe, you should use `malloc`, as follows

```c
float* a = (float*)malloc(n * sizeof(float));

// ... code for quicksort ...

free(a);
```
This will guarantee standards compliance for your code, regardless of compiler.

PR #1086, opened 2017-10-10
Right threaded binary tree in C++

arpank10:
Created implementation in C++


code/data_structures/Tree/right_threaded_binary_tree/right_threaded_binary_tree.cpp
```diff
@@ -0,0 +1,157 @@
+/* Part of Cosmos by OpenGenus Foundation */
+/*
+Right-Threaded Binary Tree implementation in C++
+*/
+//Author: Arpan Konar
+#include <bits/stdc++.h>
+using namespace std;
+typedef long long ll;
+#define REP(i,n) for(long long i=0;i<(n);i++)
```
arnavb: Don't use syntax macros. Everyone doesn't agree on what a good syntax is, and it is confusing for some people.

AdiChat:
Looks good 👍 

code/data_structures/Tree/right_threaded_binary_tree/right_threaded_binary_tree.cpp
arpank10: Thanks for pointing out. Though I haven't used it in the code.

PR #1088, opened 2017-10-10
Random number selection from a stream

Maneeeshita:
Given a stream of numbers, generate a random number from the stream using no extra space.


AdiChat:
Looks good 👍 

PR #1091, opened 2017-10-10
Create BST.c

himanshu-daga:
Successfully tested and added BST.c


AdiChat:
In general, `binary search tree` has the same basic idea as `binary search` so, expanding to a new category is not favourable 👍

PR #1092, opened 2017-10-10
Quickhull Algorithm added

AmanNim98:
Quickhull algorithm added...

I've stared your repository, star mine also.


AdiChat:
The code belongs to `computational geometry` category.

Kindly, illustrate an usage through an input data in the code and add `py` extension as well as relocating it. It will be good to merge then 👍

PR #1094, opened 2017-10-10
Add biggest_suffix.c

sketch873:
Added /code/string-algorithms/biggest_suffix


AdiChat:
This should go to `unclassified` section as this is more about implementation 👍

PR #1097, opened 2017-10-10
Remove redundancy of folder names and clean up

nirajpandkar:
There are two folders in the main folder "divide_conquer" named - "Quick_Sort" containing one cpp implementation and "quick_sort" containing python, c, and cpp implementation.

Merged the folders to remove redundancy.


arnavb:
There is also another `quick_sort` folder in `sorting`.

nirajpandkar:
Yeah was going to mention! But had no idea what to do about it because they are legit two categories - divide and conquer algos & sorting. What do you suggest is to be done? Will be happy to modify :)

arnavb:
They are both implemented in the same way (I just looked at the C++ version), so I believe it would be best to just combine them all into the `sorting` folder.

nirajpandkar:
Placed all the quick sort implementations in one place - `sorting` folder, irrespective of duplicacy. (The two `c` codes are the same except some comments and variable changes. Also arranged in folders for better visibility.

arnavb:
I think you should get rid of all the duplicate ones. I mean, for the c++ one, you could keep any one of the three, since they are pretty much all the same other than some minor differences.

arnavb:
I think this can be merged now. Good Job!

nirajpandkar:
Done. :+1: Executed the codes and saw two of them had compilation errors. In other cases kept ones which were commented over ones which weren't.

arnavb:
@AdiChat I think this PR can be merged now. Why don't you review it as well?

nirajpandkar:
Wait I wanted to create a new pull request for this python implementation. Why did these commits come over here.

arnavb:
@nirajpandkar Use the `git revert <commit-hash>` command from the command line.

nirajpandkar:
Phew :relieved:

AdiChat:
The purpose of duplicate codes at multiple location is to keep each category self satisfying in itself.

We do not have the feature to setup links (unlike in os) between folders of the repository. Currently, we have to maintain this manually but a better way to manage this will soon be incorporated in Cosmos 👍

That being said, merging categories is not feasible at the moment 👍

arnavb:
> The purpose of duplicate codes at multiple location is to keep each category self satisfying in itself.

@AdiChat  Could you clarify what you mean by that?

AdiChat:
I meant _a category should be complete in itself by removing dependencies between categories_.

arnavb:
Ok, thanks for clarifying

nirajpandkar:
Oh ok. No problem. :smiley: Can we close this PR?

arnavb:
@AdiChat Does what you said still apply to the to quick_sort directories inside divide_and_conquer? Those are exactly the same, and should be able to be merged.

Issue #1093, opened 2017-10-10
Add 0/1 knapsack implementations in other Languages.


amartyaamp:
Currently implemented with C and C++ only.

amartyaamp:
I will be going ahead with python implementation for the same.

bamsarts:
check my PR #1100

darioTecchia:
Hi! Check my PRs #1101 and #1098

amartyaamp:
Raised #1106 for python implementation.

UtkarshGupta-CS:
I will be going ahead with Javascript implementation for the same.

PR #1098, opened 2017-10-10
ADDED code in Python

darioTecchia:
#1093 Added code in Python

Credits: https://rosettacode.org/wiki/Knapsack_problem/0-1#Python

### DIFF
--- a/code/dynamic_programming/knapsack/knapsack.py
+++ b/code/dynamic_programming/knapsack/knapsack.py
@@ -1,0 +1,53 @@
+try:
+    xrange
+except:
+    xrange = range
+ 
+def totalvalue(comb):
+    ' Totalise a particular combination of items'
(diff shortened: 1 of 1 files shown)


amartyaamp:
I see the changes completely same as the rosettacode link given.
Have you tested this ? 
Also, isn't it a better/simpler idea if you assume input as list of weights and values rather than tuple of tuples?

darioTecchia:
Yup, i've tested it and it works. 

Yes, maybe. I can try this solution later.

AdiChat:
We do have array based implementation in other codes. In that sense, this code is good. 
Having other sources is not favourable, but this being your first contribution, it is. 

Looks good 👍 

darioTecchia:
Oh ok, I understand.

Thanks you!

PR #1116, opened 2017-10-10
Python Implementation of coin_change and CPP code fix

amartyaamp:
Following are the changes - 

1. Python implementation - coin_change.py with proper comments  
2. CPP code fix - order of table filling was not correct, fixed it and commented the code.
3. CPP code fix 2 - order of top-down memoization is not correct - removed it for now.

### DIFF
--- a/code/dynamic_programming/coin_change/coin_change.py
+++ b/code/dynamic_programming/coin_change/coin_change.py
@@ -1,0 +1,32 @@
+#################
+##  Part of Cosmos by OpenGenus Foundation
+#################
+
+
+def coin_change(coins, amount):
+
+	#init the dp table
+	tab = [0 for i in range(amount+1)]
+	tab[0] = 1  #base case
+
+	for j in range(len(coins)):
+		for i in range(1, amount+1):
+			if coins[j] <= i:
+				# if coins[j] < i then add no. of ways -
+				# - to form the amount by using coins[j]
+				tab[i] += tab[i-coins[j]]
+
+	#final result at tab[amount]
+	return tab[amount]
+
+def main():
+
+	coins = [1,2,3,8] # coin denominations
+	amount = 3 # amount of money
+
+	print "No. of ways to change - {}".format(coin_change(coins, amount))
+
+	return
+
+if __name__ == "__main__":
+	main()
--- a/code/dynamic_programming/coin_change/coinchange.cpp
+++ b/code/dynamic_programming/coin_change/coinchange.cpp
@@ -4,52 +4,33 @@ using namespace std;
 /*
  * Part of Cosmos by OpenGenus Foundation
 */
+#define MAX 100
 
-///Coin Change Problem 
-///Top Down DP
-int dp[100] = {0};
-int coinWaysTD(int amt,int coins[],int n){
-    if(amt==0){
-        dp[0] = 1;
-        return 1;
-    }
-    ///Recursive Case
-    int ans = 0;
+int coinWays(int amt,int coins[],int n){
+
+    // init the dp table
+    int dp[MAX] = {0};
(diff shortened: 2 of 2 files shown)


code/dynamic_programming/coin_change/coin_change.py
```diff
@@ -0,0 +1,32 @@
...
+	return tab[amount]
+
+def main():
+
+	coins = [1,2,3,8] # coin denominations
+	amount = 3 # amount of money
+
+	print "No. of ways to change - {}".format(coin_change(coins, amount))
+
+	return
```
arnavb: Unnecessary `return`.

code/dynamic_programming/coin_change/coinchange.cpp
```diff
@@ -4,52 +4,33 @@ using namespace std;
 /*
  * Part of Cosmos by OpenGenus Foundation
 */
+#define MAX 100
```
arnavb: Use a `const int` instead of a macro. Macros are generally not recommended in C++.

code/dynamic_programming/coin_change/coinchange.cpp
```diff
@@ -4,52 +4,33 @@ using namespace std;
...
-    ///Store and return for the first time
-    dp[amt] = ans;
-    return ans;
-}
-
-///Bottom Up DP
-int coinWaysBU(int amt,int coins[],int n){
-    int dp[100] = {0};
-    dp[0] = 1;
+int coinWays(int amt,int coins[],int n){
```
arnavb: Vectors would be the better choice here, since this is C++. Will not need to pass size either for vectors.

code/dynamic_programming/coin_change/coinchange.cpp
```diff
@@ -4,52 +4,33 @@ using namespace std;
...
 
-    for(int i=1;i<=amt;i++){
-        for(int j=0;j<n;j++){
-           if(i-coins[j]>=0){
-                dp[i] += dp[i-coins[j]];
+    // init the dp table
+    int dp[MAX] = {0};
```
arnavb: Same here as above.

code/dynamic_programming/coin_change/coinchange.cpp
```diff
@@ -4,52 +4,33 @@ using namespace std;
...
 
 
 int main(){
 
-    int coins[] = {1,2,3,8};
-    int amount = 3;
-    cout<<coinWaysTD(amount,coins,4)<<endl;
-    cout<<coinWaysBU(amount,coins,4)<<endl;
-return 0;
+    int coins[] = {1,2,3,8}; // coin denominations
```
arnavb: Same as above.

amartyaamp: same as above.

code/dynamic_programming/coin_change/coinchange.cpp
amartyaamp: same as above.

code/dynamic_programming/coin_change/coinchange.cpp
amartyaamp: I just extended whatever code was previously present. Prefer vectors personally. Will do the changes.

code/dynamic_programming/coin_change/coinchange.cpp
amartyaamp: Will do it.

code/dynamic_programming/coin_change/coin_change.py
amartyaamp: This return marks the end of a function. Good for readability.

AdiChat:
Looks good 👍 

PR #1118, opened 2017-10-10
Changed implementation to bottom-up DP

chronologos:
Previous implementation was naive recursion.


AdiChat:
Looks good 👍 

PR #1129, opened 2017-10-11
Add python implementation of nqueens using backtracking

nirajpandkar:


code/backtracking/n_queen/nqueen.py
```diff
@@ -0,0 +1,48 @@
+
+import json
```
hwchiu: It seems we don't need jsom library here, could you remove this unused imprort?

code/backtracking/n_queen/nqueen.py
```diff
@@ -0,0 +1,48 @@
...
+global x
+x=[]
+global n
+n=0
+global flag
```
hwchiu: how about we initialize all global variable together?
Maybe in here or in main function?

code/backtracking/n_queen/nqueen.py
```diff
@@ -0,0 +1,48 @@
...
+				flag=1
+			else:
+				nqueens(k+1)
+
+
+def main():
+	global n
+	global x
+	global flag
+	flag=0
```
hwchiu: As mentioned above, initialize all global variable together.

code/backtracking/n_queen/nqueen.py
nirajpandkar: Oops. Got past me. Will remove it.

code/backtracking/n_queen/nqueen.py
nirajpandkar: Sure thing.

nirajpandkar:
Done :+1:

hwchiu:
Hi @AdiChat  
This patch looks good to me.

AdiChat:
Looks good 👍 

Issue #991, opened 2017-10-09
Add Shaker sort in any language


AdiChat:
Add the code for `Shaker sort` in any language ( `C`, `Java`, `Go`, `Python`, `JavaScript`, `Swift` or any other)

The code should be placed at [code/sorting/shaker_sort](https://github.com/OpenGenus/cosmos/tree/master/code/sorting/shaker_sort)

Sample code in `C++`: [see](https://github.com/OpenGenus/cosmos/blob/master/code/sorting/shaker_sort/shaker_sort.cpp)

**Note**: multiple contributors can work on this issue as it has multiple parts (languages)

Your pull request will be reviewed by maintainers instantly.

For contribution guidelines, see [this](https://github.com/OpenGenus/cosmos/wiki/contribute)

If you need any help, let us know.

Jecoms:
I'll implement in C#.

etrexel:
I will do a python implementation.

protostellarsun:
Ah, Python is taken, I'll do Java then! Do you guys want any functional language implementations, e.g. OCaml?

948080952:
I will do a Objective-C implementation

arpank10:
I will do a C implementation

AdiChat:
Sure @protostellarsun functional languages will be great 😄

948080952:
@AdiChat I already finish the shaker implementation in Objective-C

JorgeAndd:
Doing javascript implementation

Ondama:
I'm doing it in rust

driesvb:
I'll do a VB.net implementation.

JeffreyCA:
I can do Swift!

Clorophila:
I can do Golang.

AdiChat:
Sure @Clorophila Go for it 👍

warned101:
Can I add shaker sort in C++?

arnavb:
@warned101 Go for it!

warned101:
I think that algorithm has already been implemented in C++.
[shaker_sort.cpp](https://github.com/OpenGenus/cosmos/blob/master/code/sorting/src/shaker_sort/shaker_sort.cpp)
Please check and if there's something to be fixed in this, let me know!

arnavb:
@warned101 There are still things that can be fixed in that file (namely `using namespace std;`). Please consider fixing those problems.

warned101:
Okay, I will take care of that!

PR #1135, opened 2017-10-11
Add shaker sort in PHP

adrianuf22:
Implements shaker sort with PHP. 

For #991


AdiChat:
Looks good 👍 

PR #1139, opened 2017-10-11
Updated Trie.java for any char, not just lowercase

savli:
Previously the Trie.java class could only work for lowercase alphabet characters (i.e. a-z). Now the Trie can work for any java character


AdiChat:
Looks good 👍 

Issue #1152, opened 2017-10-11
Add code for Lexicographic rank of a string in any language


AdiChat:
Add the code for `Lexicographic rank of a string` in any language ( `C`, `C++`, `Java`, `Go`, `Python` or any other)

The lexicographic rank of a string is the rank of the string among all permutations of its characters sorted lexicographically.

The code should be placed at [code/mathematical-algorithms/lexicographic_string_rank](https://github.com/OpenGenus/cosmos/tree/master/code/mathematical-algorithms/lexicographic_string_rank)

**Note**: multiple contributors can work on this issue as it has multiple parts (languages)

Your pull request will be reviewed by maintainers instantly.

For contribution guidelines, see [this](https://github.com/OpenGenus/cosmos/wiki/contribute)

If you need any help, let us know.

bamsarts:
check my PR #1158

shubhamcr:
I am working on the C++ implementation of this.

shubhamcr:
Check my PR #1267

kanaygupta:
Can i work on the java implementation of this ?

kanaygupta:
check my PR #1544

PR #1158, opened 2017-10-11
add lexicographic string rank in c

bamsarts:
#1152 added lexicographic rank of string implement in java


hwchiu:
In your commit, it seems you have create a Java version? But I can't see it.
Do you missing it ?

AdiChat:
Looks good 👍 
It seems only `C` code is intended for this pull request 👍 

PR #1161, opened 2017-10-11
Init Baconian Cipher

siddver007:
Added Baconian Cipher to Cryptography


AdiChat:
Looks good 👍 

PR #1177, opened 2017-10-11
Cocktail sort addition

SoumitraAgarwal:
Cocktail Sort is a variation of Bubble sort. The Bubble sort algorithm always traverses elements from left and moves the largest element to its correct position in first iteration and second largest in second iteration and so on. Cocktail Sort traverses through a given array in both directions alternatively.


AdiChat:
Cocktail sort is same as shaker sort for which we have the `C++` and `Java` code. Kindly remove them. I will merge the artificial intelligence addition 👍

AdiChat:
Looks good 👍 

PR #1181, opened 2017-10-11
Implimentation of reverse queue in c++.

abhishekgoyal28:
Implimentation of reverse queue in c++.


AdiChat:
The `C++` implementation exists ([see](https://github.com/OpenGenus/cosmos/blob/master/code/data_structures/queue/reverse_queue/reverse_queue.cpp)) 👍

PR #1187, opened 2017-10-11
Fix bug and improve style of C++ version of quicksort

matthewcurry:
List of changes:
- Fix an off-by-one error that caused invalid data to be sorted
- Quicksort is all one word, so change file name and function signatures to match
- Added project message, and removed unnecessary comment
- Programatically load vector, instead of statically
- Use specific headers instead of bits/stdc++.h, as the latter is not standard
- Print vector before and after sort


AdiChat:
Looks good 👍 

AdiChat:
Looks good 👍 

PR #1189, opened 2017-10-11
Implementation of a byte swapper in Java

Awax56:
I have implemented a byte swapper in Java to switch a byte word between little endian and big endian and vice versa


AdiChat:
Looks good 👍 

PR #1197, opened 2017-10-11
add missing type cast

lazyc97:
without cast, it will be integer division and always return the rounded down result


AdiChat:
Looks good 👍 

PR #1200, opened 2017-10-11
Implementation of topological sort in c.

abhishekgoyal28:
Implementation of topological sort in c.
In cosmos/code/graph-algorithms/topological_sort/


AdiChat:
Topological sort in `C` already exists ([see](https://github.com/OpenGenus/cosmos/blob/master/code/sorting/topological_sort/topologicalSort.c)) 👍

PR #1214, opened 2017-10-11
Enforce PSR-2 on PHP bubble sort

camuthig:
Also updated the array syntax since all currently supported versions of PHP include it.


AdiChat:
Looks good 👍 

PR #1229, opened 2017-10-11
Vigenère cipher implementation in PHP

ilcapo1:
## **VigenereCipher** PHP class
A minimal OO VigenereCipher that supports both encrypt and decrypt methods with alphabetically-only input strings.

_Example usage:_

```php
$cipher      = new VigenereCipher();

$secret      = 'lemonlemonle';
  
$encrypted   = $cipher->encrypt('attackatdawn', $secret);

$plain       = $cipher->decrypt($encrypted, $secret);

echo $plain;  // outputs 'attackatdawn'


AdiChat:
Looks good 👍 

PR #1206, opened 2017-10-11
Create reverseStack.java

Z-Fikar:
#1169 and any issues about reverse stack

Reverse Stack created in Java Language

Issue #706, opened 2017-10-06
Reverse a queue 


phoenix157:
Can i work on this on c++?

AdiChat:
Sure. Go for it 👍

phoenix157:
Implemented at #708

jnight444:
I'd like to implement this in python

jnight444:
I'd like to implement this in Java

AdiChat:
Sure @jnight444 Go for it 👍

Z-Fikar:
I have implemented it in Python #1233 
@AdiChat Please review it when you have time, don't have to rush it as I see the PR counter keep increasing.

Z-Fikar:
@AdiChat  #1233 Ready to merge.

perhaps.

AdiChat:
Merged 👍

Rishabh570:
can i do this in c ??

AdiChat:
The `C` code has been done 👍 Kindly take a look at other scopes of contribution 😄

AdiChat:
Sure 👍

PR #1233, opened 2017-10-11
Create reverse_queue.py

Z-Fikar:
#706 

Reverse Queue created in Python Language.

Actually I do not want to work twice as creating a new Queue class, considering Queue in Python has been created [here](https://github.com/OpenGenus/cosmos/blob/master/code/data_structures/queue/queue/queue.py). I prefer to import it and give a link to it in my code, like this one #1206

But there are some differences with what I know and what I want, and those will affect my code. Such as:
+ The data entered can only be in integer data type
+ dequeue() method removes the first element and returning its value, which should not.
+ No method for front() and size()
+ Attributes can still be accessed from outside the class

But still, I have some doubt. 
- The size method() returns whether the current queue size or queue's max size.
- I don't know whether I should make the `max` variable to be private or public.

In the end, I make size() for returning the current queue size, and `max` to be private because I didn't use it outside the class.

### DIFF
--- a/code/data_structures/queue/reverse_queue/reverse_queue.py
+++ b/code/data_structures/queue/reverse_queue/reverse_queue.py
@@ -31,6 +31,10 @@ class Queue:
 	def size(self):
 		return len(self.__data)
 
+	# Returns the maximum size that the queue can hold
+	def capacity(selfs):
+		return len(self.__max)
+
 
 # Return all items from queue in list.
 def get(queue):
--- a/code/data_structures/queue/reverse_queue/reverse_queue.py
+++ b/code/data_structures/queue/reverse_queue/reverse_queue.py
@@ -1,0 +1,86 @@
+# Part of Cosmos by OpenGenus Foundation.
+
+# Make a new Queue class with list.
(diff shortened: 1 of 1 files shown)


arnavb:
The `size` method is actually really useful; you should keep it for the _current_ size. For the max size, consider creating a function `capacity`. The C++ STL does this for its containers as public member functions.

Z-Fikar:
Yeah, it is true. But it is not like my `Queue` class will be used for something else. <br/>
If someone want to know the algorithm behind `Queue`, they will go to [code/data_structures/queue/queue](https://github.com/OpenGenus/cosmos/blob/master/code/data_structures/queue/queue) not to [code/data_structures/queue/reverse_queue](https://github.com/OpenGenus/cosmos/blob/master/code/data_structures/queue/reverse_queue), that is how this repo works right? at least for me.

arnavb:
@Z-Fikar sure, I guess. Don't be negative ;). Maybe your code will be very famous!

AdiChat:
Looks good 👍 

PR #1236, opened 2017-10-11
eggDropping memoized impl in Haskell

chronologos:


AdiChat:
Looks good 👍 

Issue #1239, opened 2017-10-11
Add linked list in Ruby


cameronbroe:
An implementation of a singly linked list in Ruby is currently missing. Attaching a PR to this issue to add an implementation.

IAmBullsaw:
I'm on it :)

tomlazar:
This should be marked closed since a implementation was added

PR #1240, opened 2017-10-11
Implementation of a singly linked list in Ruby

cameronbroe:
Here is an implementation of a singly linked list in Ruby. The get and delete operations are index based instead of value based, as most list operations in standard libraries will use indexes anyway.

Addresses issue #1239

### DIFF
--- a/code/data_structures/linked_list/linked_list/linked_list.rb
+++ b/code/data_structures/linked_list/linked_list/linked_list.rb
@@ -1,0 +1,92 @@
+# Part of Cosmos by OpenGenus Foundation
+class Node
+	attr_accessor :data
+	attr_accessor :next
+
+	def initialize(data)
+		@data = data
+		@next = nil
+	end
+end
+
+class LinkedList
+	attr_accessor :head
+
+	def initialize
+		@head = nil
+	end
+
+	# Return the number of elements in the linked list
+	def size
+		current_node = @head
+		length = 0
+		until current_node.nil? do
+			length += 1
+			current_node = current_node.next
+		end
+		return length
+	end
+
+	# Insert a new node at end of linked list
+	def insert!(data)
+		if @head.nil? then
+			@head = Node.new(data)
+		else
+			curr_node = @head
+			until curr_node.next.nil? do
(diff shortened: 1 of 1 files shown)


code/data_structures/linked_list/linked_list/linked_list.rb
```diff
@@ -0,0 +1,92 @@
...
+class LinkedList
+	attr_accessor :head
+
+	def initialize
+		@head = nil
+	end
+
+	# Return the number of elements in the linked list
+	def size
+		current_node = @head
```
arnavb: Why not just keep track of the size instead of calculating it every time?

code/data_structures/linked_list/linked_list/linked_list.rb
```diff
@@ -0,0 +1,92 @@
...
+		length = 0
+		until current_node.nil? do
+			length += 1
+			current_node = current_node.next
+		end
+		return length
+	end
+
+	# Insert a new node at end of linked list
+	def insert!(data)
```
arnavb: Consider renaming this function `add` and adding  a separate function called `insert` which takes a parameter of an index, and inserts an element at (or before) that position. The whole point of linked lists is to make insertion easy and fast at any location, so omitting that key function nullifies the whole point of the class.

cameronbroe:
Great suggestions! I have made these modifications and appended them to this pull request.

code/data_structures/linked_list/linked_list/linked_list.rb
```diff
@@ -0,0 +1,111 @@
...
+
+	def initialize(data)
+		@data = data
+		@next = nil
+	end
+end
+
+class LinkedList
+	attr_accessor :head
+	attr_accessor :size
```
arnavb: You should add a member function for the size. Don't want the client code modifying it, which may mess up results (unless that doesn't happen in Ruby, since I don't know the language very well.

cameronbroe:
Right, modified attr_accessor to attr_reader. In Ruby, attr_accessor creates a getter and setter, attr_reader creates a getter, and attr_writer creates a setter.

PR #1255, opened 2017-10-12
Update Queue.cs with the actual file to go out.

Walt280:
Sorry for the inconvenience, I had forgotten to commit the file with comments.


lazyc97:
You should use StringBuilder instead of "+=" of String. C# string is immutable so they are very so to append.

Walt280:
@lazyc97 I agree with that, however, I chose `+=` instead of `StringBuilder` because I feel that `+=` is easier to understand in the code.

AdiChat:
Looks good 👍 

PR #1256, opened 2017-10-12
Create Minimum bits to flip such that XOR of A and B equal to C

Sarbamoy:
Given a sequence of three binary sequences A, B and C of N bits. Count the minimum bits required to flip in A and B such that XOR of A and B is equal to C.

### DIFF
--- a/code/bit-manipulation/Minimum bits to flip such that XOR of A and B equal to C
+++ b/code/bit-manipulation/Minimum bits to flip such that XOR of A and B equal to C
@@ -1,0 +1,33 @@
+// C++ code to count the Minimum bits in A and B
+#include<bits/stdc++.h>
+using namespace std;
+ 
+int totalFlips(char *A, char *B, char *C, int N)
+{
+    int count = 0;
+    for (int i=0; i < N; ++i)
+    {
+        // If both A[i] and B[i] are equal
+        if (A[i] == B[i] && C[i] == '1')
+            ++count;
+ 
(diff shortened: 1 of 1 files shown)


code/bit-manipulation/Minimum bits to flip such that XOR of A and B equal to C
```diff
@@ -0,0 +1,33 @@
+// C++ code to count the Minimum bits in A and B
```
arnavb: This file needs a file extension, needs to be shorter, and also needs to get rid of spaces.

code/bit-manipulation/Minimum bits to flip such that XOR of A and B equal to C
```diff
@@ -0,0 +1,33 @@
...
+#include<bits/stdc++.h>
+using namespace std;
+ 
+int totalFlips(char *A, char *B, char *C, int N)
```
arnavb: Consider using vectors instead of char pointers, since this is C++.

code/cryptography/rail_fence_cipher/Rail Fence Cipher – Encryption and Decryption
```diff
@@ -0,0 +1,127 @@
+// C++ program to illustrate Rail Fence Cipher
```
arnavb: Same applies here as top comment.

PR #1262, opened 2017-10-12
Create quick_sort.php

JulianMar:
Added quick_sort in php

### DIFF
--- a/code/sorting/quick_sort/quick_sort.php
+++ b/code/sorting/quick_sort/quick_sort.php
@@ -1,4 +1,6 @@
-function quick_sort( $array ) {
+<?php
+
+functionfunction quick_sort( $array ) {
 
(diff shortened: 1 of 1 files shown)


arnavb:
PHP code needs to be inside php tags, I think...

JulianMar:
Yes you are correct. I added them to the file

code/sorting/quick_sort/quick_sort.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+functionfunction quick_sort( $array ) {
```
arnavb: Is this a syntax error? Why is `function` written twice?

JulianMar: I don't know how that happend but it is fixed

PR #1272, opened 2017-10-12
Baconian cipher implementation in PHP

ilcapo1:
## **Bacon's cipher (aka Baconian cipher)** PHP class
A minimal OO BaconianCipher that supports both encrypt and decrypt methods with alphabetically-only input strings.

_Example usage:_

```php
$cipher    = new BaconianCipher();
  
$encrypted = $cipher->encrypt('LOREMIPSUMDOLOR'); 

// encrypted: ababb abbba baaab aabaa abbaa abaaa abbbb baaba babaa abbaa aaabb abbba ababb abbba baaab

$decrypted = $cipher->decrypt($encrypted);

echo $decrypted; // output: LOREMIPSUMDOLOR

### DIFF
--- a/code/cryptography/baconian_cipher/baconian.php
+++ b/code/cryptography/baconian_cipher/baconian.php
@@ -1,0 +1,85 @@
+<?php
+/**
+* Bacon's (aka Baconian) cipher (PHP)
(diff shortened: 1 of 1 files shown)


AdiChat:
Looks good 👍 

PR #1275, opened 2017-10-12
added to binary search to reduce computation from O(N) to O(logN)

rahul1809:
Modified the program logic to reduce complexity from linear to logarithmic in time.


code/mathematical-algorithms/Check_is_square/FindingISquare.cpp
```diff
@@ -1,25 +1,51 @@
-#include <iostream>
+#include <bits/stdc++.h>
 using namespace std;
-int main()
+
+typedef long long int lld;
```
arnavb: Why the _d_ at the end. Why not _i_?

code/mathematical-algorithms/Check_is_square/FindingISquare.cpp
```diff
@@ -1,25 +1,51 @@
...
+      mid = min + (max - min)/2;
+    }
+    else if (mid*mid < number)
+    {
+      min = mid + 1;
+      mid = min + (max - min)/2;
+
+    }
+    if (mid*mid == number)
+    return true;
```
arnavb: Add a tab for clarity

code/mathematical-algorithms/Check_is_square/FindingISquare.cpp
rahul1809: In C, we have 'lld' as format specifier for long long int . I'm just sticking to that convention of naming.

code/mathematical-algorithms/Check_is_square/FindingISquare.cpp
rahul1809: added tabs for clarity

AdiChat:
Kindly do not delete the existing code `FindingISquare.cpp` as it is using a different approach 👍

rahul1809:
Ok will do that in future. Add new and better approaches in seperate files.

PR #1279, opened 2017-10-12
Add sleep sort in Objective-C and Swift

948080952:
Besides, add swap method to some sort implementation in Swift


AdiChat:
Looks good 👍 

PR #1287, opened 2017-10-12
Converted C++ Sorts from Arrays to Vectors

arnavb:
Vectors don't require tracking of their size, and don't require the user to enter a size when entering input. They are also much safer than raw arrays, which decay to pointers at the slightest provocation. Vectors are also more flexible than raw arrays, since their size can be _changed_.


AdiChat:
Looks good 👍 

arnavb:
@AdiChat You should create a bot that says the "Looks good... Hope you are.." etc. 😛  Just a suggestion.

AdiChat:
Sounds great 😄 
The bot should also answer simple queries like "_May I do it?_" and others in the issues. I will take a look into it 👍

PR #1300, opened 2017-10-12
Add print singly linked list in reverse order in C

nj2610:
I want to add this cool program to print singly linked list in reverse order using recursion in c language.
Review and add if you like it.

### DIFF
--- a/josephus.c
+++ b/josephus.c
@@ -1,0 +1,42 @@
+#include <stdio.h>
+#include<stdlib.h>
+struct node
+{
+    int data;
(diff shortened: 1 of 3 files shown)


hwchiu:
Hi
1) Don't forget the OpenGenus comment! 👍
2) Place those files on different category directories (don't put them on the root directory)
3) n_queen in C is exist, you can check [here](https://github.com/OpenGenus/cosmos/blob/master/code/backtracking/n_queen/n%20_queen.c)

PR #1303, opened 2017-10-12
Add string pallindrome in python

nj2610:
I am adding a program to check whether a string is a palindrome or not in python language using smart use of slicing. Review and add if you like it.

### DIFF
--- a/josephus.c
+++ b/josephus.c
@@ -1,0 +1,42 @@
+#include <stdio.h>
+#include<stdlib.h>
+struct node
+{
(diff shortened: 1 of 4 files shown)


arnavb:
Your python implementation already exists, see [here](https://github.com/OpenGenus/cosmos/blob/master/code/string-algorithms/palindrome_checker/palindrome.py#L10).

FrankKair:
Yep, this PR should be closed.

PR #1310, opened 2017-10-12
Added recursive factorial implementation in PHP

andreigasparovici:


AdiChat:
The recursive factorial computation in `php` already exists ([see](https://github.com/OpenGenus/cosmos/blob/master/code/mathematical-algorithms/factorial/factorial.php)) 👍

PR #1321, opened 2017-10-12
Bit manipulation for looping over all subset masks of bits

jvikstrom:
Added bit manipulation trick for looping over all subset of bits in an input.

NOTE: there is no issue for this so I don't know if you want this or not.


AdiChat:
Looks great 😍 

PR #1325, opened 2017-10-13
Add memoization to shortest common supersequence solution

Vitorvgc:
Updated C++ solution to use memoization instead of naive recursion. I also substituted
 the `bits/stdc++.h` header as it is not available in all compilers.


AdiChat:
Looks good 👍 

PR #1326, opened 2017-10-13
Added topsort in java!

williamfiset:
Added two versions of the topological sort algorithm:
1) Implemented as an adjacency list
2) Implemented as an adjacency matrix
As a bonus, I also threw in finding the shortest path on a DAG with some examples.


AdiChat:
Looks good 👍 

Issue #1297, opened 2017-10-12
Check whether a number is a power of 2 using bit manipulation


shubhamcr:
To check efficiently whether a number is power of 2 using bit manipulation

shubhamcr:
Check PR #1298

Jiraphapa:
Implemented power of 2 in javascript #1484

nickdex:
Add C# implementation to check power of 2. Refer #1492

abhijeet010304:
I would like to work on it in C++  and pyhton please assign it to me as a part of GSSoC 20

HariniJeyaraman:
I'm a GSSoC'20 participant. Would love to work on this issue. In what languages is this issue open for now? @arnavb @AdiChat Please assign me this issue if it is still open

arnavb:
@HariniJeyaraman Check the corresponding code directory. You can do this in any available language.

PR #1329, opened 2017-10-13
Power of 2 - Bit manipulation in python - Issue #1297

grubdragon:
Checks if input number is a power of 2
Compatible across Python 2 and 3


AdiChat:
Looks good 👍 

PR #1330, opened 2017-10-13
Add Python Implementation

nickdex:
I added two methods similar to given sample as well as a method for finding intersection point if given line in given in Ax + Bx + C format


AdiChat:
The `python` code exists 👍

nickdex:
My mistake should have synced forks before committing. By the way how do you maintain syncs between forks. All the methods I searched for work, though they involve many steps. Is there an quick and stable way of doing it?

rishav-pandey:
@nickdex 

first make the remote branch locally from OpenGenus/cosmos
----> $ git remote add upstream https://github.com/OpenGenus/cosmos.git
check it using
----> $ git remote -v
after confirmation enter following command
----> $ git pull upstream master
then further enter following command
----> $ git push


and you are done.
I hope this could resolve your problem :+1:

nickdex:
@rishav-pandey 
Thanks for your answer. However your answer works when I am creating a branch locally from project (as you mentioned already). Now that I am a member I can push on to repository directly. I wanted to know how do people generally maintains sync between 'forks' (this problem comes up when you can't directly push to project repo)

rishav-pandey:
@nickdex 
There is not such online way  to sync between forked repository to original one on git.
You have to simply fetch and merge files (PULL) from original git branch to local master branch and the after updating the git repository locally You just PUSH it and your fork will be updated to original repository.
All the steps mentioned above will be only required

nickdex:
@rishav-pandey 
Alright, that's what I am doing for time being.

PR #1331, opened 2017-10-13
Optimized using ES6

IgorAntun:
Also added more tests and made it return the value (true/false) instead of console.logging it.


AdiChat:
Looks good 👍 

PR #1333, opened 2017-10-13
Area of Polygon - implemented under Python

grubdragon:
Added an implementation of the area of a polygon in python
(Compatible across Python 2 and 3)


AdiChat:
Looks good 👍 

PR #1337, opened 2017-10-13
2-D Line Intersection in Python

grubdragon:
Python 2 & 3 compatible implementation of point of intersection between two lines in their 2-point form


AdiChat:
Looks good 👍 

PR #1341, opened 2017-10-13
Implemented the distance between two points in python

grubdragon:
Compatible across python versions, python implementation of distance between two points added.


AdiChat:
Looks good 👍 

PR #1347, opened 2017-10-13
Create Quick_select.java

yatharthshahjpr:
Find kith largest element is equivalent to find (n - k)th smallest element in array. It is worth mentioning that (n - k) is the real index (start from 0) of an element.


AdiChat:
Looks good 👍 

yatharthshahjpr:
thank you will be contributing further !

PR #1363, opened 2017-10-13
recursive solution to searching linked lists

coreywho:


AdiChat:
Iterative solutions are preferred to recursive ones. Instead of modifying existing code, open a pull request with your code named `linked_list_recursive.py` 👍

PR #1369, opened 2017-10-13
Logistic regression addition

SoumitraAgarwal:
In statistics, logistic regression, or logit regression, or logit model is a regression model where the dependent variable (DV) is categorical.


AdiChat:
Looks good 👍 

PR #1374, opened 2017-10-13
Create Linear_regression.cpp

yatharthshahjpr:
code for linear regression in c++


arnavb:
You should combine this PR with your other one, since they depend on each other and are for the same purpose.

yatharthshahjpr:
Ok I have done the required Changes! :)

PR #1375, opened 2017-10-13
Create LinearRegression.h

yatharthshahjpr:
header files for cpp code of which a created the PR 5 min before!


arnavb:
You should combine this PR with your other one, since they depend on each other and are for the same purpose.

yatharthshahjpr:
Ok !  I have done the required changes! :)

PR #1382, opened 2017-10-13
Create Palindrome.kt

aborysionek:
Its a main with a boolean function that's lazily written, but works just fine
(it's easier to understand this way)


AdiChat:
Looks good 👍 

PR #1392, opened 2017-10-13
aes.py

angus-y:


UtkarshGupta-CS:
@angus-y this doesn't look like an algorithm implementation but more like the use of the existing library. Check other submissions they are complete algorithm implementation.

PR #1405, opened 2017-10-13
Update quick sort in python

nj2610:
I am adding quicksort in a very short method in python.Hope you will like it and add it.

### DIFF
--- a/quicksort.py
+++ b/quicksort.py
@@ -1,0 +1,10 @@
+l = [int(x) for x in input().split()]
+
+def Qsort(a):
+    if len(a)<=1:
+        return a
+    else:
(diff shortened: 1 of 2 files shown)


hwchiu:
Thanks your contribution, But there is a python implementation [here](https://github.com/OpenGenus/cosmos/blob/master/code/sorting/quick_sort/quick_sort.py).
You can modify the exist file to support more implement way.
Also, you need to put source code file into corresponding directory rather than outside.

FrankKair:
Also, remember to be compliant with the [PEP8](https://www.python.org/dev/peps/pep-0008/)!

quicksort.py
```diff
@@ -0,0 +1,12 @@
+# Part of Cosmos by OpenGenus Foundation
```
hwchiu: This file should be remove, since you already have the other same file on `sorting/quick_sort/quicksort.py`

PR #1411, opened 2017-10-13
Fix some compile error/warning and also adjust some golang indent style to tab.

hwchiu:


AdiChat:
Looks good 👍 

PR #1447, opened 2017-10-14
Added the exponential_search in c lang...

Rishabh570:


code/bit-manipulation/thrice_unique_number/thrice_unique_number.c
```diff
@@ -0,0 +1,17 @@
+#include<stdio.h>
```
AdiChat: The solution is expected to use bit manipulation. Kindly take a look at [this C++ code](https://github.com/OpenGenus/cosmos/blob/master/code/bit-manipulation/thrice_unique_number/threeUnique.cpp) and rewrite this 👍

code/bit-manipulation/twice_unique_number/twice_unique_number.c
```diff
@@ -0,0 +1,18 @@
+#include<stdio.h>
```
AdiChat: Same as above. The solution should use bit manipulation 👍

code/unclassified/create_prime_factors.c
```diff
@@ -0,0 +1,40 @@
+#include<stdio.h>
```
AdiChat: The code to generate prime factor should be placed [here](https://github.com/OpenGenus/cosmos/tree/master/code/mathematical-algorithms/prime_numbers_of_n). Kindly relocate 👍

PR #1456, opened 2017-10-14
Added Haskell implementation of the 2D line intersection finding program

sidmishraw:
Added a Haskell implementation of the 2D line intersection finding program and the Readme file for the package.


AdiChat:
Looks great 😍 

PR #1460, opened 2017-10-14
Decision tree implementation

SoumitraAgarwal:
A decision tree is a decision support tool that uses a tree-like graph or model of decisions and their possible consequences, including chance event outcomes, resource costs, and utility.


AdiChat:
Looks great 😍 

PR #1462, opened 2017-10-14
Added binary indexed tree

shumon84:
I added binary indexed tree for c++

### DIFF
--- a/code/data_structures/segment_tree/binary_indexed_tree.cpp
+++ b/code/data_structures/segment_tree/binary_indexed_tree.cpp
@@ -1,0 +1,23 @@
+#define MAX_N 100000
+int bit[MAX_N+1];
+int n;
+
+int sum(int i)
+{
+  int s=0;
+  while(i>0)
+    {
+      s+=bit[i];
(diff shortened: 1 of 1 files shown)


code/data_structures/segment_tree/binary_indexed_tree.cpp
```diff
@@ -0,0 +1,23 @@
+#define MAX_N 100000
+int bit[MAX_N+1];
+int n;
+
```
AdiChat: You need to define the structure of the node in the form of a class 👍

AdiChat:
The code is not correct. Moreover, you can learn about _binary index tree_ also know as _fenwick tree_ through this [C++ implementation](https://github.com/OpenGenus/cosmos/blob/master/code/data_structures/fenwick_tree/fenwick_tree.cpp)

shumon84:
Sorry, I did not know that the Fenwick tree and the binary index tree are the same.

AdiChat:
No problem @shumon84 Glad you learned something new 😄

PR #1469, opened 2017-10-14
add inversion count implementation with javascript

Jiraphapa:
This inversion count implementation with javascript is based on merge sort approach. The number of inversion can be achieved from calling mergeSort(array, size) function.


AdiChat:
Looks great 😍 
Your first open-source contribution has been awesome 🎉 

PR #1472, opened 2017-10-14
Reversed a stack using PUSH and POP. Fixes #1159

### LINKED ISSUE #1159: Create reverse_stack.cpp
creating reverse_stack.cpp in **/code/data_structures/stack/reverse_stack**

blaze-nitd:
Reversed a stack without using an auxiliary stack and using only PUSH() and POP() operations. Applied stack STL. Fixes #1159


AdiChat:
This code with the same strategy exists ([see](https://github.com/OpenGenus/cosmos/blob/master/code/data_structures/stack/reverse_stack/reverse_stack_without_extra_space.cpp)) 👍

PR #1494, opened 2017-10-14
add power of 2 implementation in C#

Jiraphapa:
add power of 2 implementation in C# under bit-manipulation topic


AdiChat:
Kindly remove the file `power_of_2.cs` as the `C#` implementation exists 👍 

Your `Swift` implementation is good to be merged 👍

Issue #1315, opened 2017-10-12
Add Soft heap implementation


andreigasparovici:
More about this DS [here](https://en.wikipedia.org/wiki/Soft_heap) and [here](https://www.cs.princeton.edu/~chazelle/pubs/sheap.pdf)

JonNRb:
Hey this seems pretty cool. Mind if I pick this up for Hacktoberfest?

AdiChat:
Sure. Go for it 👍

sgrG24:
can i also complete this??

AdiChat:
Sure. Go for it 👍

JonNRb:
Updated PR: https://github.com/OpenGenus/cosmos/pull/1499

PR #1499, opened 2017-10-14
Add soft heap made up of binary heaps

JonNRb:


JonNRb:
https://github.com/OpenGenus/cosmos/issues/1315

code/data_structures/soft_heap/soft_heap.h
```diff
@@ -0,0 +1,259 @@
+/**
```
AdiChat: Kindly correct the extension of this code 👍

JonNRb: Sure, but a library of data structures as C++ source files are much more difficult to use in other projects than just a header.

JonNRb:
@AdiChat bump

code/data_structures/soft_heap/soft_heap.h
arnavb: @JonNRb AdiChat is correct, since you have defined a main function here, meaning this cannot be used as a library anyway.

JonNRb:
I am well aware, which is why there was a `#if 1` that you could easily
make `#if 0`. All of the data structures that people are writing for Cosmos
*absolutely* need tests, but we don’t have any packaging or build system to
define how to link things together.

code/data_structures/soft_heap/soft_heap.h
JonNRb: I am well aware, which is why there was a `#if 1` that you could easily make `#if 0`. All of the data structures that people are writing for Cosmos *absolutely* need tests, but we don’t have any packaging or build system to define how to link things together.

(github email replys don't seem to attach to the right thread...)

Since most code is C++, Java, and Python, we could be using something like https://bazel.build/ to add unit tests and let people leverage the code in this repository without just copying the raw files. I'd be happy to investigate something like this for the existing codebase.

AdiChat:
Looks great 😍 

PR #1527, opened 2017-10-14
Program to check basic properties of graph in python

nj2610:
I am adding a program that enables to find basic properties of graphs in python.
Please review and add it.


nj2610:
Have you reviewed the code??

PR #1530, opened 2017-10-14
Program to add two polynomial in C

nj2610:
I am adding a program to add two polynomial in C.


AdiChat:
Looks great 😍 

Issue #1380, opened 2017-10-13
add doubly linked list implementation in C


flavian11:
hi, i would like to implement a doubly linked list in C.

vshelke:
This is interesting. Can i complete this ?

AdiChat:
@vshelke Possibly, @flavian11 is working on it. You can browse through our code base to see scopes of contribution 👍

Rishabh570:
can i do this....?

AdiChat:
Sure @Rishabh570 Go for it 👍

rishav-pandey:
@AdiChat 
You just stopped to vshelke for doing this and welcomed to Rishabh70
What's the scenario??

AdiChat:
Oh. I did not notice that the issue is in C. Sorry for the confusion 👍 

@vshelke @Rishabh570 You can work on this in any language other than `C` 👍

Gupta-shifali:
Can I implement this in c++ @AdiChat?

yatharthshahjpr:
@AdiChat , Please review Create DoublyLinkedList.java #1539 ! 👍

rajatbajaj000:
@flavian11 can i work on this?

AdiChat:
Sure @rajatbajaj000 Go for it 👍

accakks:
@AdiChat Can I work on this using python?

AdiChat:
Sure @accakks Go for it 👍

Q-Mart:
@AdiChat Can I do this in Scala?

Q-Mart:
Added a scala implementation in #2322, feel free to accept/decline

anushree816:
Can i do this in C ?

abhijeet010304:
I would like to work on it in C++  and pyhton please assign it to me as a part of GSSoC 20

PR #1558, opened 2017-10-15
Doubly linked list in c++

Gupta-shifali:
Doubly linked list in c++ #1380

### DIFF
--- a/code/data_structures/linked_list/doubly_ll.cpp
+++ b/code/data_structures/linked_list/doubly_ll.cpp
@@ -1,0 +1,199 @@
+#include<iostream>
+#include<malloc.h>
+using namespace std;
+struct node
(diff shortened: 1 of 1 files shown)


AdiChat:
Doubly linked list in `C++` already exists ([see](https://github.com/OpenGenus/cosmos/blob/master/code/data_structures/linked_list/doubly_linked_list/Doubly_Linked_List.cpp)) 👍

Gupta-shifali:
Yes, but that's using templates. My code is without templates.

Gupta-shifali:
Isn't it ok @AdiChat ??

AdiChat:
I see but we add duplicate code when they approach the problem differently. This is more about the input it serves.

PR #1587, opened 2017-10-15
Improved binary_search implementation written in rust

martin-fink:
This commit uses some of the nice features rust has.

Instead of returning a negative value when no index is found, it now returns `Option::None`.
In addition, the numbers get passed as a slice instead of a borrowed `Vec`, which looks a bit cleaner in my eyes.


AdiChat:
Looks great 😍 

PR #1590, opened 2017-10-15
[Python] log of factorial

symc87:
Here is log of factorial in Python language. Please have a look at it.


AdiChat:
Looks great 😍 

PR #1597, opened 2017-10-15
Add bubble sort and quicksort in elm

dawehner:


AdiChat:
Looks great 😍 

dawehner:
Thank you for this super quick merge. I might do some other PR here and
then :)

PR #1613, opened 2017-10-15
Add Palindrome checker in sh

dnne184:
Added Palindrome checker in sh

### DIFF
--- a/code/string-algorithms/palindrome_checker/palindrome.sh
+++ b/code/string-algorithms/palindrome_checker/palindrome.sh
@@ -1,0 +1,27 @@
+#!/bin/sh
+
+# Check if a given string is palindrome
+palindrome()
+{
+    str=$1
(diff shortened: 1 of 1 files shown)


AdiChat:
Looks great 😍 

DarthFudd:
There's a more concise solution in bash:
`
palindrome_bash4(){
    for((i=0,j=$((${#1}-1));i<j;i++,j--))
    do [[ "${1:i:1}" = "${1:j:1}" ]]||return
    done
}
`
...but it doesn't work in plain old POSIX sh. I thought about checking in an alternate version, but if I name it "palindrome.sh", it will overwrite @dnne184's file, which works just fine in .sh (so long as 'rev' is installed). @AdiChat : Should I add the bash4 version as a comment, or create a separate file called "palindrome.bash" or somesuch?

AdiChat:
You can create a new file 👍

PR #1640, opened 2017-10-15
Adding bogo sort to ruby for #1635

xoorath:


arnavb:
This already exists, which is why there is a merge conflict.

xoorath:
@arnavb  Wasn't conflicting when I wrote this (right after the issue was posted).  No worries, can be closed.

PR #1643, opened 2017-10-15
Update README.md

vpatel95:


arnavb:
This PR title is somewhat misleading.

code/sorting/radix_sort/README.md
```diff
@@ -1,4 +1,39 @@
...
+
+**Example**
+
+Original, unsorted list:
+
+  ```170, 45, 75, 90, 802, 24, 2, 66```
+  
+Sorting by least significant digit (1s place) gives: **Notice that we keep 802 before 2, because 802 occurred before 2 in the original list, and similarly for pairs 170 & 90 and 45 & 75**
+
+  ```170, 90, 802, 2, 24, 45, 75, 66```
```
arnavb: no need for the triple grave accents, none or just one would suffice.

PR #1652, opened 2017-10-15
Create graham_scan.py

yatharthshahjpr:
Code for graham scan in python !

### DIFF
--- a/code/computational_geometry/graham_scan/graham_scan.py
+++ b/code/computational_geometry/graham_scan/graham_scan.py
@@ -1,0 +1,101 @@
+/* Part of Cosmos by OpenGenus Foundation */
+
+
+#!/usr/bin/env python3
+
+import functools
+
+@functools.total_ordering
+class Point:
+    """A point class for use in the Graham Scan."""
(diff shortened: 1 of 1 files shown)


code/computational_geometry/graham_scan/graham_scan.py
```diff
@@ -0,0 +1,101 @@
+/* Part of Cosmos by OpenGenus Foundation */
```
arnavb: Wrong comment type.

AdiChat:
It is sad that I found this code at some other repository (not yours) ([see source](https://github.com/samfoy/GrahamScan/blob/master/graham_scan.py)) which means you took it from there 👍 

This is the reason you got the header comment format wrong when comments have been used in the code 😕 No problem 👍 

*Kindly add support for user input or hardcoded input instead of using file input to this merged* 👍  

I reviewed 2 other codes from your pull requests and they are, also, using wrong algorithm. #1656 #1655 

Kindly justify this 👍 
This is no offence and we love you as we do everyone.

PR #1653, opened 2017-10-15
Add union-find data structure in python

hir12111:
It's based on a dictionary, that way it's possible to use any hashable type
on it


AdiChat:
Looks great 😍 

PR #1655, opened 2017-10-16
Create BoyerMoore.java

yatharthshahjpr:
Code for Boyer Moore in Java ! :)


AdiChat:
This code is, also, erroneous. 

In Boyer Moore algorithm, bad character heuristics is the central idea which I see nowhere. The code is using suffix array to find patterns.

AdiChat:
This code is, also, erroneous. 

In Boyer Moore algorithm, `bad character heuristics` is the central idea which I see nowhere. The code is using suffix array to find patterns.

PR #1664, opened 2017-10-16
Adding Count Digits function in go

jdalrymple:


code/mathematical-algorithms/count_digits/count_digits.go
```diff
@@ -0,0 +1,23 @@
+package main
+
+import "fmt"
+
+func countDigits(n int) int {
```
hwchiu: Could you fix the indent of this function?
Make all consistent.

jdalrymple: oops! Will do

PR #1669, opened 2017-10-16
Updated alignment for license

suhitmajumdar:
A good alignment improves readability of any document.


AdiChat:
The shift denotes the beginning of a paragraph and is a standard practice 👍

PR #1675, opened 2017-10-16
Create affine.c

ansh997:
The affine cipher is a type of monoalphabetic substitution cipher, wherein each letter of an alphabet is mapped to its numeric equivalent, encrypted using a simple mathematical function, and converted back to a letter. The formula used means that each letter encrypts to one other letter, and back again, meaning the cipher is essentially a standard substitution cipher with a rule governing which letter goes to which. As such, it has the weaknesses of all substitution ciphers. Each letter is enciphered with the function (ax + b) mod 26, where b is the magnitude of the shift.

### DIFF
--- a/code/cryptography/affine_cipher/affine.c
+++ b/code/cryptography/affine_cipher/affine.c
@@ -1,0 +1,80 @@
+//affine.c
+/* affine_cipher in c
+  * How it works: It uses modular arithmetic to transform the integer that 
+  each plain text letter corresponds to into another integer that correspond to a cipher text letter.*/
+  
+  #include<stdio.h>
+  #include<string.h>
+  
+  const int a=15;
+  const int b=17;
+  
+  
+  //Encryption algo
+  string encrypt(string msg)
+  {
+    string cipher='';
+    
+    for(int i=0;i<strlen(msg);i++)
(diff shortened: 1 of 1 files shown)


code/cryptography/affine_cipher/affine.c
```diff
@@ -0,0 +1,80 @@
+//affine.c
```
yatharthshahjpr: Please add comment  /* Part of Cosmos by OpenGenus Foundation */ at the top of every code you contribute to this project  ! 👍

arnavb:
This code does not compile at all, since this uses `string`, which exists in C++, not C. `char*` needs to be used instead.

ulvgard:
I was getting to that conclusion, just asking in a more defensive way :)
If this ends up as a C++ submition in the end, change the file ending to .cpp to make it more recognizable.

arnavb:
@ulvgard I think this will still be a C submission, since the rest of the code uses general C practices anyways.

hwchiu:
@arnavb It seems it uses c++ style string object.  maybe refactor to use char array?

arnavb:
@hwchiu Yes, that is what I said before.

PR #1685, opened 2017-10-16
made change

pankajobjects:


AdiChat:
The purpose is not clear and moreover, the change is not needed 👍

PR #1701, opened 2017-10-16
 Add min cost path in java

rht18:
Added code for minimum cost calculation in java


AdiChat:
Looks great 😍 

PR #1709, opened 2017-10-16
Update count_digits.py

RithvikKasarla:
I made it so the user can input a number instead of it hardcoded in.


AdiChat:
Looks good 👍  

PR #1726, opened 2017-10-16
Fix polygon area

Phil9l:
Since we remove `abs` from the triangle area calculation and use the signed area, this method also works for non-convex polygons.
For example, for polygon
```
0 0
20 20
40 0
20 10
```
the old code returned 600, but the new one returns 200 (correct value, can be checked like `(20 * 0 - 40 * 20) / 2`).

### DIFF
--- a/code/computational_geometry/area_of_polygon/area_of_polygon.cpp
+++ b/code/computational_geometry/area_of_polygon/area_of_polygon.cpp
@@ -1,3 +1,4 @@
+#include <cmath>
 #include <iostream>
 typedef std::pair<double, double> point;
 
@@ -9,7 +10,7 @@ point points[size];
 #define y second
 
 double calcArea(point a, point b, point c) {
-  return abs( (a.first - b.first) * (c.second - b.second) - (a.second - b.second) * (c.first - b.first) ) / 2;
+  return ((a.first - b.first) * (c.second - b.second) - (a.second - b.second) * (c.first - b.first)) / 2;
 }
 
 int main () {
@@ -26,5 +27,5 @@ int main () {
     answer += calcArea(points[0], points[i - 1], points[i]);
   }
 
-  std::cout << answer;
+  std::cout << abs(answer);
 }


code/computational_geometry/area_of_polygon/area_of_polygon.cpp
```diff
@@ -1,3 +1,4 @@
+#include <cmath>
```
yatharthshahjpr: Please add open genus comment here ! 👍

AdiChat:
Looks great 😍 

AdiChat:
This change might lead to erroneous results 👍

arnavb:
@AdiChat I agree, the formula was made for a reason, so it should be followed.

AdiChat:
Yes, it computes area by breaking it into several parts some of which can be negative and does giving lower area. It is important to keep individual area positive 👍

Phil9l:
@AdiChat, well, I guess, you mean, we can get `-S` instead of `S`. But I am not sure, it is "erroneous", it is just a signed value of the area, which by the way, may be used to find the order of points (CW or CCW).
When I removed this `abs` function, I thought it can be counterintuitive, that the result value can be negative. I can implement few more functions such as `get_signed_area()`, `get_area()`, if it will be more acceptable.
But if you mean, we can get just a wrong value of the area (not about its sign), then I believe, we can't. It is one of the 2 default methods of the simple polygon area finding. Here are some proofs: [http://geomalgorithms.com/a01-_area.html](http://geomalgorithms.com/a01-_area.html) (Ctrl + F "2D Polygons").

code/computational_geometry/area_of_polygon/area_of_polygon.cpp
Phil9l: @yatharthshahjpr, fixed.

AdiChat:
Yes, orientation of points determines the result.
Consider this input:
```
4
0 0
1 0
0 1
1 1
```

In first iteration, points (`0, 0`) (`1, 0`) and (`0, 1`) are considered where `calcArea()` will return -0.5 and for the next set, it will return 0.5 👍 So, after the sum the area turns out to be 0.

If you place `abs` in the summation, it will be good again 👍

PR #1728, opened 2017-10-16
Add union find in javascript

xanecs:


code/data_structures/union_find/union_find.js
```diff
@@ -0,0 +1,55 @@
+class UnionFind {
```
yatharthshahjpr: Please add open genus comment here ! 👍

AdiChat: Kindly make a note it is optional and do not make such comments

AdiChat:
Looks great 😍 

Issue #806, opened 2017-10-07
Find the standard deviation of array of numbers 


win72:
Can I implement in C++?

AdiChat:
Sure. Go for it 👍

win72:
Done. PR #828

ntrinquier:
Done in python as well in #901 :)

bvwells:
Added in Go.

PR #1734, opened 2017-10-17
Standard Deviation function in JS

LalithD:
I implemented a standard deviation function in JavaScript. #806


AdiChat:
Looks great 😍 

Issue #1702, opened 2017-10-16
Add a standard build system to Cosmos


jonnrb:
With most of the code in this repository being written in compiled languages (Java, C++, C, and Go make up over 75% of the codebase) we should have a standard build system.

1. A standard build system lets us run CI on the codebase to make sure all PRs actually build before they are merged.
2. If we have a build system, we can build more complex data structures and algorithms on top of simpler ones that have already been implemented in Cosmos.
3. Most modern build systems let code be directly imported from github. This encourages people to use those features instead of exporting the raw files from this repo and not contributing their changes back.

I personally like https://bazel.build/ because it allows for reasonably self-contained units and supports most of the languages that people are using to write code for Cosmos. How do other people feel about this?

arnavb:
This would also allow the creation of a coding standard, since CI's allow that. I personally agree, but it is really up to AdiChat.

PR #1736, opened 2017-10-17
Add Bazel WORKSPACE and a single BUILD file

JonNRb:
The WORKSPACE file is global for Cosmos and lets Bazel know where the project root is. When Bazel builds the libraries in Cosmos, it creates some directories that start with "bazel-" in the project root, so I wildcard added these to `.gitignore`.

Just to try this out, I added a BUILD file for the red-black tree library that builds each language's implementation. Each library also has a unit test associated with it, to demonstrate how we can run
```bash
bazel test //...
```
and every test in Cosmos will be run.


JonNRb:
#1702

PR #1741, opened 2017-10-17
Adding JS implementation of greatest digit in number

schie:
implements 'greatest digit in number' in JS.


AdiChat:
Looks great 😍 

PR #1748, opened 2017-10-17
Reclassify 2Sum

chrlee:
Since the purpose of 2Sum is to find a+b=c, I believe it can be classified under mathematical-algorithms, as opposed to its current location in unclassified.


AdiChat:
There are 5 files in the concerned folder but only 4 seems to be moved. Kindly make appropriate change and we will get this merged 👍

chrlee:
@AdiChat Added the 5th file!

AdiChat:
Looks great 😍 

PR #1749, opened 2017-10-17
standard deviation for stl-like containers in c++

peter-hf:


code/mathematical-algorithms/std/std.hpp
```diff
@@ -0,0 +1,27 @@
+#include <cmath>
```
yatharthshahjpr: Please add open genus comment  ! 👍

AdiChat:
Looks great 😍 

PR #1761, opened 2017-10-17
Ternary_heap_cpp_implementation

Daman98:
Having implementation of ternary min heap with basic heap functions and also printing k smallest no.'s in heap while keeping the heap intact also have a function to print no.'s smaller from 'a' present in heap.

### DIFF
--- a/code/data_structures/heap/Ternary_Heap.cpp
+++ b/code/data_structures/heap/Ternary_Heap.cpp
@@ -1,0 +1,111 @@
+#include <bits/stdc++.h>
+using namespace std;
+
+void swap(int *a,int *b){
+	int temp=*a;
+	*a=*b;
+	*b=temp;
+}
+
+void minheapify(int H[],int i,int n){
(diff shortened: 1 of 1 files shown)


code/data_structures/heap/Ternary_Heap.cpp
```diff
@@ -0,0 +1,111 @@
+#include <bits/stdc++.h>
+using namespace std;
+
+void swap(int *a,int *b){
```
arnavb: Since this is C++, consider using pass by reference instead.

code/data_structures/heap/Ternary_Heap.cpp
```diff
@@ -0,0 +1,111 @@
...
+	int temp=*a;
+	*a=*b;
+	*b=temp;
+}
+
+void minheapify(int H[],int i,int n){
```
arnavb: Use `vector`s, not raw arrays for all these function parameters.

PR #1762, opened 2017-10-17
add: bloom filter python implementation

abhishekjiitr:
adds bloom filter python implementation, which is basically a space-efficient probabilistic data structure, that is used to test whether an element is a member of a set.(kind of Hash Table except deletion is not possible here)


abhishekjiitr:
@AdiChat pls review! :+1:

AdiChat:
Looks great 😍 

PR #1771, opened 2017-10-17
Create median_of_medians in C

abdouskamel:
Here is the median of medians algorithm. It's a deterministic select algorithm that executes in O(n) in the worst case.


AdiChat:
Looks great 😍 

PR #1777, opened 2017-10-17
better way to print array

euronautic:
Use Arrays.toString instead of verbose loop


AdiChat:
Looks good 👍  

PR #1786, opened 2017-10-17
add: bloom filter c++ implemetation

wilsonwen:
Support bloom filter of integer value.
More types would be added with proper hash function.


AdiChat:
Looks great 😍 

PR #1792, opened 2017-10-17
fix variable name for decode and irregular \n on alphabet

paoloo:


AdiChat:
Looks great 😍 

PR #1797, opened 2017-10-17
FCFS in C++

praharijner:
FCFS in C++ using pointers

Fixes issue #[Mention the issue number it fixes or add the details of the changes if it doesn't has a specific issue.]

Changes: [Add here what changes were made in this pull request.]

### DIFF
--- a/fcfs is c++
+++ b/fcfs is c++
@@ -1,0 +1,154 @@
+#include <iostream>
+using namespace std;
+
+//defines a node
+struct NODE
+{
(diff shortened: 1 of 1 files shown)


AdiChat:
The code in this pull request is an implementation of Queue even though it is claimed to be `FCFS` implementation 👍

divyam3897:
@praharijner why not use the STL library for queue instead of coding it from scratch in this?

PR #1823, opened 2017-10-17
Add Gaussian elimination

Phil9l:
Add implementation of Gaussian Elimination. In linear algebra, Gaussian elimination (also known as row reduction) is an algorithm for solving systems of linear equations.


AdiChat:
Looks great 😍 

PR #1825, opened 2017-10-17
Add Hamiltonian Path implementation

Phil9l:
Hamiltonian path (or traceable path) is a path in an undirected or directed graph that visits each vertex exactly once. This problem is NP-complete.
Use a random-based algorithm to optimize search time.


AdiChat:
Looks great 😍 

Issue #1784, opened 2017-10-17
Min size sub array with the same degree as the original array


karanchawla:
Given a non-empty array of non-negative integers nums, the degree of this array is defined as the maximum frequency of any one of its elements.

The task is to find the smallest possible length of a (contiguous) subarray of nums, that has the same degree as nums.

For e.g: 
```
Input: [1, 2, 2, 3, 1]
Output: 2
Explanation: 
The input array has a degree of 2 because both elements 1 and 2 appear twice.
Of the subarrays that have the same degree:
[1, 2, 2, 3, 1], [1, 2, 2, 3], [2, 2, 3, 1], [1, 2, 2], [2, 2, 3], [2, 2]
The shortest length is 2. So return 2.
```

Adding a PR for CPP version for this using unordered_map.

karanchawla:
Please review https://github.com/OpenGenus/cosmos/pull/1785

PR #1830, opened 2017-10-17
minSubarraySizeWithDegree

AugustoQueiroz:
**Fixes issue:** #1784

**Changes:**
Created a function that returns the size of the smallest subarray with the same degree as the original array.
The function uses a two-pointers algorithm to sweep through the array, looking for the subarrays that satisfy the requirement, and then returns the size of the smallest one.
Time Complexity: O(n)


code/unclassified/minSubarraySizeWithDegree/minSubarraySizeWithDegree.cpp
```diff
@@ -0,0 +1,66 @@
+#include <bits/stdc++.h>
+#include <unordered_map>
+
+using namespace std;
```
Vitorvgc: Don't forget to add the required comment as stated in the [contribution guidelines](https://github.com/OpenGenus/cosmos/wiki/contribute).

code/unclassified/minSubarraySizeWithDegree/minSubarraySizeWithDegree.cpp
```diff
@@ -0,0 +1,66 @@
+#include <bits/stdc++.h>
```
Vitorvgc: Is is not preferable to use this header as it is not supported by all compilers.

code/unclassified/minSubarraySizeWithDegree/minSubarraySizeWithDegree.cpp
```diff
@@ -0,0 +1,66 @@
...
+using namespace std;
+
+int minSubArraySizeWithDegree(const vector<int> nums) {
+	unordered_map <int, int> m;
+	priority_queue<int> pq;
+	int best_degree = 1;
+	int curr_best_degree = nums[0];
+	int i,
+		j,
+		s = nums.size();
```
Vitorvgc: I think it is better to put these lines aligned or in the same line.

code/unclassified/minSubarraySizeWithDegree/minSubarraySizeWithDegree.cpp
AugustoQueiroz: Ok, makes sense

AugustoQueiroz:
Your comments should be fixed now.

AdiChat:
Looks great 😍 

PR #1836, opened 2017-10-17
coprime_numbers: simplify go implementation

faye-sama:
- use tuple assignment instead of temp variable
- remove unnecessary check when calling gcd, as the function works
  correctly independently of which argument is larger


AdiChat:
Looks great 😍 

Issue #1115, opened 2017-10-10
Shuffle Array


Josh5231:
I just added a pull request for the javascript version of this. #1114

IAmBullsaw:
Made an implementation for shuffling arrays in ruby, please revise #1153 :smirk_cat:

Z-Fikar:
Can I use the Shuffle function from library (if it provided)? or I have to use some kind algorithms, like Fisher-Yates?

MarsMSJ:
I thought we were implementing the algorithms, not using code already in the standard library? ( shuffle_array.cpp )

Z-Fikar:
@MarsMSJ that is exactly the reason why I asked before.

PR #1849, opened 2017-10-18
Add shuffle randomizer in PHP language

adrianuf22:
#1115 

Implements the shuffle algorithm in PHP language. 

Note: PHP has a native method for shuffle, so I include in this script a shuffle for array with assoc. indexes.


AdiChat:
Looks great 😍 

PR #1867, opened 2017-10-18
Modified Coin change Problem in JAVA

ajstrider:
Modified Coin Change problem in JAVA.Using a Hashmap to speedup runtime.


AdiChat:
Coin change and lonely integer in `Java` exists. Kindly remove them and it will be good to be merged then 👍

PR #1877, opened 2017-10-18
Added min and max heap using javascript

brunocalou:
**Changes:**
This commit adds an implementation of Min and Max binary heap using javascript.


AdiChat:
Looks great 😍 

PR #1890, opened 2017-10-18
change: fixed memory leak, refactor variables, and add new functions

iattempt:
**Fixes issue:**
1. not has destructor, the memory will leak
2. some function return the pointer, it will probably change inner value

**Changes:**
1. add "destuctor and call new function: release" to solve issue 1.
2. add const qualifier and move functions to solve issue 2.
3. rename ugly variables

related #1759


AdiChat:
Looks great 😍 

PR #1893, opened 2017-10-18
In exponential power added python

RithvikKasarla:
**Changes:**
In exponential power I added python


code/mathematical-algorithms/exponentiation_power/exponents.py
```diff
@@ -0,0 +1,4 @@
+base = input("Give me the base")
```
arnavb: Since you are using python 2 for this code, use `raw_input` and cast it to an int, for catching all errors.

PR #1901, opened 2017-10-18
Refactor existing c++ implementation.

hwchiu:
**Changes:**
1.  Rename the file to c since there're no c++ style in the original file, except the "cout".
2. Support four direction to solve the problem. (Only Right/Down previous version)


AdiChat:
Looks great 😍 

PR #1778, opened 2017-10-17
change: implement quick sort using generalization in cpp

iattempt:

PR #1902, opened 2017-10-18
change: fixed POD partial specialization problem

iattempt:
**Fixes issue:** 
1. since we need to compare two element, and need use less<Type> to default,
    but function template cannot implement by partial specialization,
    so need to implement full specialization for all of POD.

**Changes:**
1. fix issue 1. by using iterator_traits to extract value type of iterator or pointer

related #1778


AdiChat:
Looks great 😍 

Issue #1788, opened 2017-10-17
create fcfs in any language


tbhaxor:
create code fof FCFS (first come first serve) algo in any language at **code/operating_system/first_come_first_serve**

@AdiChat  bro i am adding in `python3` and `csharp` in branch _OS_fcfs_

## What is FCFS?
_The first come, first served (commonly called FIFO ‒ first in, first out) process scheduling algorithm is the simplest process scheduling algorithm. It is rarely used in modern operating systems, but is sometimes used inside of other scheduling systems._

Abhishek2019:
is it fine if i provide java code for FCFS (First in First Out)?

tbhaxor:
I have a request plz include a function of printing gantt chart like i have modified in my PR 1903

Abhishek2019:
ya sure
i will change my java code and create pull again.
Thanks for suggestion

dowoncha:
I put in a request for this in Rust

PR #1903, opened 2017-10-18
modified fcfs.py 

tbhaxor:
#1788 

new feature added of gantt chart in` fcfs.py`  and used the real algo implementation

My request is to implement same code in all language


AdiChat:
Looks great 😍 

tbhaxor:
@AdiChat  ireq u to plz check if there is function that prints gantt chart or not 😄

PR #1907, opened 2017-10-18
change: implement linear search by generalization in cpp

iattempt:
**Fixes issue:**
1. the older version only for integer array

**Changes:**
1. fix issue 1. by using generalized implement,
this new version can run on every data structure,
if it is based on "Input Iterator" that defined in STL.


AdiChat:
Looks great 😍 

PR #1908, opened 2017-10-18
A more pythonicway to swap elements

euronautic:
swap of elements is now more pythonic


AdiChat:
Looks great 😍 

Issue #1880, opened 2017-10-18
Adding knight tour in c.


ReimuA:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

I'd like to add my C version of the knight tour (backtracking) as it doesn't exist yet.

PR #1911, opened 2017-10-18
Knight Tour in C (no backtracking).

ReimuA:
**Fixes issue:** #1880 

**Changes:**
So, it turns out someone else did the knight tour in c before i finished it ;_;.
But,  in the end, i didn't use backtracking to solve this problem (it took way too much time) and i used Warnsdorf's rule.
I think this approach is worth being in parrallele to the one existing, but it don't really fit the backtracking folder.


AdiChat:
Looks great 😍 

PR #1914, opened 2017-10-18
Convert all line endings to UNIX format

tudurom:
**Fixes issue:** Git likes it more like so. Users can configure git to use CRLF on
windows platforms.

**Changes:**
Converted all line endings from CRLF (windows) to LF (unix).


tudurom:
the other pr had all the changes in master, oops

AdiChat:
Looks great 😍 

PR #1919, opened 2017-10-18
change: implement binary search by generalization in cpp

iattempt:
**Fixes issue:**
1. some variables' name is ugly
2. not very convenient for user-defind or build-in container of STL

**Changes:**
1. fix issue 1. refactory function and variable
2. fix issue 2. replace pointer with iterator what use more data-insensitive to implement

note:
there are two version(recursive and iterative) in this file,
it can be change the "trait" to run another.


AdiChat:
Looks great 😍 

PR #1922, opened 2017-10-18
Adding Better Solution for JS number reversal

schie:
**Fixes issue:**
Adds ability to reverse decimal numbers in JS implementation of 'reverse number'

**Changes:**
* adds 'cleaner' implementation of reversal
* allows reversal of non-integer numbers
* adds additional console log


AdiChat:
Looks great 😍 

PR #1948, opened 2017-10-19
Fix generics

thefifthisa:

**Changes:**

Hi! Noticed that the insertNth method takes in a data parameter of type int, but since this is a generic class, I think it's more appropriate for it to be of a generic type T.


AdiChat:
Looks great 😍 

Issue #1962, opened 2017-10-19
Add diameter of trees algorithm


bhaargav006:

iattempt:
I’ll try to implement an cpp version.

AdiChat:
We have a directory for [diameter of binary tree](https://github.com/OpenGenus/cosmos/tree/master/code/data_structures/tree/binary_tree_diameter) under `data structures/ tree`. We may need to move it to `graph_algorithms` 👍

shreyneil:
I have made a pull request(https://github.com/OpenGenus/cosmos/pull/1982) with regard to this issue please go through it.

Issue #1915, opened 2017-10-18
Stop Using <bits/stdc++.h> for Portability


arnavb:
**This is a(n):**

- [ ] New algorithm
- [ ] Update to an existing algorithm
- [x] Error
- [x] Proposal to the Repository

**Details:**

I was googling up about the header file `<bits/stdc++.h>`, and found that it is consider generally bad practice to use it, because it is nonportable (It only works on g++, not MSVC or XCode).

1. ([Source](https://stackoverflow.com/a/31816096/6525260))
2. ([Source](https://stackoverflow.com/a/33764128/6525260))

Thus, I vote not using this header file and instead directly including files from the standard.

tbhaxor:
@arnavb yes its a bad practice coz it binds all header files and std library

arnavb:
I will open a PR soon that will attempt to solve this. Also, I think we should mandate any new PRs to NOT use the header file, but to directly include the STL files.

PR #1963, opened 2017-10-19
added diameter of trees

bhaargav006:
fixed issue #1962

### DIFF
--- a/code/graph-algorithms/diameter_trees/cpp/Diameter.cpp
+++ b/code/graph-algorithms/diameter_trees/cpp/Diameter.cpp
@@ -1,0 +1,65 @@
+//Learn O(n)
+
+
+// The following is O(n^2)
+#include<bits/stdc++.h>
+#define ll long long
+using namespace std;
+
+typedef struct tree node;
+
+struct tree
+{
+	ll data;
+	node * right;
+	node * left;
+};
+
+node * insert(ll a)
+{
+	tree * temp = new tree;
+	temp->data=a;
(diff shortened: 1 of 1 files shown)


code/graph-algorithms/diameter_trees/cpp/Diameter.cpp
```diff
@@ -0,0 +1,65 @@
+//Learn O(n)
```
Vitorvgc: Don't forget to add the required comment as stated in the [contribution guidelines](https://github.com/OpenGenus/cosmos/wiki/contribute).

code/graph-algorithms/diameter_trees/cpp/Diameter.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+
+// The following is O(n^2)
+#include<bits/stdc++.h>
```
Vitorvgc: It is not preferable to use this header as it is not supported by all compilers (#1915).

code/graph-algorithms/diameter_trees/cpp/Diameter.cpp
```diff
@@ -0,0 +1,65 @@
...
+typedef struct tree node;
+
+struct tree
+{
+	ll data;
+	node * right;
+	node * left;
+};
+
+node * insert(ll a)
```
Vitorvgc: This function name seems a bit misleading as it actually creates a new node rather than inserting it in something.

PR #1964, opened 2017-10-19
Maybe you missed my name

anmol27katyani:
I added my name.
I am also a contributor and i also want my name to be here.
Please


AdiChat:
Contributors are supposed to place their names on the list on their own ❤️ 

Kindly fix the format of the list in this pull request. It will be good to merge then 👍

PR #1986, opened 2017-10-19
Added FastFactorialUsingStack.java

kanikasaini:
Allows speedy calculation of factorial.

### DIFF
--- a/code/mathematical-algorithms/factorial/FastFactorialUsingStack.java
+++ b/code/mathematical-algorithms/factorial/FastFactorialUsingStack.java
@@ -1,0 +1,135 @@
+import java.util.*;
+import java.io.InputStreamReader;
+import java.io.BufferedReader;
+/*
+	@author Kanika Saini (https://github.com/kanikasaini)
+*/
+class mystack implements Cloneable //stack class
+{
+	public int[] s=new int[36000];
+	public int top;
+	public mystack()
+	{
+		top=-1;		
+	}	
+	public void push(int x)
+	{
+		top+=1;
+		s[top]=x;
+	}	
+	public int pop()
+	{
+		int temp=s[top];
+		top-=1;
+		return temp;
+	}
+	int seek()
+	{
+		return s[top];
+	}
+	public void display()
+	{
+		while(top!=-1)
+			System.out.print(this.pop());
+			System.out.println();
+	}
+	
+	public Object clone() throws CloneNotSupportedException 
+	{
+		return super.clone();
+	}
+}
+class FastFactorialUsingStack
+{
+public static void main(String[] args) throws Exception
+{
+
+    BufferedReader inp = new BufferedReader (new InputStreamReader(System.in));
+	int n= Integer.parseInt(inp.readLine()); //input n
+	long current= System.currentTimeMillis();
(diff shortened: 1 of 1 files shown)


code/mathematical-algorithms/factorial/FastFactorialUsingStack.java
```diff
@@ -0,0 +1,135 @@
...
+	}
+	
+	public Object clone() throws CloneNotSupportedException 
+	{
+		return super.clone();
+	}
+}
+class FastFactorialUsingStack
+{
+public static void main(String[] args) throws Exception
```
Croolis: Could you please fix indentation?

code/mathematical-algorithms/factorial/FastFactorialUsingStack.java
```diff
@@ -0,0 +1,135 @@
...
+			l=(int)p/10;
+		}
+		if (l>0)	
+			prod.push(l);	
+		temp= add(temp, prod);
+		i+=1;
+	}
+	return temp;
+}
+public static mystack reverse(mystack a)
```
Croolis: Are you sure you need to use this reverse method? What do you think about using double-ended queue instead of stack so we can avoid using reverse?

code/mathematical-algorithms/factorial/FastFactorialUsingStack.java
```diff
@@ -0,0 +1,135 @@
...
+	}
+	int seek()
+	{
+		return s[top];
+	}
+	public void display()
+	{
+		while(top!=-1)
+			System.out.print(this.pop());
+			System.out.println();
```
Croolis: Wrong indent, please fix

code/mathematical-algorithms/factorial/FastFactorialUsingStack.java
```diff
@@ -0,0 +1,135 @@
+import java.util.*;
+import java.io.InputStreamReader;
+import java.io.BufferedReader;
+/*
+	@author Kanika Saini (https://github.com/kanikasaini)
+*/
+class mystack implements Cloneable //stack class
```
Croolis: Class name violates naming convention. It should be MyStack (or even better Stack) instead.

Croolis: Also are you sure you need your own implementation of data structure? Why not use java.util.Stack?

moonpatroller: You shouldn't implement Cloneable.

code/mathematical-algorithms/factorial/FastFactorialUsingStack.java
```diff
@@ -0,0 +1,135 @@
...
+{
+	mystack sum= new mystack();
+	int t, c=0;
+	a1 = reverse(a1);
+	b1= reverse(b1);
+	while(a1.top!=-1 && b1.top!=-1)
+	{
+		t=a1.pop() + b1.pop() + c;
+		sum.push(t%10);
+		c= (int)(t/10);
```
moonpatroller: Don't cast, this is a noop.  An int divided by an int is already an int.

PR #1987, opened 2017-10-19
c++ code for Centroid Decomposition of a Tree.

spj29:
Centroid decomposition of a tree is a very powerful divide and conquer technique used to solve problems based on tree data structures.
A centroid is a node which when removed from the tree decomposes the tree into components each with a size less than or equal to half of the initial size of the tree.
Thus, building this tree recursively on each component leads to a powerful technique which can perform sevaral tasks in logarithmic time.

I present a very short and easy to grasp code for the same.


arnavb:
Remove the space in the  folder name.

PR #1997, opened 2017-10-19
Create README.md for Decision Tree algorithm

genome21:
**Changes:**
Created a README file for the Decision Tree algorithm for explanation and possible usages.


arnavb:
This just seems a little wordy, otherwise, as far as content goes, it is good. 👍

PR #2009, opened 2017-10-19
Optimized code and Python Support

lordozb:
Added Support for Python 2.X and Python 3.x

**Changes:**


code/mathematical-algorithms/Check_is_square/Finding_Square.py
```diff
@@ -1,21 +1,40 @@
...
+		print 'No'
+
 else:
-    print(number, "is a perfect square")
+	print('Enter a positive number :')
+	n = input()
+	if(checkIfSquare(n)):
+		print 'Yes'
+	else:
+		print 'No'
```
Vitorvgc: It is missing parentheses in these prints for Python 3

lordozb: On it

PR #2011, opened 2017-10-19
Init Divided Differences

CaramelHeaven:
Divided differences. More: [https://en.wikipedia.org/wiki/Divided_differences](https://en.wikipedia.org/wiki/Divided_differences)


AdiChat:
Looks great 😍 

PR #2012, opened 2017-10-19
C++ code for centroid decomposition on a tree

spj29:
Centroid decomposition of a tree is a very powerful divide and conquer technique used to solve problems based on tree data structures.
A centroid is a node which when removed from the tree decomposes the tree into components each with a size less than or equal to half of the initial size of the tree.
Thus, building this tree recursively on each component leads to a powerful technique which can perform sevaral tasks in logarithmic time.

I present a very short and easy to grasp code for the same.

### DIFF
--- a/code/divide_conquer/Centroid_decomposition/Centroid_deco_on_a_tree.cpp
+++ b/code/divide_conquer/Centroid_decomposition/Centroid_deco_on_a_tree.cpp
@@ -1,0 +1,95 @@
+// spj_29's Template.
+/***********************************************
+    Code for centroid decomposition of a tree.
+    overall complexity is nlogn.
+************************************************/
+#include <bits/stdc++.h>
+ 
+using namespace std;
+ 
+#define SPEED ios::sync_with_stdio(false); cin.tie(0); cout.tie(0)
+#define fileio freopen("in.in", "r", stdin),freopen("out.out", "w", stdout);
+#define ll long long int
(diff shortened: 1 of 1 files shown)


code/divide_conquer/Centroid_decomposition/Centroid_deco_on_a_tree.cpp
```diff
@@ -0,0 +1,95 @@
...
+    //mark node as visited so that it is not processed in components it does not belong to.
+    visited[j]=1;
+    for(auto i:adj[j])
+        if(!visited[i])
+            //get centroid of subtree and put its parent to be the current centroid (j).
+            cenpar[getcen(i)]=j;
+    return j;
+}
+
+int main() {
```
arnavb: Fix indentation in this function.

code/divide_conquer/Centroid_decomposition/Centroid_deco_on_a_tree.cpp
```diff
@@ -0,0 +1,95 @@
+// spj_29's Template.
+/***********************************************
+    Code for centroid decomposition of a tree.
+    overall complexity is nlogn.
+************************************************/
+#include <bits/stdc++.h>
```
arnavb: Don't include this. See #1915

KtlTheBest:
@spj29 Can I also help?

PR #2041, opened 2017-10-20
fix bug in comb sort C++

pbnjay:
fixes an out-of-bounds error when gap=0


AdiChat:
Looks great 😍 

PR #2046, opened 2017-10-20
Add AES-128 in python

debeltranc:
**Changes:**
Add AES-128 implementation using python and its builtin libraries only.


AdiChat:
Looks great 😍 

PR #2083, opened 2017-10-20
update singly_linked_list.cpp to use all features of cpp.

dparikh85:
Use of classes can be done rather than still using struct.

**Changes:**
Classes are used rather than struct and "using namespace std" is added.


InfiniteCoder:
It's good idea to use classes.
But using `using namespace std` is not recommended. Read [this post on SO](https://stackoverflow.com/q/1452721)

dparikh85:
Okay  Thanks.

‌

AdiChat:
Looks great 😍 

dparikh85:
Yes☺️

PR #2096, opened 2017-10-20
Update Sum_of_digits_function.cpp

Sanchit-20:
Updated the range of n from int to long long int. Now sum of digits of long long int number can also be obtained.

**Changes:**
Updated the range of input number from int to long long int. Now, Sum of digits of large number upto 10^18 can also be obtained.


AdiChat:
Looks great 😍 

PR #1591, opened 2017-10-15
add: implement hash table in cpp

iattempt:

PR #2099, opened 2017-10-20
change: generalize contains type and hash function, and add common functions

iattempt:
**Fixes issue:**
1. old version only can storing integer type
2. can't customize hash function

**Changes:**
1. generalize contains type
2. generalize hash function by passing function object
3. add more common function for insert/erase/find
4. and add some usable function: size/empty 

refer to #1591


AdiChat:
Looks great 😍 

PR #2113, opened 2017-10-20
median.cpp

dparikh85:
file in cpp format with use of stl.

### DIFF
--- a/code/dynamic_programming/Array_Median/median.cpp
+++ b/code/dynamic_programming/Array_Median/median.cpp
@@ -1,0 +1,21 @@
+#include<bits/stdc++.h>
+using namespace std;
+
+int main()
+{
+    int n,median;
+    cout<<"Enter number of elements\n";
+    cin>>n;
+    vector<int> arr(n);
+    for(int i=0;i<n;i++){
+        cin>>arr[i];
+    }
+    sort(arr.begin(),arr.end());
+    if(arr.size()%2==0){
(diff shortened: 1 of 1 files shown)


code/dynamic_programming/Array_Median/median.cpp
```diff
@@ -0,0 +1,21 @@
+#include<bits/stdc++.h>
```
arnavb: Don't include this. See #1915.

code/dynamic_programming/Array_Median/median.cpp
```diff
@@ -0,0 +1,21 @@
...
+using namespace std;
+
+int main()
+{
+    int n,median;
+    cout<<"Enter number of elements\n";
+    cin>>n;
+    vector<int> arr(n);
+    for(int i=0;i<n;i++){
```
arnavb: If using vectors, don't ask for size, and allow user to enter elements until a EOF. (Use `vector::push_back`)

code/dynamic_programming/Array_Median/median.cpp
```diff
@@ -0,0 +1,21 @@
...
+        cin>>arr[i];
+    }
+    sort(arr.begin(),arr.end());
+    if(arr.size()%2==0){
+        median=arr[arr.size()/2]+arr[(arr.size()/2)-1]/2;
+    }
```
arnavb: Either don't use braces here, or use braces below. (I prefer the latter)

PR #2117, opened 2017-10-20
Normalize a desire output to sigmoid range.

au12113:
**Fixes issue:**  if we use 0 or 1, data will be out-ranged, because sigmoid range is (0, 1). 

**Changes:**  so I recommend to put some error to a desire output for normalizing data on sigmoid range. 

sorry for my bad English.


AdiChat:
Looks great 😍 

PR #2129, opened 2017-10-20
Birthday paradox problem

thesagarpasrija:
c++ implementation of Birthday Paradox problem

### DIFF
--- a/code/unclassified/birthday_paradox_problem/bday_paradox.cpp
+++ b/code/unclassified/birthday_paradox_problem/bday_paradox.cpp
@@ -1,0 +1,23 @@
+#include<iostream>
+// Part of Cosmos by OpenGenus Foundation
+using namespace std;
+
+int main(){
+    float res=1;
+    int n=0;
+
+    float numerator=365;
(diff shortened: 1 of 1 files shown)


code/unclassified/birthday_paradox_problem/bday_paradox.cpp
```diff
@@ -0,0 +1,23 @@
+#include<iostream>
+// Part of Cosmos by OpenGenus Foundation
+using namespace std;
+
+int main(){
+    float res=1;
+    int n=0;
+
+    float numerator=365;
+    int denominator=365;
```
ulvgard: Can be made `const`

code/unclassified/birthday_paradox_problem/bday_paradox.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+    float prob;
+    cout<<"Enter probability: ";
+    cin>>prob;
```
ulvgard: Check the input range. Program runs forever if input is > 1.

PR #2132, opened 2017-10-20
Create factorial.cpp

mintoo511:
C++ Program to Find Factorial of a Number using Dynamic Programming.

**Changes:**

### DIFF
--- a/code/dynamic_programming/factorial/factorial.cpp
+++ b/code/dynamic_programming/factorial/factorial.cpp
@@ -1,0 +1,31 @@
+/* 
+ * C++ Program to Find Factorial of a Number using Dynamic Programming 
+ */
+#include <iostream>
+#define ll long long
+using namespace std;
+ 
+ll result[50] = {0};
+/* 
+ * Find Factorial of a Number using Dynamic Programming 
+ */
+ll Factorial(int n)
+{
+    if (n >= 0) 
+    {
+        result[0] = 1;
+        for (int i = 1; i <= n; ++i) 
+        {
+            result[i] = i * result[i - 1];
+        }
+        return result[n];
+    }
+}
+
+int main()
+{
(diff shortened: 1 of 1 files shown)


code/dynamic_programming/factorial/factorial.cpp
```diff
@@ -0,0 +1,31 @@
+/* 
+ * C++ Program to Find Factorial of a Number using Dynamic Programming 
+ */
+#include <iostream>
+#define ll long long
```
ulvgard: This preprocessor macro is not used in the preprocessing phase. Also, the short name `ll` does not improve readability for me. If you like to rename types, checkout C++11 type aliases.

code/dynamic_programming/factorial/factorial.cpp
```diff
@@ -0,0 +1,31 @@
...
+using namespace std;
+ 
+ll result[50] = {0};
+/* 
+ * Find Factorial of a Number using Dynamic Programming 
+ */
+ll Factorial(int n)
```
ulvgard: The return type is somewhat arbitrary. Since the type argument is `int`, why not return an `int` or vice versa for long?

code/dynamic_programming/factorial/factorial.cpp
```diff
@@ -0,0 +1,31 @@
...
+    if (n >= 0) 
+    {
+        result[0] = 1;
+        for (int i = 1; i <= n; ++i) 
+        {
+            result[i] = i * result[i - 1];
+        }
+        return result[n];
+    }
+}
```
ulvgard: Compiling with Clang 4.0.0 and `-Werror` yields:
```
error: control may reach end of non-void function
     [-Werror,-Wreturn-type]
}
```

PR #2133, opened 2017-10-20
russian peasant in csharp

tbhaxor:


code/mathematical-algorithms/russian_peasant_multiplication/russian_peasant.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    class russ_peasant
+    {
+        int a, b;
+        public russ_peasant(int number1,int number2)
+        {
+            a = number1;
+            b = number2;
+        }
+        public int multiply()
```
arnavb: Instead of having to make a new object for every new pair of numbers, consider making a static method called multiply that takes parameters a and b instead, and remove the private variables a and b along with the constructor.

tbhaxor: making function would be better then 😕

arnavb: Not necessarily. It just seems extra to have to create a new object every time.

tbhaxor: @arnavb  ohk bro i am changing to static then 😄

tbhaxor:
@arnavb i have made some changes 👍

AdiChat:
Looks great 😍 

Issue #2143, opened 2017-10-20
Very easy fix: Change Java class names and match the file names


AliKarpuzoglu:
**This is a(n):**

- [ ] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x ] Proposal to the Repository

**Details:**
Java classes should all have names starting with a capital letter. They should also have matching file names otherwise they won't compile. Very easy to fix

edgardmota:
@AliKarpuzoglu , I'll do this one!

PR #2148, opened 2017-10-20
Match Java classes to their filenames.

arwinneil:
Java class name must match filename

**Fix for issue:** #2143


arwinneil:
Closed because PRs do not follow good Java practice where class start with capital letters.

PR #2149, opened 2017-10-20
created bfs & dfs in C

InfiniteCoder:
created a bfs and a dfs program in C


InfiniteCoder:
@arnavb Is there a tool to automatically reformat code into KNF form? It would be better to use such a tool rather than manually reformatting all the programs.

arnavb:
Not yet, since the style guide is still in beta, you will have to do it yourself. What I said was really just a suggestion, since the guides are rapidly changing. You can choose to make the changes if you want. It will just be easier to maintain this in the future if you do.

AdiChat:
Looks great 😍 

PR #2169, opened 2017-10-20
add: Implement aa tree in cpp

iattempt:
**Fixes issue:**
#2170 

**Changes:**
1. create aa tree prototype, I will complete it tomorrow.

### DIFF
--- a/code/data_structures/aa_tree/aa_tree.cpp
+++ b/code/data_structures/aa_tree/aa_tree.cpp
@@ -101,7 +101,7 @@ public:
     typedef ptrdiff_t         difference_type;
     typedef size_t            size_type;
 
-    binary_tree(node_type *r = nullptr) :_root(r), _sz(0), _comp(_Comp()) {};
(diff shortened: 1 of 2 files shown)


code/data_structures/aa_tree/aa_tree.cpp
```diff
@@ -0,0 +1,246 @@
...
+    node_type const *find(const_reference value) {
+        
+    }
+
+private:
+};
+
+
+// for test
+#include <iostream>
```
arnavb: Comment out the test.

iattempt: not complete~  😅

iattempt:
@AdiChat
thanks for uniform code style

but, I'm sorry for scope of code style, some type or other need ordered definition, it is difficult to implement.
then, I use a viable alternative: still use public->private->protected, but may have multi times, like:
```
public:
  ...
protected:
  ...
public:
  ...
private:
  ...
```

could we use this way or other recommend?

arnavb:
Sure, sure, Just make sure that the main public interface comes first.

AdiChat:
Looks great 😍 

PR #2173, opened 2017-10-20
Updated BucketSort.java

azkarmoulana:


code/sorting/bucket_sort/BucketSort.java
```diff
@@ -3,49 +3,49 @@
 import java.util.Collections;
 
 public class BucketSort {
+    private static final int DEFAULT_BUCKET_SIZE = 5;
```
arnavb: You could make this public.

PR #2182, opened 2017-10-20
check if power of 2 in julia

ishikasoni:
**Fixes issue:** #1297

**Changes:** Added Julia code


code/bit-manipulation/power_of_2/power_of_2.jl
```diff
@@ -0,0 +1,8 @@
+print("Enter a number if it is a power of 2: ")
```
arnavb: How about: "Enter a number to check if it is a power of 2: "

AdiChat:
Looks great 😍 

PR #2194, opened 2017-10-20
Create simple-julia.cpp

jasonu:
This is a simple code to generate Julia sets of quadratic functions such as f(z) = z^2 + c. The algorithm used is the backwards iteration algorithm with each root weighted equally.  (Note:  This algorithm often misses a good portion of the Julia set.)


AdiChat:
Looks great 😍 

PR #2198, opened 2017-10-21
heap.py

Tikam02:
heap.py module for python's prims mst.

**Changes:**


code/greedy-algorithms/prim_minimum_spanning_tree/heap.py
```diff
@@ -0,0 +1,162 @@
+class Heap:
```
yatharthshahjpr: Please add open genus comment here ! 👍

PR #2223, opened 2017-10-21
Create FlashSortAlgorithm.java

yatharthshahjpr:
Code for Flash Sort Algorithm using Java !


moonpatroller:
Can you remove all the Deprecated methods?  They don't really make any sense, and you shouldn't be writing *new* code with deprecated methods.

Issue #712, opened 2017-10-06
Top View of Binary Tree


nimit95:
I would like to implement in C++

AdiChat:
Sure, Go for it 👍 

Be sure to check existing implementations 👍

anku580:
i would like to work in c

AdiChat:
Sure. Go for it 👍

ananyasethi56:
i would like to take this issue in java

AdiChat:
Sure @ananyasethi56 Go for it

ananyasethi56:
i want to do in c++

AdiChat:
Sure. Go for it 👍

Amit030:
is this closed or still something  needs to be done in this particular "Top View of a binary tree in C++ " topic

PR #2246, opened 2017-10-21
Top View of a binary tree in C++

ramini1996:
**Fixes issue:** #712

**Changes:**
Implemented complete code in C++

### DIFF
--- a/code/data_structures/tree_topview.cpp
+++ b/code/data_structures/tree_topview.cpp
@@ -1,0 +1,62 @@
+#include<bits/stdc++.h>
+using namespace std;
+
+struct node
+{
+	int data;
+	struct node* left;
+	struct node* right;	
+};
+
+struct node* insertNode(int data)
+{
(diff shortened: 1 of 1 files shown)


arnavb:
Same problems as your other PRs, other than the file extension problem.

code/data_structures/tree_topview.cpp
```diff
@@ -0,0 +1,62 @@
...
+		m[val]=root->data;
+		if(root->left!=NULL)
+		q.push(make_pair(root->left,val-1));
+		if(root->right!=NULL)
+		q.push(make_pair(root->right,val+1));
+	}
+	map<int,int> :: iterator it;
+	for(it=m.begin();it!=m.end();it++)
+	{
+		printf("%d ",m[it->first]);
```
ulvgard: Why use printf when you include `<iostream>`?
This program does not compile using Clang 4.0.0 because printf is in cstdio

Issue #692, opened 2017-10-06
Right View of Binary tree


piyush6348:
I would like to implement this in c++

AdiChat:
Sure. Go for it 👍

anku580:
i would like to do in c++

AdiChat:
Sure @anku580 Go for it 👍

anku580:
i was trying to create a seperate pull request but it merges in one pull request how do i resolve this issue.

https://github.com/OpenGenus/cosmos/compare/master...anku580:master

Abishek10:
I have implemented it in Python. Please review #884

PR #2253, opened 2017-10-21
Right View of Binary Tree in C++

ramini1996:
**Fixes issue:** #692

**Changes:**
Implemented the complete code in C++

### DIFF
--- a/tree_rightview
+++ b/tree_rightview
@@ -1,0 +1,61 @@
+#include<bits/stdc++.h>
+using namespace std;
+
+struct node
+{
+	int data;
+	struct node* left;
+	struct node* right;	
+};
+
+struct node* insertNode(int data)
+{
+	struct node* temp=(struct node*)malloc(sizeof(struct node));
+	temp->data=data;
+	temp->left=NULL;
+	temp->right=NULL;
+	return temp;
+}
+
+void rightView(struct node* root)
+{
+	if(root==NULL)
+	return;
+	queue<node*> q;
+	q.push(root);
+	while(1)
+	{
+		int nodeCount=q.size();
+		if(nodeCount==0)
+		break;
+		while(nodeCount--)
+		{
+			struct node* root=q.front();
+			q.pop();
+			if(nodeCount==0)
+			{
+				printf("%d ",root->data);
+			}
+			if(root->left!=NULL)
+			q.push(root->left);
+			if(root->right!=NULL)
+			q.push(root->right);
+		}
+	}
+}
+
+int main()
+{
+	struct node* root=insertNode(1);
+	root->left=insertNode(2);
+	root->right=insertNode(3);
+	root->right->left=insertNode(8);
+	root->right->right=insertNode(9);
+	root->right->right->left=insertNode(10);
+	root->left->left=insertNode(4);
+	root->left->right=insertNode(5);
+	root->left->left->left=insertNode(6);
+	root->left->right->right=insertNode(7);
+	rightView(root);
+	return 0;	
+}


arnavb:
Please check out the [C++ Style Guide](https://github.com/OpenGenus/cosmos/blob/master/guides/coding_style/c%2B%2B/style.md) before submitting to the repository!

tree_rightview
```diff
@@ -0,0 +1,61 @@
+#include<bits/stdc++.h>
```
arnavb: Don't include this. See #1915.

tree_rightview
```diff
@@ -0,0 +1,61 @@
...
+using namespace std;
+
```
arnavb: This file needs a file extension and needs to be moved to the correct location.

tree_rightview
```diff
@@ -0,0 +1,61 @@
...
+struct node
```
arnavb: Since you are using C++, consider making a class instead, with member functions instead of standalone functions like this, which is the C style way of doing this. This will create an easier interface that is pointer free for the user.

tree_rightview
```diff
@@ -0,0 +1,61 @@
...
+			q.push(root->left);
+			if(root->right!=NULL)
+			q.push(root->right);
+		}
+	}
+}
+
+int main()
+{
+	struct node* root=insertNode(1);
```
arnavb: No need to prefix with struct in C++.

tree_rightview
```diff
@@ -0,0 +1,61 @@
...
+struct node
+{
+	int data;
+	struct node* left;
+	struct node* right;	
+};
+
+struct node* insertNode(int data)
+{
+	struct node* temp=(struct node*)malloc(sizeof(struct node));
```
arnavb: new operator exists in C++. Also, no need for `struct` here and anywhere else (with the function return values).

tree_rightview
```diff
@@ -0,0 +1,61 @@
...
+	temp->data=data;
+	temp->left=NULL;
+	temp->right=NULL;
```
arnavb: C++ NULL == nullptr. Use nullptr instead.

tree_rightview
ramini1996: Updated all the codes as mentioned

tree_rightview.cpp
```diff
@@ -0,0 +1,62 @@
...
+{
+	if(root==NULL)
+	return;
+	queue<node*> q;
+	q.push(root);
+	while(1)
+	{
+		int nodeCount=q.size();
+		if(nodeCount==0)
+		break;
```
arnavb: Fix indentation

tree_rightview.cpp
```diff
@@ -0,0 +1,62 @@
...
+		while(nodeCount--)
+		{
+			node* root=q.front();
+			q.pop();
+			if(nodeCount==0)
+			{
+				printf("%d ",root->data);
```
arnavb: printf -> cout

ramini1996: cout is slow compared to printf (same as #include<bits/stdc++.h>)

ramini1996: Code is already indented with one tab after each opening curly brace

arnavb: @ramini1996 You're really worried about speed? When properly used, cout can be as fast, even faster than printf. Here are some sources: (https://stackoverflow.com/a/33273370), (https://stackoverflow.com/a/12045514), (https://stackoverflow.com/a/896670). Since you are only printing to a console, the amount of efficiency your are striving for is premature.

> Premature optimization is the root of all evil - Donald Knuth

arnavb: As for your second comment, if you have a control statement, you should always indent what is inside for readability. It is confusing what you mean with the `bits/stdc++.h` thing.

ramini1996: Updated it

PR #2261, opened 2017-10-21
fixed java linear_search array initialization

leoperezpulido:
A Java array cannot be declared as in the previous version of the file, in fact the class doesn't compile. The array in main() method was declared like this:

`int arr[n];    // This is an illegal construct in Java.`

Please, have a look: [https://stackoverflow.com/questions/1200621/how-do-i-declare-and-initialize-an-array-in-java](url)

There are also other changes to the file so, please be so kind to examine it.


code/search/linear_search/LinearSearch.java
```diff
@@ -0,0 +1,113 @@
+/*
```
arnavb: This copyright notice is unnecessary and already implied.

AdiChat:
Looks great 😍 

leoperezpulido:
Thanks @AdiChat and @arnavb!

Issue #1927, opened 2017-10-18
add Check_is_square in any language


tbhaxor:
i am adding check_is_square in `php` and `csharp`

blockloop:
I'll try for C#

tbhaxor:
@blockloop i have commited now bro

blockloop:
Sorry. I misread the opened PR. Thought you were requesting it to be done. :)

AdiChat:
@blockloop Kindly take a look at other scopes of contribution. We have plenty 😄

lordozb:
Can I add in python and Java?

AdiChat:
Sure @lordozb Go for it 👍

kakshay21:
Can I add in Swift?

lordozb:
@AdiChat Fixed the code for Java

AdiChat:
Sure @kakshay21 Go for it 👍

teleginzhenya:
Can I make it with JavaScript?

AdiChat:
Sure @teleginzhenya Go for it 👍

edwardrichtofen115:
I'M doing this in c++

PR #2262, opened 2017-10-21
A Golang script to check if number is perfect square

nsurampu:
A Golang script with reference to issue #1927

### DIFF
--- a/code/mathematical-algorithms/Check_is_square/square_checker.go
+++ b/code/mathematical-algorithms/Check_is_square/square_checker.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"fmt"
+	"math"
+)
+
+func square_check(a float64) int {
+
+	var int_root int
+	var float_root float64
+	
+	int_root = int(math.Sqrt(a))
+	float_root = math.Sqrt(a)
+
+	if float_root - float64(int_root) != 0 {
+		return -1
+	} else {
+		return 1
+	}
+
+}
+
+func main() {
+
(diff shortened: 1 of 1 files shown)


code/mathematical-algorithms/Check_is_square/square_checker.go
```diff
@@ -0,0 +1,38 @@
+package main
+
+import (
+	"fmt"
+	"math"
+)
+
+func square_check(a float64) int {
```
arnavb: Make the return type bool, and call the function `is_square`. Return true if the number is a perfect square, and false otherwise.

code/mathematical-algorithms/Check_is_square/square_checker.go
```diff
@@ -0,0 +1,38 @@
...
+
+	var int_root int
+	var float_root float64
+	
+	int_root = int(math.Sqrt(a))
+	float_root = math.Sqrt(a)
```
arnavb: Remove one call to sqrt:

```go
float_root = math.Sqrt(a)
int_root = int(float_root)
```
On the other hand, you don't need the float root, you can simplify this whole function to:

```go
func square_check(a int) bool {
	 var int_root int = int(math.Sqrt(a))
	 return (int_root * int_root) == a
}
```

arnavb: Forgive me for any errors with the code. I am inexperienced with Go.

PR #2309, opened 2017-10-22
a^b %m using fermats little theorem in python

ansonb:
Added code for finding (a^b)%m in (log(a)+log(b)+log(m-2)) time using Fermat's little theorem in python.

Assumption is that m is prime.


AdiChat:
Looks great 😍 

PR #2311, opened 2017-10-22
improved performance of Bitcount example in C/C++, also fixed comments

stuxxnet42:
**Fixes issue:**
C/C++ example for Bitcounting used a non optimal algorithm.
**Changes:**
 gcc will compile this algorithm to use the "blsr" instruction when avaliable meaning that 
n&=(n-1);
only uses one instruction.
Clang will even compile it to use the builtin popcnt instruction resulting in a single instruction for the whole algorithm


AdiChat:
Looks great 😍 

Issue #2150, opened 2017-10-20
Remove use of platform specific libraries in C 


ajiteshr7:
**This is a(n):**

- [ ] New algorithm
- [ ] Update to an existing algorithm
- [x] Error
- [ ] Proposal to the Repository

**Details:**

Few C programs are using platform specific libraries.

PR #2336, opened 2017-10-22
Added a new program in the bit-manipulation category and fixed issue #2150

Ravikc:
**Changes:**
Added a new program to check if a ( big ) binary number represented as a string is a multiple of 3 or not.


Ravikc:
Addressing issue #2150 .

PR #2352, opened 2017-10-22
added Fast Fourier Transform in java

dadobt:
added Fast Fourier Transform in java


code/divide_conquer/fast_fourier_transform/FFT.java
```diff
@@ -0,0 +1,144 @@
+public class FFT {
```
yatharthshahjpr: Please add open genus comment here !
You can refer to the community guidelines for the same ! 👍 
:)

PR #2361, opened 2017-10-23
Rail fence cipher readme

IvanFon:
**Changes:**
Add README for rail fence cipher


AdiChat:
There are 203 modified files which are coming from other contributions which is resulting in a conflict. Kindly fix this 👍

IvanFon:
Sorry about that, I think this should fix it.

PR #2371, opened 2017-10-23
Add `QueueStream` to Data Structures in C#

willnode:
**Fixes issue:** Add a new algorithm from my own implementation

**Changes:**
I have an unique data structure concept so I like to share it here.
This algorithm is an alternative implementation of queue, when working with bytes and stream.
It is very optimized, and run well on multi-threaded environment. 

I have a [software](https://github.com/willnode/socket-clipboard/) that uses this algorithm in case you want a proof if it works.

I hope this pull requests help. Sorry if something wrong, it is the first time I made pull requests on a real project :)

### DIFF
--- a/code/data_structures/queue/queue_stream/queue_stream.cs
+++ b/code/data_structures/queue/queue_stream/queue_stream.cs
@@ -1,0 +1,170 @@
+using System;
+
+namespace Cosmos_Data_Structures
+{
+    /// <summary>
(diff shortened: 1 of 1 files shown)


AdiChat:
Looks great 😍 
It is perfect to see this algorithm in work in your tool. It is a very cool tool ❤️ 

PR #2382, opened 2017-10-23
Created armstrong.rb

AnkDos:


code/mathematical-algorithms/armstrong_numbers/armstrong.rb
```diff
@@ -0,0 +1,26 @@
...
+ reminder=0
+ sum=0
+ hold=num
+ while(quotient>0)
+   quotient=num/10
+   reminder=num%10
+   sum=sum+cube(reminder)
+   num=quotient
+ end
+if hold==sum
```
arnavb: return hold==sum also works here.

code/mathematical-algorithms/armstrong_numbers/armstrong.rb
```diff
@@ -0,0 +1,26 @@
+#created by AnkDos
+#You can Simply use '**' to cube ,ex. 3**3=27 ,so cube(remainder) can be written as (reminder**3) 
+def cube(num)
+    return num*num*num
```
arnavb: no need for the return.

AnkDos: Yeah mate, But thats just a habbit.
Thanks!

code/mathematical-algorithms/armstrong_numbers/armstrong.rb
AnkDos: yes , But a rubyist will get the trick.Like using '<<' in place of 'push' and etc etc.
I tried making it for the begginers like me.As you can see there is also a useless cube() method too.
Thanks !

AdiChat:
Looks great 😍 

PR #2390, opened 2017-10-23
password strength checker

Ahmadkhan12345566:
Added password strength checker script in php


AdiChat:
Kindly remove the `xml` and `iml` files 👍

PR #2392, opened 2017-10-23
Diameter of Tree

hrsvrdhn:
Diameter of a Tree is basically finding two farthest points in a acylic graph.
This is achievable in linear complexity by using two DFS/BFS traversal.

### DIFF
--- a/code/graph-algorithms/diameter_of_tree/diameter_of_a_tree.cpp
+++ b/code/graph-algorithms/diameter_of_tree/diameter_of_a_tree.cpp
@@ -57,3 +57,17 @@ int main()
     printf("The diameter of tree is %d\n.", dist[endpoint2]);
 	return 0;
 }
+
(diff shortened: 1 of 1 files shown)


deesims:
can i do this in go ?

hrsvrdhn:
No, not possible. As we have to find two points which are farthest from each other. @deesims

AdiChat:
@deesims Yes, you can implement this in `Go` 👍

AdiChat:
@hrsvrdhn This code in `C++` exists ([see](https://github.com/OpenGenus/cosmos/blob/master/code/data_structures/tree/binary_tree_diameter/binary_tree_diameter.cpp)) 👍

hrsvrdhn:
@AdiChat Great ! I think my code is more suitable for finding the diameter of a graph. Am I right?

PR #2398, opened 2017-10-23
Added the rope datastructure in python

moore3071:
**Changes:**
Added Ropes to the datastructure list. Ropes are an alternative to strings used in some text-editors.


AdiChat:
Looks great 😍 

PR #2403, opened 2017-10-24
Added Go solution to the Josephus problem.

jlr0045:
**Changes:**
I added a Golang solution to the Josephus problem.


AdiChat:
Looks great 😍 

Issue #2406, opened 2017-10-24
Add implicit treap implementation


daacaceressa:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

PR #2407, opened 2017-10-24
Add implicit treap implementation in C++

daacaceressa:
Fixes issue #2406

There is an explicit treap implementation, I considered good to add an implicit one.

With this implemenation you are able to handle a list by performing operations such as get the minimum of an interval and reverse the interval both in O(Log N) time complexity.

The reverse interval query is also implemented with a lazy propagation approach.


AdiChat:
Looks great 😍 

AdiChat:
Kindly remove the executable and object file ( `.exe` and `.o` files). It will be good to merge then 👍

daacaceressa:
I removed the files requested

PR #2413, opened 2017-10-24
Create xor_swap.c

alyssoncs:
swaps two integers


code/bit-manipulation/XOR_swap/xor_swap.c
```diff
@@ -0,0 +1,6 @@
+void xor_swap(int * restrict a, int * restrict b)
```
yatharthshahjpr: Please add open Genus comment here ! 
Refer community guidelines  ! 👍

PR #2423, opened 2017-10-24
Travis: build support for both C and C++, using gcc and clang compilers

InfiniteCoder:
**Fixes issue:** Now both C and C++ are supported, for both gcc and clang compiler

**Changes:**
changed travis.yml to use matrix. Just add new languages following the same format,
changed makefile to add C++ compilation.


arnavb:
You opened another PR?

InfiniteCoder:
@arnavb I thought it'd be better to push everything now, rather than waiting for something to cause merge conflict. I don't know why the PR won't add into the last one. So I created new one. The old one could be accepted or rejected, it shouldn't make a difference(I guess).

arnavb:
You should probably close the old one, then.

AdiChat:
Kindly remove `n_queen.c` and `Exponentation_power.c` and include them in separate pull requests and remove the travis batch from readme as well. We will add the batch later. 

It will be good to be merged 👍

InfiniteCoder:
@AdiChat  Sure. I'll make a new PR, as this one is riddled with commits. 😆

PR #2453, opened 2017-10-25
An Iterative alternative for binary search

haritha1997:
Since iterative methods are more commonly used, seems useful.

**Fixes issue:** #[Recursive methods are quite complicated to understand for beginners. ]

**Changes:**
[Iterative methods are fast and simple.]

### DIFF
--- a/code/search/binary_search/binary_search.c
+++ b/code/search/binary_search/binary_search.c
@@ -2,46 +2,40 @@
 /*
  * Part of Cosmos by OpenGenus Foundation
 */
-int recursiveBinarySearch(int arr[], int l, int r, int x)
-{
-   if (r >= l)
-   {
-        int mid = l + (r - l)/2;
- 
-        if (arr[mid] == x)  
-            return mid;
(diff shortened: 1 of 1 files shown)


code/search/binary_search/binary_search.c
```diff
@@ -2,46 +2,40 @@
...
-            return mid;
-         if (arr[mid] > x) 
-             return recursiveBinarySearch(arr, l, mid-1, x);
-       
-         return recursiveBinarySearch(arr, mid+1, r, x);
-   }
-   return -1;
-}
-
 int binarySearch(int arr[], int l, int r, int x)
```
Vitorvgc: It already seems to have an iterative version implemented, and your version is equivalent to it.

Vitorvgc: You can make an update to add comments to explain the code, but if so you should just add the new information.

haritha1997: sure

PR #2463, opened 2017-10-25
Add tags

iattempt:
**Fixes issue:**
1. if parameter is not correspond to particular type then compiler will not provide clear and understood information to user.

**Changes:**
1. add tags to provide more information
2. change coding style


AdiChat:
Looks great 😍 

Issue #807, opened 2017-10-07
Find if two numbers are coprime


win72:
Can I implement in C++?

AdiChat:
Sure. Go for it 👍

win72:
Done. PR #829

Raghav17:
I am doing it in C.

Raghav17:
Done. PR #836

Jay9596:
Rust implementation #857

ArjitJ:
i can solve this in c++ using recursion technique
#879

skydans:
Done PR java implementation. #2478

abhijeet010304:
I would like to work on it in C++  and pyhton please assign it to me as a part of GSSoC 20

PR #2478, opened 2017-10-25
Added java implementation of coprime numbers.

skydans:
**Fixes issue:** #807

**Changes:**
Added java implementation of coprime numbers.


code/mathematical-algorithms/coprime_numbers/CoprimeNumbers.java
```diff
@@ -0,0 +1,15 @@
+class CoprimeNumbers{
+	public static long gcd(long c,long d){
+		if(d==0){
```
arnavb: Ternary operator could be used here:

```java
return (d == 0) ? c : gcd(d,c%d);
```

PR #2491, opened 2017-10-25
added eulerian path algorithm in java

chugh22:
added algorithm to find if a given undirected graph has a Eulerian path and circuit.


AdiChat:
Looks great 😍 

PR #2502, opened 2017-10-25
Improve union find implementation in python

hir12111:
- Allow counting elements in each group
- Use path compression


AdiChat:
Looks great 😍 

PR #2532, opened 2017-10-26
Create sqrt_decomposition

ipmankus:

**Changes:**


code/sqrt_decomposition
```diff
@@ -0,0 +1 @@
+
```
Vitorvgc: It seems that the code is missing...

PR #2536, opened 2017-10-26
Add python 3.x compatibility

Akyshnik:
**Fixes issue:** 
Add example that compatible with latest version of python
**Changes:**
Add python 3.x compatibility as well as make the algorithm simple and pythonic.

### DIFF
--- a/code/mathematical-algorithms/replace_0_with_5/change0to5.py
+++ b/code/mathematical-algorithms/replace_0_with_5/change0to5.py
@@ -1,4 +1,6 @@
 # Part of Cosmos by OpenGenus Foundation
+
+# python 2 code:
 user_input = raw_input("Enter the number: ")
 try:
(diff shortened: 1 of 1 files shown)


code/mathematical-algorithms/replace_0_with_5/change0to5.py
```diff
@@ -14,3 +16,11 @@
 
 except Exception:
   print("You did not specify an integer")
+
+  
+# python 3 code:
```
Vitorvgc: You could use `sys.version_info` to check in what version the code is running and then run only the appropriate code for the version.

Akyshnik: Done

AdiChat:
Looks great 😍 

PR #2541, opened 2017-10-26
Add average for C#

AaronC81:


williamfiset:
I think this is too simple to be worth adding.

PR #2546, opened 2017-10-26
calculate sum of GCD(i,j) of all pairs

HoaNP:

**Changes:**


williamfiset:
Please get rid of all the boilerplate #defines and only keep what's needed, TY.

PR #2552, opened 2017-10-26
Create avl_tree.sc

rakeshpatel967:


williamfiset:
The names in the code suggest a Red-Black tree are you sure this is an AVL tree PR?

rakeshpatel967:
Yes william,  it was a mistake you can see the updated code.

PR #2563, opened 2017-10-26
Update topological_sort.cpp

ollgromova:
- [ ] **Fixes

 issue:** #[Added topological_sort.cpp]

**Changes:**
[topological_sort.cpp]

### DIFF
--- a/code/sorting/topological_sort/topological_sort.cpp
+++ b/code/sorting/topological_sort/topological_sort.cpp
@@ -1,97 +1,55 @@
-// A C++ program to print topological sorting of a DAG
+#include <bits/stdc++.h>
+
 // Part of Cosmos by OpenGenus Foundation
-#include<iostream>
-#include <list>
-#include <stack>
+
 using namespace std;
- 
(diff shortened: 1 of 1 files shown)


code/sorting/topological_sort/topological_sort.cpp
```diff
@@ -1,97 +1,55 @@
-// A C++ program to print topological sorting of a DAG
+#include <bits/stdc++.h>
```
hwchiu: We prefer to don't use `<bits/stdc++.h>` here.
list all headers what this program actually needs is better.

code/sorting/topological_sort/topological_sort.cpp
```diff
@@ -1,97 +1,55 @@
...
-{
-    this->V = V;
-    adj = new list<int>[V];
+
+int n;
+
+vector <int> vertex[100];
+vector <int> ans;
+
+bool used[100];
```
hwchiu: Use a const variable for that hardcode `100`

PR #2567, opened 2017-10-27
Add GCD and LCM code in Kotlin

thiagobbt:
**Fixes issue:** Missing GCD and LCM algorithms in Kotlin

**Changes:**
Adds Kotlin code implementing GCD and LCM


williamfiset:
Because of the possibility of overflow, I would do early division:
```kotlin
fun lcm(a: Int, b: Int): Int = if (b == 0) 0 else (a / gcd(a, b)) * b
```
instead of 
```kotlin
fun lcm(a: Int, b: Int): Int = if (b == 0) 0 else a * b / gcd(a, b)
```

PR #2571, opened 2017-10-27
I have added my program of factorial of 100.

anushree816:
As normal programs for factorial does not work for large digits.

**Changes:**

### DIFF
--- a/factorial.cpp
+++ b/factorial.cpp
@@ -1,0 +1,46 @@
+#include<iostream>
+using namespace std;
+#define MAX 500
+ 
+int multiply(int x, int res[], int res_size)
+ void factorial(int n)
+{
+    int res[MAX];
+
+    res[0] = 1;
+    int res_size = 1;
+ 
+      for (int x=2; x<=n; x++)
+        res_size = multiply(x, res, res_size);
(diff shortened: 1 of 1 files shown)


factorial.cpp
```diff
@@ -0,0 +1,46 @@
+#include<iostream>
+using namespace std;
+#define MAX 500
```
arnavb: const int is better than macro.

factorial.cpp
```diff
@@ -0,0 +1,46 @@
...
+ 
+int multiply(int x, int res[], int res_size)
+ void factorial(int n)
```
arnavb: These two lines make no sense at all. Why is multiply here as well?

anushree816:
Thanks for the suggestions. I made the changes.

arnavb:
Where?

anushree816:
I have changed the code and created the pull request.

arnavb:
You don't have to create a new PR, you could have just added commits to this.

anushree816:
Const int done. Anything else?

arnavb:
No, nothing.

PR #2590, opened 2017-10-27
Create reverse_list.c

hemantyadaviitg2016:
created new .c file to reverse the linked list using recursive method.


hwchiu:
There's a existing implementation in c and you can find [here](https://github.com/OpenGenus/cosmos/blob/master/code/data_structures/linked_list/linked_lists_operations/reverse_list.c).
Besides, that files should be located in linked_lists_operations.

PR #2609, opened 2017-10-27
Added parallel algorithm

abdelhakeem:
Added *countSetBitsParallel(int)* function for counting set bits in less operations.


AdiChat:
Looks great 😍 

PR #2617, opened 2017-10-27
Create check_sqaure.c

hemantyadaviitg2016:
Program to find whether a no. is a perfect square or not.

### DIFF
--- a/code/mathematical-algorithms/Check_is_square/check_sqaure.c
+++ b/code/mathematical-algorithms/Check_is_square/check_sqaure.c
@@ -1,0 +1,29 @@
+#include <stdio.h>
+#include <math.h>
+#include <stdbool.h>
+
+bool issqaure(int n)
+{
+	int sroot=sqrt(n);
+	if(sroot*sroot == n)
+		return true;
+
+	return false;
+}
+
+int main()
+{
+	printf("Enter the no. to be examined.\n");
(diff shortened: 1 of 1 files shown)


code/mathematical-algorithms/Check_is_square/check_sqaure.c
```diff
@@ -0,0 +1,29 @@
...
+
+	return false;
+}
+
+int main()
+{
+	printf("Enter the no. to be examined.\n");
+	double n;
+	scanf("%lf",&n);
+	int root=issqaure(n);
```
hwchiu: What is the purpose of this variable `root` ?

code/mathematical-algorithms/Check_is_square/check_sqaure.c
```diff
@@ -0,0 +1,29 @@
...
+	if(sroot*sroot == n)
+		return true;
+
+	return false;
+}
+
+int main()
+{
+	printf("Enter the no. to be examined.\n");
+	double n;
```
hwchiu: Since the `issquare` take `int` as its input, why should we use a `double` to get out input ?

hemantyadaviitg2016: made the relevant changes.

PR #2623, opened 2017-10-28
Add SingletonPattern.js

TwoTau:
Uses ES6

Added a singleton implementation in JavaScript.


TwoTau:
According to the [Contribution guidelines](https://github.com/OpenGenus/cosmos/wiki/contribute#contribution-guidelines), the naming convention for this file would be `singleton_pattern.js` but the other files in this directory are named using camelcase: `SingletonPattern.cpp` and `SingletonPattern.java`.

Should I rename this file to `singleton_pattern.js`?

Issue #2619, opened 2017-10-28
Add factorial in C


etrexel:
**This is a(n):**

- [ ] New algorithm
- [x] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

Gupta-shifali:
Can I implement this in c++ @AdiChat?

pruthiviraj71:
I added a new algorithm for factorial.

GingerDragon7:
Added Factorial in C using recursion technique #2656

swapnilbaluja:
kindly review #2693

PR #2650, opened 2017-10-28
Added Factorial using Stirling's approximation in c language.

pruthiviraj71:
**Fixes issue:** #2619 

**Changes:**

### DIFF
--- a/code/mathematical-algorithms/factorial/factorialstirling.c
+++ b/code/mathematical-algorithms/factorial/factorialstirling.c
@@ -1,0 +1,27 @@
+// Factorial using Stirling's approximation in c language
+
+float stirling_approximation(int n) {
+    int fact;
+    float stirling, ans;
+    fact = factorial(n);
+    stirling = sqrt(2.0*3.1416*n) * pow(n / 2.71828, n);
+    ans = fact / stirling;
+    return ans;
+}
+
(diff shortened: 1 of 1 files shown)


code/mathematical-algorithms/factorial/factorialstirling.c
```diff
@@ -0,0 +1,27 @@
+// Factorial using Stirling's approximation in c language
+
+float stirling_approximation(int n) {
+    int fact;
+    float stirling, ans;
+    fact = factorial(n);
+    stirling = sqrt(2.0*3.1416*n) * pow(n / 2.71828, n);
```
arnavb: Make both π and _e_ constants (#define, const int)

code/mathematical-algorithms/factorial/factorialstirling.c
```diff
@@ -0,0 +1,27 @@
...
+    ans = fact / stirling;
```
hwchiu: The Stirling's approximation seems wrong.
It should be `pow(exp,-n) * pow(n,n) * sqrt(2*n*PI) ;`
In you case, it's n! / (sqrt(2*n*PI) * pow(exp, -n) * n^n)
You can use the input as 10 to check its value, it should close to 3628800.

PR #2673, opened 2017-10-29
remove broken c implementation because of its uncomplete function and…

hwchiu:
**Fixes issue:**
- This file does't contain the implementation of data structure and some functions. such as `MoveNode`.
- User can't know how to use this code to do `merge_two_linked_list`.
- Maybe we should possible test all PR before accept it.

**Changes:**
- remove `merge_two_linked_list.c`


AdiChat:
Looks like somehow this got overlooked. 
Looks great 😍 

Issue #1242, opened 2017-10-12
Add Conway's Game of Life


Rhendz:
https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life
Added an implementation in C++

PR #2678, opened 2017-10-29
Add Conway's Game of Life in Ruby

polinadotio:
Reference #1242

### DIFF
--- a/code/cellular_automaton/conways_game_of_life/life.rb
+++ b/code/cellular_automaton/conways_game_of_life/life.rb
@@ -1,0 +1,61 @@
+#!/usr/bin/env ruby
+# Part of Cosmos by OpenGenus Foundation
+
+class GameOfLife
+  attr_accessor :board
+  
+  def initialize(board = [])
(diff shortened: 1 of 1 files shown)


polinadotio:
thanks for the follow up @Vitorvgc. I read through the contribution guidelines, is this the required comment? And does this need to be part of the commit message or just in the PR? 

```
Every code contribution may contain the following comment as a part of it:

/* Part of Cosmos by OpenGenus Foundation */
```

I would also recommend that the maintainers put a link to this https://github.com/OpenGenus/cosmos/wiki/contribute in the README so it's easier for people to find it and read. I originally looked and couldn't find a link to the contribution guidelines. Thanks!

Vitorvgc:
Yes, this is the required comment, and it needs to be presented only in the code.

Thanks for the suggestion @polinadotio, it would be really helpful 👍

polinadotio:
thanks for the clarification, just added it in

AdiChat:
Looks great 😍 

PR #2748, opened 2017-10-30
fix syntax error of existing python implementation.

hwchiu:
**Changes:**
- Change the __name__ from `main` to `__name__` to fix the logical error.


AdiChat:
Looks great 😍 

PR #2751, opened 2017-10-30
Update bfs.cpp

libro:
use "define" preprocessor that will make easy to write long syntax..
such "first" to 'F',"second" to 'S'
and after writing  "using namespace std" we do not need  to add prefix "std" before any predefined function

**Changes:**


arnavb:
using namespace std is considered bad practice anyway, so if someone avoids it, it is good.

PR #2754, opened 2017-10-30
Stooge_Sort_in_C++

maheshjindal:

**Changes:**


code/sorting/stooge_sort/stoogesort.cpp
```diff
@@ -0,0 +1,39 @@
+#include <iostream>
+using namespace std;
+
+void stoogesort(int arr[],int l, int h)
```
arnavb: Use vectors, not raw arrays.

PR #2818, opened 2017-10-30
Singleton pattern implementation at PHP

maxprihodko8:
Singleton pattern implementation at PHP, 2 classes - one singleton and another - class which handles singleton instance

**Changes:**


AdiChat:
Looks great 😍 

PR #2837, opened 2017-10-31
greedy-minimum-coins: Warn about wrong solutions

ljdelight:
**Fixes issue:**
This warns users about the solutions that don't use dynamic programming to determine a valid solution. A `while` loop overlooks the complexities of the problem.

**Changes:**
This updates the `minimum_coins` README.md to document the invalid solutions in the project. Most will return a `wrong answer` when there is an overly-optimistic greedy algorithm.


ljdelight:
Closing this as not needed. The patch is a small documentation note on the DP answers being wrong in some cases, and there are many other more useful PRs.

PR #2842, opened 2017-10-31
doing early return to solve division by 0 issue

ferrelucas:
**Fixes issue:** #805 Division by zero

**Changes:**
I added an early return in the average method for when the size of the array is 0 to avoid division by zero.


AdiChat:
Looks great 😍 

PR #2881, opened 2017-10-31
Added a new file of solving nqueen problem using stl

sakshi7896:
Added a new file of solving nqueen problem using stl in c++. N is variable and can be changed.

**Changes:**


arnavb:
Since you have two separate codes here, separate them into branches and separate pull requests.

PR #2884, opened 2017-10-31
Code3

aditicode4:


arnavb:
This already exists. Look in the code/search/binary_search/binary_search.cpp.

PR #2893, opened 2017-10-31
Create bogo_sorting.cpp

akshaysharma2277:

**Changes:**

### DIFF
--- a/code/sorting/bogo_sort/bogo_sorting.cpp
+++ b/code/sorting/bogo_sort/bogo_sorting.cpp
@@ -1,0 +1,66 @@
+/*
+ * C Program to Implement BogoSort in an Integer Array
+ */
+#include <stdio.h>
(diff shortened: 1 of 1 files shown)


code/sorting/bogo_sort/bogo_sorting.cpp
```diff
@@ -0,0 +1,66 @@
...
+    }
+    printf("\n");
+}
+ 
+/* Code to check if the array is sorted or not */
+int is_sorted(int *a, int n)
+{
+    while (--n >= 1)
+    {
+        if (a[n] < a[n - 1])
```
OvermindDL1: Shouldn't this be `<=` instead of `<` so it handles duplicate values properly (or it will infinite loop)?

arnavb:
This file already exists, like your other PRs. Same advice goes here.

PR #2898, opened 2017-10-31
Create rad_sort.cpp

akshaysharma2277:

**Changes:**

### DIFF
--- a/code/sorting/radix_sort/rad_sort.cpp
+++ b/code/sorting/radix_sort/rad_sort.cpp
@@ -1,0 +1,69 @@
+#include <iostream.h>
+#include <conio.h>
+#define MAX 10
+
+class radixsort{
+    int arr[MAX],n;
+    public:
+    void getdata();
+    void showdata();
+    void sortLogic();
+};
+
+void radixsort :: getdata(){
+    cout<<"How many elements you require : ";
+    cin>>n;
+    for(int i=0;i<n;i++)
+        cin>>arr[i];
+}
+
+void radixsort :: showdata(){
+    cout<<"\n--Display--\n";
+    for(int i=0;i<n;i++)
+        cout<<arr[i]<<"   ";
+}
+
+void radixsort :: sortLogic(){
+    //for base 10int temp;
+    int bucket[10][20], buck_count[10], b[10];
+    int i,j,k,r,no_of_passes=0,divisor=1,largest,pass_no;
+
+    largest=arr[0];
+
+    for(i=1;i<n;i++)  //Find the largest Number
+    {
+        if(arr[i] > largest)
+            largest=arr[i];
+    }
+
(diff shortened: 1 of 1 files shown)


code/sorting/radix_sort/rad_sort.cpp
```diff
@@ -0,0 +1,69 @@
+#include <iostream.h>
+#include <conio.h>
```
arnavb: Don't include this, it is non-standard.

code/sorting/radix_sort/rad_sort.cpp
```diff
@@ -0,0 +1,69 @@
...
+#define MAX 10
```
arnavb: const int is better than #define.

code/sorting/radix_sort/rad_sort.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+class radixsort{
+    int arr[MAX],n;
+    public:
```
arnavb: This line should be same level of indentation as class according to style guide, and the private variables need to be below the public ones.

code/sorting/radix_sort/rad_sort.cpp
```diff
@@ -0,0 +1,69 @@
+#include <iostream.h>
+#include <conio.h>
+#define MAX 10
+
+class radixsort{
+    int arr[MAX],n;
```
arnavb: Use vectors, not raw arrays.

code/sorting/radix_sort/rad_sort.cpp
```diff
@@ -0,0 +1,69 @@
...
+    public:
+    void getdata();
+    void showdata();
+    void sortLogic();
+};
+
+void radixsort :: getdata(){
+    cout<<"How many elements you require : ";
+    cin>>n;
+    for(int i=0;i<n;i++)
```
arnavb: If you use vectors, you don't have to ask for size, only elements, and can read/push_back until an EOF.

code/sorting/radix_sort/rad_sort.cpp
```diff
@@ -0,0 +1,69 @@
...
+            for(j=0; j<buck_count[k]; j++)
+                arr[i++] = bucket[k][j];
+        }
+
+        divisor *= 10;
+    }
+}
+
+void main(){
+    clrscr();
```
arnavb: Don't.

code/sorting/radix_sort/rad_sort.cpp
```diff
@@ -0,0 +1,69 @@
...
+    cout<<"\n*****Radix Sort*****\n";
+    radixsort obj;
+    obj.getdata();
+    obj.sortLogic();
+    obj.showdata();
+    getch();
```
arnavb: Don't.

PR #2899, opened 2017-10-31
Rename bucket_sort.cpp to bucket_sorting.cpp

akshaysharma2277:
**Changes:**
[in bucket sort algorithm]


arnavb:
You haven't really renamed anything, otherwise it would say here: File renamed without changes.

PR #2918, opened 2017-10-31
Modify return type of main

EroMonsterSanji:
**Changes:**
Change return type of main func from `void` to `int`, 
and to print a `long int` value, "%ld" should be used instead of "%d".


AdiChat:
Looks great 😍 

PR #2924, opened 2017-10-31
Implemented factorial in Rust

sbxx:


AdiChat:
Looks great 😍 

The reason for the delay in merging has been that we were expecting changes. No problem, it looks great in its original form. 

PR #2930, opened 2017-11-01
implemented strategy pattern in cpp

omarmmali:

I implemented the strategy design pattern of the first example in Head First Design Patterns book

**Changes:**

Added strategy_pattern.cpp

### DIFF
--- a/code/design_pattern/strategy_pattern/strategy_pattern.cpp
+++ b/code/design_pattern/strategy_pattern/strategy_pattern.cpp
@@ -1,0 +1,141 @@
+#include <iostream>
+using namespace std;
+
+class IFlyBehavior
+{
+	public :
+	virtual ~IFlyBehavior(){}
+	virtual void fly() = 0;
+};
+
+class FlyWithWings : public IFlyBehavior
+{
+	public :
+	void fly()
+	{
+		cout << "I fly with wings!" << endl;
+	}	
+};
+
+class FlyNoWay : public IFlyBehavior
+{
+	public :
+	void fly() 
+	{
+		cout << "I don't fly, buddy." << endl;
+	}
+};
+
+class IQuackBehavior 
+{
+	public :
+	virtual ~IQuackBehavior(){}
+	virtual void quack() = 0;
+};
+
+class Quack : public IQuackBehavior
+{
+	public :
+	void quack()
+	{
+		cout << "Quack!" << endl;
+	}
+};
+
+class Squeak : public IQuackBehavior
+{
+	public :
+	void quack() 
+	{
+		cout << "Squeak!" << endl;
+	}
+};
+
+class MuteQuack : public IQuackBehavior
+{
+	public :
+	void quack()
+	{
+		cout << "I'm a mute duck." << endl;
+	}
+};
+
+class Duck 
+{
+	public :
+	IFlyBehavior* flyBehavior;
+	IQuackBehavior* quackBehavior;
+	
+	Duck()
+	{
+		flyBehavior = new FlyNoWay();
+		quackBehavior = new Quack();
+	}
+
+	void setFlyBehavior(IFlyBehavior* fb)
+	{
+		flyBehavior = fb;
+	}
+	
+	void setQuackBehavior(IQuackBehavior* qb)
+	{
+		quackBehavior = qb;
+	}
+	
+	void performFly()
+	{
+		flyBehavior->fly();
+	}
+	
+	void performQuack()
+	{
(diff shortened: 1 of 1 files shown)


arnavb:
Please remove the extra binary file!

code/design_pattern/strategy_pattern/strategy_pattern.cpp
```diff
@@ -0,0 +1,141 @@
+#include <iostream>
+using namespace std;
+
+class IFlyBehavior
+{
+	public :
```
arnavb: The word public should be at the same level as the class, as per the coding style guidelines.

code/design_pattern/strategy_pattern/strategy_pattern.cpp
```diff
@@ -0,0 +1,141 @@
...
+	virtual ~IFlyBehavior(){}
+	virtual void fly() = 0;
+};
+
+class FlyWithWings : public IFlyBehavior
+{
+	public :
```
arnavb: Same as above.

code/design_pattern/strategy_pattern/strategy_pattern.cpp
```diff
@@ -0,0 +1,141 @@
...
+	void quack()
+	{
+		cout << "I'm a mute duck." << endl;
+	}
+};
+
+class Duck 
+{
+	public :
```
arnavb: The public thing applies for all your classes.

code/design_pattern/strategy_pattern/strategy_pattern.cpp
```diff
@@ -0,0 +1,141 @@
...
+	IFlyBehavior* flyBehavior;
```
arnavb: These members should not be public, but should be protected.

code/design_pattern/strategy_pattern/strategy_pattern.cpp
```diff
@@ -0,0 +1,141 @@
...
+	IQuackBehavior* quackBehavior;
```
arnavb: Also, since these are dynamic pointers, either add a destructor to delete these objects, or turn them into smart pointers. We don't want memory leaks in this code!

code/design_pattern/decorator_pattern/decorator_pattern.cpp
```diff
@@ -0,0 +1,101 @@
...
+    {
+        return 0.10 + CondimentDecorator::cost();
+    }
+};
+
+int main() 
+{
+    IBeverage* beverage = new HouseBlend();
+    std::cout << beverage->getDescription() + " $" << beverage->cost() << std::endl;
+    beverage = new Mocha(beverage);
```
arnavb: You have to delete the beverage before you set it to a new one in either case. Here I would seriously recommend using unique_ptr.

```C++
std::unique_ptr<IBeverage> beverage(new HouseBlend());
std::cout << beverage->getDescription() + " $" << beverage->cost() << std::endl;
beverage.reset(new Mocha(beverage.get()));
std::cout << beverage->getDescription() + " $" << beverage->cost() << std::endl;
beverage.reset(new Whip(beverage.get()));
std::cout << beverage->getDescription() + " $" << beverage->cost() << std::endl;
```

arnavb: Remember, in C++ there is no automatic garbage collection.

omarmmali: Thank you, I really appreciate the feedback. :D

arnavb:
Will you make the changes?

code/design_pattern/decorator_pattern/decorator_pattern.cpp
omarmmali: I can't seem to make unique_ptr work, so I resorted to making a tmp pointer to delete after wrapping beverage with mocha, but deleting said pointer causes a crash

arnavb: What are the errors you are getting for the unique_ptr? Are you using C++11? Are you including `<memory>`? If you are making a temp pointer, then do not delete anything, until the very end. Delete only once all pointers that point to the same thing, and delete once for each pointer otherwise.

omarmmali: Honestly I've been looking at the code for a while, I can't seem to figure out where the memory leak is. If I just delete beverage in the end of the main function, shouldn't that cover it?

PR #2931, opened 2017-11-01
Implemented graph cycle detection using union find

kashifmin:
**Changes:**
Added implementation of graph cycle detection in python

### DIFF
--- a/code/graph-algorithms/cycle_directed_graph/cycle_directed_graph.py
+++ b/code/graph-algorithms/cycle_directed_graph/cycle_directed_graph.py
@@ -6,55 +6,52 @@ class Graph:
         self.v = v
         self.adj = defaultdict(list)
 
-    def addEdge(self, u, v):
+    def add_edge(self, u, v):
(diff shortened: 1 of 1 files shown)


code/graph-algorithms/cycle_directed_graph/cycle_directed_graph.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def __init__(self, v):
+        self.v = v
+        self.adj = defaultdict(list)
+
+    def addEdge(self, u, v):
+        self.adj[u].append(v)
+
+    def findParent(self, v, setMap):
+        if setMap[v] == -1:
```
arnavb: Can be simplified

```Python
return v if setMap[v] == -1 else self.findParent(setMap[v], setMap)
```

kashifmin:
@arnavb made changes as suggested. Thanks.

AdiChat:
Looks great 😍 

Issue #2940, opened 2017-11-03
File name conflict


HidekiHokuto:
**This is a(n):**

- [ ] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

Some OSX may be case-insensitive, algorithms have same name in a folder may be replaced.
Like `cosmos/code/search/exponential_search.py` and `cosmos/code/search/Exponential_search.py`

arnavb:
#2941

PR #2941, opened 2017-11-03
Fixed name conflict

EroMonsterSanji:
**Changes:**
Renamed `Exponential_search.py` to `exponentialSearch.py` to fix the problem on case-insensitive system. (`Exponential_search.py` conflicts with `exponential_search.py`)


arnavb:
If there are two copies of the same algorithm for one language, then just delete one of them (it doesn't matter which one), and then move on.

arnavb:
This PR resolves #2940

PR #2945, opened 2017-11-04
rename-files

iattempt:
**Fixes issue:** 
the duplicate files (same name but not same case) is annoying in non-sensitive OS

**Changes:**
1. rename
2. duplicate files -> append serial number


AdiChat:
Looks great 😍 

PR #2970, opened 2017-11-09
provide a implementation of larget_sum with DP in golang

hwchiu:
**Fixes issue:**
- provide a implementation of larget_sum with DP in golang

**Changes:**


AdiChat:
Looks great 😍 

PR #3029, opened 2017-11-16
leap year algorith in python3

Titowisk:
**Changes:**
Added a simple program that identifies all the leap years between two given years, in python 3.
I did that using a normal for...in loop and also using List Comprehension.

**Location:** cosmos\code\unclassified\leap_year.py


AdiChat:
Looks great 😍 

PR #3032, opened 2017-11-16
Create Optimal_BST.py

ManasGupta1998:
Given a sorted array keys[0.. n-1] of search keys and an array freq[0.. n-1] of frequency counts, where freq[i] is the number of searches to keys[i]. Construct a binary search tree of all keys such that the total cost of all the searches is as small as possible.

**Changes:**
Adds Optimal Binary Search Tree problem  in Dynamic Programming


AdiChat:
Looks great 😍 

PR #3045, opened 2017-11-17
naive pattern search in cpp

tbhaxor:


code/string_algorithms/naive_pattern_search/naive_pattern_search.cpp
```diff
@@ -0,0 +1,39 @@
+// part of cosmos by opengenus foundation
+
+#include <iostream>
+#include <vector>
+using namespace std;
+
+vector<int> naive_search(string pattern, string text)
```
iattempt: could we return pair it consists of bool and vector<int> or pair it consists of bool and iterator?
e.g., std::pair<bool, std::string::iterator>  -> the first is true if found, the second is text start at.

iattempt: could we return pair it consists of bool and vector or pair it consists of bool and iterator?
e.g., std::pair<bool, std::string::iterator> -> the first is true if found, the second is text start at.

code/string_algorithms/naive_pattern_search/naive_pattern_search.cpp
```diff
@@ -0,0 +1,39 @@
...
+{
+	vector<int> indeces;
+	int plen = pattern.length();
```
iattempt: replace tab with 4 blank spaces

Issue #3048, opened 2017-11-17
add Fibonacci  search in any language


tbhaxor:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**
check [this](http://www.geeksforgeeks.org/fibonacci-search/)

ps : Don't copy paste the code

tbhaxor:
i can add in `cpp` or `python`

erwolfe:
I will do this in swift!

iattempt:
I'll try this in c++~

iattempt:
#3059 @erwolfe

nipunbharti:
I'll try this in JS

nipunbharti:
I have implemented this in C

PR #3080, opened 2017-11-22
JS implementation of Fibonacci search added

nipunbharti:
**Fixes issue:** #3048 

**Changes:**
JS implementation added

### DIFF
--- a/code/search/fibonacci_search/fibonacci_search.js
+++ b/code/search/fibonacci_search/fibonacci_search.js
@@ -41,5 +41,5 @@ function fibonacciSearch(ar,x,n) {
 
 }
 function min(a,b) {
-	a > b ? return b : return a;
+	return a>b?b:a;
 }
--- a/code/search/fibonacci_search/fibonacci_search.js
+++ b/code/search/fibonacci_search/fibonacci_search.js
@@ -1,0 +1,52 @@
+function fibonacciSearch(ar,x,n) {
+	var secondPreceding = 0,
+	firstPreceding = 1;
+	var nextNum = secondPreceding + firstPreceding;
+
+	while(nextNum < n){
+		secondPreceding = firstPreceding;
+		firstPreceding = nextNum;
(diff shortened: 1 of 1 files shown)


code/search/fibonacci_search/fibonacci_search.js
```diff
@@ -0,0 +1,52 @@
...
+
+	if(firstPreceding && ar[range+1] == x){
+		return range + 1;
+	}
+
+	return -1;
+
+}
+function min(a,b) {
+	if(a > b){
```
arnavb: Use the ternary operator here.

code/search/fibonacci_search/fibonacci_search.js
```diff
@@ -0,0 +1,52 @@
+function fibonacciSearch(ar,x,n) {
+	var secondPreceding = 0,
+	firstPreceding = 1;
```
arnavb: Any reason why you aren’t using var here?

nipunbharti: There is a comma to separate the two.

code/search/fibonacci_search/fibonacci_search.js
nipunbharti: Okay I'll fix it

code/search/fibonacci_search/fibonacci_search.js
arnavb: Oh right, didn’t see that. I think you should put both variables on the same line.

code/search/fibonacci_search/fibonacci_search.js
```diff
@@ -0,0 +1,45 @@
...
+
+	if(firstPreceding && ar[range+1] == x){
+		return range + 1;
+	}
+
+	return -1;
+
+}
+function min(a,b) {
+	a > b ? return b : return a;
```
arnavb: return a > b ? b : a;

AdiChat:
Looks great 😍 

PR #3089, opened 2017-11-25
Create biggest_of_n_numbers.py

Vaibhav1671998:
Added code to find biggest of n numbers in python

### DIFF
--- a/code/unclassified/biggest_of_n_numbers/biggest_of_n_numbers.py
+++ b/code/unclassified/biggest_of_n_numbers/biggest_of_n_numbers.py
@@ -1,6 +1,6 @@
 elements = []
 n = int(input("Enter number of elements:"))
-for i in range(1, n + 1):
+for i in range(0, n):
   elements.append(int(input("Enter element:")))
 elements.sort()
 print("Largest element is : ", elements[n - 1])
--- a/code/unclassified/biggest_of_n_numbers/biggest_of_n_numbers.py
+++ b/code/unclassified/biggest_of_n_numbers/biggest_of_n_numbers.py
@@ -1,0 +1,6 @@
+elements = []
+n = int(input("Enter number of elements:"))
+for i in range(1, n + 1):
+  elements.append(int(input("Enter element:")))
+elements.sort()
+print("Largest element is : ", elements[n - 1])


Vaibhav1671998:
@arnavb Check this out too :100:

code/unclassified/biggest_of_n_numbers/biggest_of_n_numbers.py
```diff
@@ -0,0 +1,6 @@
+elements = []
+n = int(input("Enter number of elements:"))
+for i in range(1, n + 1):
+  elements.append(int(input("Enter element:")))
```
arnavb: Enter the next element: is a better prompt.

code/unclassified/biggest_of_n_numbers/biggest_of_n_numbers.py
```diff
@@ -0,0 +1,6 @@
+elements = []
+n = int(input("Enter number of elements:"))
+for i in range(1, n + 1):
```
arnavb: Why not range(0, n)?

code/unclassified/biggest_of_n_numbers/biggest_of_n_numbers.py
```diff
@@ -0,0 +1,6 @@
...
+  elements.append(int(input("Enter element:")))
+elements.sort()
+print("Largest element is : ", elements[n - 1])
```
arnavb: You don’t even need an array, you could just keep track of the largest number constantly.

Vaibhav1671998: @arnavb That can be one of the logic but i think this is much simpler one .

code/unclassified/biggest_of_n_numbers/biggest_of_n_numbers.py
Vaibhav1671998: It's running in loop so if i write  "Enter the next element:"
this would be shown for the first element too which is incorrect . 
So i think it is not better

Vaibhav1671998:
Check it out now @arnavb

Vaibhav1671998:
@AdiChat Merge this one if u r happy with this :100:

AdiChat:
Looks great 😍 

PR #3098, opened 2017-11-30
Generalize sort algorithms and update test file

iattempt:
**Fixes issue:**

**Changes:**
- [x] generalize insertion/selection/bubble sort
- [x] improve sorting test
- [x] remove unnecessary template parameter


code/sorting/bubble_sort/bubble_sort.cpp
```diff
@@ -1,82 +1,50 @@
...
+            for (auto j = begin; j != frontOfSorted; ++j)
+            {
+                auto nextOfJ = j;
+                if (compare(*++nextOfJ, *j))
+                {
+                    std::iter_swap(nextOfJ, j);
+                    swapped = true;
+                }
+            }
+            if (swapped == false)
```
arnavb: if (!swapped)

code/sorting/sort_test/sort_test.cpp
```diff
@@ -2,109 +2,163 @@
...
+
+        psf(arr, arr_end);
+        vsf(stdContainer.begin(), stdContainer.end());
+
+        CHECK(arr[0] == 1);
+        CHECK(arr[1] == 2);
+        CHECK(arr[2] == 3);
+        CHECK(*stdContainer.begin() == 1);
+        CHECK(*++stdContainer.begin() == 2);
+        CHECK(*++++stdContainer.begin() == 3);
```
arnavb: Any way you could improve the readability of this (With a comment)?

code/sorting/sort_test/sort_test.cpp
```diff
@@ -2,109 +2,163 @@
...
 
-            // random elems
-            auto actualIt = actualVec.begin(), expectIt = expectVec.begin();
+            int *actualDynamicArray = new int[rand];
+            ContainerType<int> actualStdContainer{}, expectStdContainer{};
+
+            // randomize elems
+            std::back_insert_iterator<ContainerType<int>> actualIt(actualStdContainer), expectIt(expectStdContainer);
             for (int i = 0; i < rand; ++i, ++actualIt, ++expectIt) {
                 int rv = std::rand()%100;
```
arnavb: Spaces around the operators. Also, this form of random is not recommended.

code/sorting/bubble_sort/bubble_sort.cpp
iattempt: In my opinion the old version is more clear than `!swapped`, since the `!` is not easy to see.

code/sorting/sort_test/sort_test.cpp
iattempt: 😅 I updated, please review.

code/sorting/sort_test/sort_test.cpp
iattempt: "Also, this form of random is not recommended."
I don't understood your mean.
did you say the variable `rand` name is not recommend?

arnavb: No, that, and C style random number generation is pretty slow, although it shouldn't matter here.

code/sorting/bubble_sort/bubble_sort.cpp
arnavb: I just went by the idea that while(!swapped) could  be read as: while not swapped (makes more sense)

Issue #3113, opened 2017-12-03
Add Palindome.py and palindrome_substring.py


engrravijain:
**This is a(n):**

- [x ] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**
These two programs are for finding weather a string is a palindrome or not while the other finds all the sub strings of a given string that are palindrome

to be added in utility/palindrome

iattempt:
The first file existed in string_algorithms, I moved into utility in P.R. above.

PR #3119, opened 2017-12-06
implement palindrome checker in python

iattempt:
**Fixes issue:** 
#3113 

**Changes:**


code/utility/palindrome/palindrome_check/palindrome_check.py
```diff
@@ -0,0 +1,26 @@
+def is_palindrome(arr):
```
arnavb: I remember already implementing this in the past: https://github.com/OpenGenus/cosmos/blob/master/code/string_algorithms/palindrome_checker/palindrome.py. The only thing that needs to be fixed is the recursive implementation.

arnavb: You could consider moving all the files there to palindrome_check.

iattempt: I remove my file, and move all palindrome from string_algos. to utiility

iattempt:
The base commit  from this branch is mixed other, I will create new P.R. for this

PR #3138, opened 2017-12-18
Runge–Kutta method in Python

OkarinRintaro:
Classical Runge–Kutta method
Changes:

Created rk4.py

### DIFF
--- a/code/numerical_analysis/runge_kutt/src/rk4.py
+++ b/code/numerical_analysis/runge_kutt/src/rk4.py
@@ -15,6 +15,7 @@ def runge_kutt(start_conditions, finish_interval, step, f):
         res_x += step
     return res
 
+
 print('Enter start condition: ')
 x0 = float(input('X= '))
 y0 = float(input('Y= '))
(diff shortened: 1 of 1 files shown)


code/numerical_analysis/runge_kutt/src/rk4.py
```diff
@@ -0,0 +1,29 @@
...
+    return res
+
+print('Enter start condition: ')
+x0 = float(input('X= '))
+y0 = float(input('Y= '))
+st_cond = [x0, y0]
+fin_int = float(input('Enter finish interval: '))
+stp = float(input('Enter step: '))
+res = runge_kutt(st_cond, fin_int, stp, lambda x, y: math.pow(x, 2) * (3 * y + math.exp(math.pow(x, 3))))
+i = 0
```
arnavb: How about:

```Python
for i in range(0, len(res)):
    result_x, result_y = res[i]
    print("X= %8.6f, Y= %8.6f" % (result_x, result_y))
```

Then you wouldn't need the extra variable, and you would not need to manually increment it.

OkarinRintaro: Thank you, very much! I didn't notice.

AdiChat:
Looks great 😍 

PR #3145, opened 2017-12-22
Implement kruskal algorithm and dynamic union sort in c++

iattempt:
**Fixes issue:**

**Changes:**

### DIFF
--- a/code/greedy_algorithms/src/kruskal_minimum_spanning_tree/kruskal.cpp
+++ b/code/greedy_algorithms/src/kruskal_minimum_spanning_tree/kruskal.cpp
@@ -1,0 +1,52 @@
+#include <algorithm>
+#include <vector>
+
+namespace kruskal_impl{
+    using namespace std;
+
+    template<typename _ForwardIter, typename _ValueType>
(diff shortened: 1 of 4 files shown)


code/greedy_algorithms/src/kruskal_minimum_spanning_tree/kruskal.cpp
```diff
@@ -0,0 +1,52 @@
...
+         typename _EdgeFunc,
+         typename _UnionFind>
+std::vector<typename _HeapedContainer::value_type>
+kruskal(const size_t minimum_edge_count,
+        _HeapedContainer edges,
+        _HeapCompare compare,
+        _EdgeFunc ef,
+        _UnionFind union_sets)
+{
+    using namespace std;
```
arnavb: What advantage does this give you? You use something with std only once/twice, so writing std:: would be clearer in the long run.

iattempt: I think this  implement is not good enough, so I will take care of it next time.

AdiChat:
Looks great 😍 

PR #3150, opened 2017-12-24
Rewrote mergesort in Python

oddProton:
Removed excess variables and ifs, added comments, added test.

**Fixes issue:**

Fixes unnamed variables by removing them. Fixes if blocks by removing them and abstracting insertion and removal of smallest element. 

**Changes:**
As above, kept only 1 function, added test.


AdiChat:
Looks great 😍 

PR #3151, opened 2017-12-25
Added min_heap.c

YashovardhanSiramdas:
**Changes:**
Added Min Heap in C

### DIFF
--- a/code/data_structures/src/tree/heap/min_heap/min_heap.c
+++ b/code/data_structures/src/tree/heap/min_heap/min_heap.c
@@ -1,7 +1,7 @@
 #include <stdio.h>
 
 #define asize 100
-#define min(a, b) ((a < b) ? a : b)
+#define min(a, b) (((a) < (b)) ? (a) : (b))
 
 struct heap
 {
@@ -23,7 +23,7 @@ heapInsert(int key)
 	++H.hsize;
 	H.a[H.hsize] = key;
 	int i = H.hsize;
-	while(i > 1 && H.a[i] < H.a[i / 2]) {
+	while (i > 1 && H.a[i] < H.a[i / 2]) {
 		swap(&H.a[i], &H.a[i / 2]);
 		i = i / 2;
 	}
@@ -35,14 +35,14 @@ displayHeap()
 {
 	int i;
 	printf("Elements in the heap:- \n");
-	for(i = 1; i <= H.hsize; i++)
+	for (i = 1; i <= H.hsize; i++)
 		printf("%d ", H.a[i]);
 	printf("\n");
 }
 
 void delMin()
 {
-	if(H.hsize == 0) {
+	if (H.hsize == 0) {
 		printf("Heap is empty,No element can be popped out \n");
 		return;
 	}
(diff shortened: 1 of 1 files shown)


code/data_structures/src/tree/heap/min_heap/min_heap.c
```diff
@@ -0,0 +1,110 @@
...
+	*b = *a;
+	*a = temp;
+}
+
+int 
+min(int a, int b)
+{
+	return (a < b ? a : b);	
+}
+
```
oddProton: You could write this as a macro since it doesn't do much and is called often.

code/data_structures/src/tree/heap/min_heap/min_heap.c
```diff
@@ -0,0 +1,110 @@
...
+	int x = H.a[1];
+	H.a[1] = H.a[H.hsize];
+	--H.hsize;
+	int i = 1;
+	
+	while((2 * i) <= H.hsize) {
+		if((2 * i) + 1 <= H.hsize && H.a[i] > min(H.a[2 * i], H.a[(2 * i) + 1])) {
+			if(H.a[(2 * i) + 1] < H.a[2 * i]) {
+				swap(&H.a[i], &H.a[(2 * i) + 1]);
```
oddProton: ``` swap(&H.a[i], &H.a[(2 * i) + 1]); ```

You could write this as macro as well.

    #define swap (array,first,second) array[second] = array[first] + 0 * ( array[first] = array[second] )
    swap(H.a , i , 2*i+1);

oddProton:
Also, the heap is global. It'd be better if the heap was allocated dynamically and passed as pointer to functions.

YashovardhanSiramdas:
@oddProton Thanks for your suggestions ! :) But I have been using functions instead of macros since beginning for more simplicity and readability purpose.Also,I declared heap as global to avoid any unnecessary confusion for readers.Passing array with pointers might cause confusion to some readers.Keeping it as global will help readers to easily focus on the algorithm of Min Heap.

 @arnavb Are those utility functions fine, or should I change them to macros ? I think it is better if the heap is Global.

oddProton:
There is no issue with the utility functions, it was just suggestions if you wanted to. I think that in a language like C and Cpp it's important to focus on the pointers and where things can go wrong instead of just the algorithm. I'd suggest to make the functios take a pointer as parameter and to avoid globals.

arnavb:
@YashovardhanSiramdas I suggest using macros only for the min function, using it for the swap one seems a bit overboard.

YashovardhanSiramdas:
@arnavb I changed min function to macro. Does it look good now?

code/data_structures/src/tree/heap/min_heap/min_heap.c
```diff
@@ -1,6 +1,7 @@
 #include <stdio.h>
 
 #define asize 100
+#define min(a, b) ((a < b) ? a : b)
```
arnavb: Every time you reference a and b inside the macro, put parentheses around them.

YashovardhanSiramdas:
@arnavb Added parentheses inside macro.

AdiChat:
Looks great 😍 

PR #3155, opened 2017-12-26
Rename files and directories to remove illegal characters

InfiniteCoder:
**Changes:**

Changed 3 file names so that they do not contain spaces or quotes.
For simplicity, it would be best to limit file and directory names to only alphanumeric characters(even better to limit them to A-Z a-z and 0-9, that is, don't allow other languages), dashes and underscore.


InfiniteCoder:
@iattempt  The problem with that file is that there is already another file with similar name, but an underscore in the name. Would be nice if only one of them is kept. I am not very good with python, so I don't know which one is better(and I didn't wanted to delete the programs)

iattempt:
In my opinion, the [old](https://github.com/OpenGenus/cosmos/blob/master/code/mathematical_algorithms/src/amicable_numbers/amicable_numbers.py) file is better than [new](https://github.com/OpenGenus/cosmos/blob/master/code/mathematical_algorithms/src/amicable_numbers/amicable%20numbers.py) since the latter looks like a practice.
But the old version should add a condition: `if __name__ == '__main__'` before the `print` test.

InfiniteCoder:
@iattempt  Can you make that change to this PR if possible?

iattempt:
@InfiniteCoder ok, please merge this https://github.com/InfiniteCoder/cosmos/pull/1

AdiChat:
Looks great 😍 

Issue #3159, opened 2017-12-27
Implementation of AES-128, AES-192 and AES-256 in C#


abdouskamel:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

Hello everybody. This one is an implementation of AES-128, AES-192 and AES-256 in C#, with an example of how to cipher a stream. 

Hope you enjoy it.

abdouskamel:
See pull request #3160

PR #3160, opened 2017-12-27
Implementation of AES-128, AES-192 and AES-256 in C#

abdouskamel:
**Fixes issue:**

#3159 

**Changes:**

This one is an implementation of AES-128, AES-192 and AES-256 in C#, with an example of how to cipher a stream.

Hope that helps.


AdiChat:
Looks great 😍 

PR #3137, opened 2017-12-18
Added combsort,egyptian fraction and minimum coins in C

YashovardhanSiramdas:
**Changes:**
Added 3 new files in C:-
1)Comb Sort
2)Egyptian Fraction
3)Minimum Coins

PR #3166, opened 2017-12-30
Added Prime Factors of N in C

YashovardhanSiramdas:
**Changes:**
Added a program which prints all prime factors of any given Natural Number in C.

### DIFF
--- a/code/mathematical_algorithms/src/prime_numbers_of_n/prime_numbers_of_n.c
+++ b/code/mathematical_algorithms/src/prime_numbers_of_n/prime_numbers_of_n.c
@@ -1,13 +1,15 @@
 # include <stdio.h>
+# include <stdlib.h>
 # include <math.h>
 
-int 
-primeFactors(int number, int prime_factors[100])
+int*
+primeFactors(int number)
 {
+    int* prime_factors = (int *)malloc(100 * sizeof(int));
     int i, j = -1;
     while (number % 2 == 0) {
-    	++j;
-    	prime_factors[j] = 2;
+        ++j;
+        prime_factors[j] = 2;
         number = number / 2;
     }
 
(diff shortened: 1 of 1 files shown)


YashovardhanSiramdas:
@InfiniteCoder Thanks for the suggestions !! :) I've renamed the variable "n" to "number". I'm deliberately declaring looping variables outside the loop as @arnavb mentioned earlier (#3137 ) that old C compilers will throw up a compilation error if looping variables are declared within the loop.

code/mathematical_algorithms/src/prime_numbers_of_n/prime_numbers_of_n.c
```diff
@@ -0,0 +1,35 @@
...
+
+    for (i = 3; i <= sqrt(number); i = i + 2) {
+        while (number % i == 0) {
+            printf("%d ", i);
+            number = number / i;
+        }
+    }
+
+    if (number > 2)
+        printf ("%d ", number);
```
arnavb: I think returning a value would be better than printing one.

YashovardhanSiramdas: The variable "number" is getting modified inside the function multiple times. Printing prime factors is happening at 3 places (line 9,15,21)inside the function.So, I think it is not possible to return the value to main function at line 21. Moreover, the code is more readable if a single function prints all prime factors.

YashovardhanSiramdas: @arnavb What do you suggest about it? Is this program fine?

arnavb: I meant returning an array of prime numbers should be better. A function should only have 1 job.

YashovardhanSiramdas:
@arnavb Now the function stores prime factors in array. Does this look good?

YashovardhanSiramdas:
@arnavb I made suggested changes. Function will now return an array pointer.

arnavb:
Good job!

AdiChat:
Looks great 😍 

PR #3174, opened 2017-12-31
Link math library while compiling C programs

InfiniteCoder:
**Fixes issue:**

Added the -lm option, so that C programs which use math library compile properly. If any other libraries are needed, do add them.

**Changes:**

 - Added `-lm` flag.


AdiChat:
Looks great 😍 

Issue #3128, opened 2017-12-11
Code quality needs to be maintained


InfiniteCoder:
**This is a(n):**

- [ ] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [X] Proposal to the Repository

**Details:**

There are many programs here which have very poor quality. Programs that use single character variables, those literally abusing the language syntax to make code shorter(and unreadable) and programs using legacy tools/code. In this form, they are unhelpful to anyone. At one point, this repository will become a pile of useless codes, if these codes are not dealt with properly.

1. Really bad code needs to be removed
2. Salvageable code needs to be improved
3. All codes need to follow same guidelines
4. Whatever tools from GitHub marketplace that are applicable here need to be used. Especially those that ensure code quality.

I think it might even be a good idea to pause accepting new codes for a while until the existing are fixed. Then set up strict guidelines and a proper workflow for accepting submissions, and accept only those codes which are good.

Otherwise, all this effort might go in vain.

the-saif-ahmad:
Not only do we need to fix the existing code, we need to create style guides for all languages to prevent it from happening again. Then even if someone does submit bad code, they can't use the lack of a style guide as an excuse.

AdiChat:
Yes, this is an issue that needs to be attended immediately and the solutions by both of you look promising. 
You must share this at [our new forum](https://discourse.opengenus.org/) where it is more accessible and continue discussing there. 

We must arrive at definite measures to implement as soon as possible. 

**Let us make Cosmos great again.** 💪

InfiniteCoder:
Created a post here https://discourse.opengenus.org/t/fixing-code-quality/47

amirags:
Happy to contribute to this cause!

AdiChat:
That is great @amirags 😃 Go ahead and solve issues one by one 👍 
You can, also, suggest and implement new solutions.

amirags:
Sorry, lying low these days (blame the weather!). Planning to make a list of todos and take it forward. Will add further updates here.

AdiChat:
No problem 👍 🗻  Stay safe and warm 🔥 
Sure. We are looking forward towards your plan. 😃

amirags:
Happy New Year all!

Right! Started having a look at the code, guidelines etc. Adding some points to start with (let me know if discourse is better place for this):

[...]

PR #3175, opened 2018-01-01
Fix compile error

iattempt:
**Fixes issue:**
1. #3128 
2. auto test

**Changes:**
1. [x] fix inclusion error
2. [x] fix some typo
3. [x] add tree serializer
~~4. [ ] fix redefinition of the main function~~

[Done]


iattempt:
@AdiChat please check this error of travis-ci:  0Koci runtime error: exec failed: container_linux.go:265: starting container process caused "could not create session key: disk quota exceeded"

InfiniteCoder:
@iattempt  Looks like an issue with Travis to me. Searching online shows that some other projects have had this issue sometimes. I have rebuilt the jobs, and they are building fine now. By the way, looks like you went into beast mode to fix those issues, great 👍

iattempt:
@InfiniteCoder How can I edit compile arguments/flags or contribute the travis-ci?

iattempt:
@AdiChat 
The compile error of the whole files has been fixed, 
but we got the new issue: [multiple definition of `main'].

To fixed that, I have two ideas, which one is better?

1. remove all main of the src files
2. or add a precondition before main function, like:

```c++
#include <vector>
void sort(std::vector<int> arr)
{
    // ...
}

// the precondition we add
#ifdef COSMOS_DEBUG_MODE
// test driver
int main()
{
    // ...
}
#endif //COSMOS_DEBUG_MODE
```

InfiniteCoder:
@iattempt  There are two files, `.travis.yml` contains the configuration for travis builds, such as what compiler to use, what command to run, and so on. It contains the `make c` and `make cpp` commands. These commands refer to the `makefile` file.

I think we'd have to change the way makefile compiles the program. Currently, it finds all the files with .c extention(or .cpp for C++), and compiles them together. Have to see if there is a way to seperately compile each file.

iattempt:
@InfiniteCoder 
Because the files is part lib which will be included to one's project,
I think we still need to 
1. compile the files together,
2. add a main.cpp to be ignore the error: [no main function],
3. and, however, we will compile the test files each.

InfiniteCoder:
I think you should keep seperate PR for travis. Some more changes might be needed for C as well, and clang needs to be fixed or removed as well. This PR has already become too huge 🐘 

Regarding whether the main function should exist or not, I think it depends upon how the programs are meant to be used. If they are to be used as guides/tutorials, then the current method is appropriate. But if they are meant as library, then not only main methods, but also output to stdout should be removed, as replaced with appropriate comments. I think @AdiChat can tell more about this.

iattempt:
No big, I'll revert changes.

InfiniteCoder:
The object files that are giving errors should be deleted. I don't think there is any need for object files in a repo meant for code. Delete them and the build should succeed.

iattempt:
@InfiniteCoder Did you mean the catch_pch.o? but the catch.hpp is too large, I want to save time on compilation.

AdiChat:
Looks great 😍 

PR #3180, opened 2018-01-02
Update travis.yml for c++

iattempt:
**Fixes issue:**
#3186 
> 3. how to compile code in the src/test?

**Changes:**
- [x] the files in the *src* es will only compile and assemble each, but do not link.
- [x] the files in the *test* will also link, and compile will include automatically pch file which is stored in test/c++/catch_pch.cpp (this means that the coder must use Catch in their testing).


AdiChat:
Looks great 😍 

iattempt:
edecc35 6ba23a4

amirags:
May I suggest including "-std=c++1y" for compiling cpp files? Many of latest commits are using newer c++ features but resulting in build errors with Travis. CXXFLAGS (CXXFLAGS = -Wall -std=c++1y) can be updated for same.

While running makefile on Mac, find complained too, but I was able to fix it locally by following the syntax expected.

iattempt:
@amirags 
In my opinion, we don't need it. those standards are not fully supported even c++11.
Can we see your code?

Issue #3186, opened 2018-01-03
Test rules (the travis-ci) for c++


iattempt:
**This is a(n):**

- [ ] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

##### Fixes issue:

existed:
- [ ] 1. error message of compiler throws: redefinition multiple main functions.
- [ ] 2. some files has stdin/stdout (the `cin` or `cout`, etc.) in the *src* (if the *test*/*src* has been fully separated, then ignore this issue).
> to fix two issues above:

* remove it
* add precondition before it (best, if this would be used for tutorial, the output is useful, and since the c++ is compiled language, the efficiency will not be affected).
* redirect it (redundancy and more complex to modify).

new:
- [ ] 3. forbidding the inclusion (library/header) is dependent.
> the compile will throws error when this error occurs.

[some discussion refer to](https://github.com/OpenGenus/cosmos/pull/3175)

PR #3193, opened 2018-01-07
Add c++ makefiles for Travis-CI

iattempt:
**Fixes issue:**

#3186 
> * the make not exit with error code even compile error.

*advantage*:
I don't know how write make to do that retrieve output of shell script to find error,
but this commit is used another way to compile files, it based on 'auto-dependency',
we don't need retrieve error code from shell script in foreach.

*disadvantage*:
the make will stop at any error occurs.

**Changes:**

file info:
- generate_dependencies.make:
    * this file generate dependencies it involve target, prerequisite, and commands.

- testing.make:
     * this file include dependencies which named *.d and make.


iattempt:
@AdiChat Done. please merge this PR.

AdiChat:
Looks great 😍 

PR #3204, opened 2018-01-16
fixes compile error for clang

iattempt:
**Fixes issue:**
#3175 

**Changes:**
1. rearrange some file code
2. remove older script in makefile
3. remove clang support

Fixed ~~**Bottleneck**:~~
According to this page: http://en.cppreference.com/w/cpp/container/list/insert
```
clang++-5.0 -o -Wall -std=c++11 ./code/data_structures/test/list/test_list.cpp
./code/data_structures/test/list/test_list.cpp:856:26: error: variable has incomplete type 'void'
                    auto expectReturnPos = expect.insert(expect.end(), sz, randomValue);
                         ^
./code/data_structures/test/list/test_list.cpp:869:37: error: no viable overloaded '='
                    expectReturnPos = expect.insert(expect.end(), sz, randomValue);
                    ~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/bin/../lib/gcc/x86_64-linux-gnu/4.8/../../../../include/c++/4.8/bits/stl_list.h:125:12: note: candidate function (the implicit copy assignment operator) not viable: cannot convert argument of incomplete type 'void' to 'const std::_List_iterator<int>' for 1st argument
    struct _List_iterator
```
The clang is still not supported this issue.

@AdiChat Should I remove clang and two old script of c++ in the Travis-CI config?

### DIFF
--- a/code/mathematical_algorithms/src/steepest_descent/steepest_descent.cpp
+++ b/code/mathematical_algorithms/src/steepest_descent/steepest_descent.cpp
@@ -33,6 +33,70 @@ double norm(const std::array<double, S> &x, const std::array<double, S> &y)
 }
 
 /*
+Adds up two matrix
+*/
+template <long unsigned int S>
+std::array<double, S> operator+(const std::array<double, S> &x, const std::array<double, S> &y)
+{
+    std::array<double, S> result;
+
+    for (unsigned int i = 0; i < S; ++i)
+    {
+        result[i] = x[i] + y[i];
+    }
+    return std::move(result);
+}
+
+/*
+Subtrack two matrix
+*/
+template <long unsigned int S>
+std::array<double, S> operator-(const std::array<double, S> &x, const std::array<double, S> &y)
+{
+    std::array<double, S> result;
+
+    for (unsigned int i = 0; i < S; ++i)
+    {
+        result[i] = x[i] - y[i];
+    }
+    return std::move(result);
+}
+
+/*
+Multiplication of matrix and constant value
+*/
+template <long unsigned int S>
+std::array<double, S> operator*(const double x, const std::array<double, S> &y)
+{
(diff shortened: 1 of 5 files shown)


code/mathematical_algorithms/src/steepest_descent/steepest_descent.cpp
```diff
@@ -33,57 +33,48 @@ double norm(const std::array<double, S> &x, const std::array<double, S> &y)
...
-std::array<double, S> steepestdescent(const std::array<std::array<double, S>, S> &A, const std::array<double, S> &b,
-    const double e = 0.0001, const double step = 0.1, const unsigned int IterMax = 10000)
+std::array<double, S> operator+(const std::array<double, S> &x, const std::array<double, S> &y)
 {
-    std::array<double, S> x = {0.0};
-    unsigned int k = 0;
+    std::array<double, S> result;
 
-    do
+    for (unsigned int i = 0; i < S; ++i)
```
arnavb: Why not long unsigned int?

iattempt: I don't know, only rearrange order of functions

arnavb: Ok I see.

AdiChat:
`GCC` seems to have better support compared to `clang`, so, removing `clang` should be fine 👍 
I am not sure how the removal would impact the usage though, in my opinion, testing against one compiler should be fine. The change sounds good. 👍

AdiChat:
Looks great 😍 

iattempt:
#3816

PR #3210, opened 2018-01-20
Rosalind challenges

cpackingham:
**Changes:**
Adds rosalind challenges to cosmos


AdiChat:
Hi Cole, you need to create another file named `README.md` which would include the problem statement for `counting_dna_nucleotides` and a link to the problem. This should be in the same folder level as `counting_dna_nucleotides.js`. See [this](https://github.com/OpenGenus/cosmos/blob/master/code/online_challenges/src/project_euler/problem_001/README.md)
Kindly make the above change and it will be good to be merged 👍

PR #3211, opened 2018-01-20
Elixir rosalind problem

cpackingham:
**Changes:**
Added count_dna_nucleotides.exs


abdouskamel:
This pull request has been forgotten. I am not able to review it, is there anyone who can ?

OvermindDL1:
Hmm, I'd probably have written it like (as this is shorter and would run *MUCH* faster:

```elixir
defmodule CountingDnaNuleotides do
  def count_dna_nucleotides(strand, acc \\ %{})
  def count_dna_nucleotides("", acc), do: acc
  def count_dna_nucleotides(<<c::binary-size(1), strand::binary>>, acc), do: count_dna_nucleotides(strand, Map.update(acc, c, 1, &(1+&1))
end
```
Or less cramped as:
```elixir
defmodule CountingDnaNuleotides do
  def count_dna_nucleotides(strand, acc \\ %{})
  def count_dna_nucleotides("", acc), do: acc
  def count_dna_nucleotides(<<c::binary-size(1), strand::binary>>, acc) do
    count_dna_nucleotides(strand, Map.update(acc, c, 1, &(1+&1))
  end
end
```
And if you want to not accept invalid input then just on the final function add a guard of `when c in ["A", "C", "G", "T"]`

PR #3235, opened 2018-01-31
C Code Fixes.

InfiniteCoder:
**Fixes issue:**

Some code fixes in C programs

**Changes:**

Fixed errors and warning in some programs. Deleted a program which was plagiarised.


abdouskamel (dismissed):
Good job !

arnavb (dismissed):

AdiChat:
It is not encouraged to delete files. 
You may modify the source code or add a comment regarding the source (on consulting the concerned contributor) as we encourage distributing credit.

Kindly make the changes and it will be good to be merged 👍

InfiniteCoder:
@AdiChat The problem is that the [original author](http://www.codewithc.com/c-program-for-newton-raphson-method/) has not mentioned any licence, so by default we may consider it as copyright protected.

As [Jeff Atwood says](https://blog.codinghorror.com/pick-a-license-any-license/), 

> Without a license, the code is copyrighted by default. People can read the code, but they have no legal right to use it.

So this program would be incompatible with the license of cosmos, GNU GPLv3. So I'd rather delete a few plagiarised codes than using other people's code illegally.

code/data_structures/src/tree/tree/trie/trie.c
```diff
@@ -7,63 +7,68 @@
...
     trie_node_t *pNode = NULL;
     pNode = (trie_node_t *)malloc(sizeof(trie_node_t));
- 
+    
     if( pNode )
     {
         int i;
         pNode->value   = 0;
- 
+
```
polyzen: Is whitespace on blank lines like this actually wanted?

InfiniteCoder: @polyzen I don't know what the standards are in such cases, but my editor(Kate) adds whitespaces in blank lines when aligning the code. I couldn't find anything about it on the Internet.

phillid: It's subjective, not objective, but trailing whitespace is more often frowned upon than not. Even more frowned upon when a patch clobbers history of lines like this for no reason other than whitespace changes.

code/data_structures/src/list/singly_linked_list/operations/unclassified/union_intersection_in_list.c
```diff
@@ -3,34 +3,37 @@
...
- 
+ *   a linked list*/
+void 
+push(struct Node** head_ref, int new_data);
+
 /* A utility function to check if given data is present in a list */
-bool isPresent(struct Node *head, int data);
- 
+bool 
+isPresent(struct Node *head, int data);
```
phillid: I'm not convinced that putting types on separate lines like this in function prototypes is normal. I have mostly seen prototypes as single lines, with function definitions having the type on the line above.

code/data_structures/src/tree/tree/trie/trie.c
```diff
@@ -7,63 +7,68 @@
 #define ALPHABET_SIZE (26)
 #define INDEX(c) ((int)c - (int)'a')
 #define FREE(p) \
-    free(p);    \
-    p = NULL;
- 
+free(p);    \
+p = NULL;
+
```
phillid: The indentation here was probably intentional, since the statements are part of the `FREE` macro. I don't see what removing the indent adds.

InfiniteCoder: Thanks for noting that. I'll fix it.

code/data_structures/src/list/singly_linked_list/operations/unclassified/union_intersection_in_list.c
InfiniteCoder: Thanks! fixed!

abdouskamel:
@AdiChat The removed code is just a copy-paste of the original, without any credit. I think that even though there were credits, it's not a good practice to provide a copy-paste of a third person code. We can take inspiration from a code found in the Internet, and mention the author, but copy and paste it gives a bad image.

PR #3242, opened 2018-02-07
Added a new Boyer-Moore implementation in C

thatguy227:
**Changes:**
Added a pointer-based implementation of the Boyer-Moore algorithm using the C language.

### DIFF
--- a/code/string_algorithms/src/boyer_moore_algorithm/boyer_moore_algorithm2.c
+++ b/code/string_algorithms/src/boyer_moore_algorithm/boyer_moore_algorithm2.c
@@ -1,0 +1,95 @@
+/* Part of Cosmos by OpenGenus Foundation */
+// Pointer based implementation of the Boyer-Moore String Search Algorithm
+#include <stdio.h>
+#include <string.h>
+
+
+// Number of characters in the text alphabet used by the search and pattern strings
+// For 8-bit strings, use 256
+#define NO_OF_CHARS 256
+
+
+// The core of Boyer-Moore algorithm is the skip value array used to advance more than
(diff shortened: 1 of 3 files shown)


thatguy227:
I will do that after work today. Anything in particular that stands out as non-compliant?


-Gerry


________________________________

code/string_algorithms/src/boyer_moore_algorithm/boyer_moore_algorithm2.c
```diff
@@ -0,0 +1,95 @@
...
+{
+	const char * testText;
+	const char * testPattern;
+	const char * result;
+	int spaces;
+	char * matchMarker = "^";
+
+
+	testText = "A string searching example consisting of this sentence.";
+	testPattern = "sting";
```
abdouskamel: Should be "string" no ?

thatguy227:
No, it should be "sting" so it is found in the text "..consi>>sting<<...


That example is borrowed from the Sedgewick Algorithms textbook.


-Gerry Jacobus


________________________________

code/string_algorithms/src/boyer_moore_algorithm/boyer_moore_algorithm2.c
thatguy227: Needs to be "sting" so it is found in the testText array. The pattern "sting" occurs in the word "consisting" from the text.

abdouskamel: Right, that's a good example. 👍

AdiChat:
Looks great 😍 

PR #3246, opened 2018-02-08
Pro

kanika9068:
**Fixes issue:**

Issue#2690

**Changes:**

Added a code to reverse a user input string in C.

### DIFF
--- a/pr/reverse.c
+++ b/pr/reverse.c
@@ -1,0 +1,20 @@
+#include <stdio.h>
+int main()
+{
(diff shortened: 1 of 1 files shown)


abdouskamel:
Thank you for your contribution, but there exists an implementation in C for this algorithm ([here](https://github.com/OpenGenus/cosmos/blob/master/code/mathematical_algorithms/src/reverse_number/reverse_number.c)).

You should look through the code base, and search for an algorithm that hasn't been implemented yet.

PR #3263, opened 2018-02-12
Adds guard statements to Erlang and Elixir factorial

FrankKair:
**Fixes issue:**
No particular issue.

**Changes:**
Adds `guard` statements to Erlang and Elixir implementations of factorial.

### DIFF
--- a/code/mathematical_algorithms/src/factorial/factorial.erl
+++ b/code/mathematical_algorithms/src/factorial/factorial.erl
@@ -1,5 +1,5 @@
 -module(factorial).
--export([fac/1]).
+-export([factorial/1]).
 
-fac(0) -> 1;
-fac(N) -> N * fac(N-1).
+factorial(0) -> 1;
+factorial(N) when N > 0 -> N * factorial(N - 1).
--- a/code/mathematical_algorithms/src/factorial/factorial.ex
+++ b/code/mathematical_algorithms/src/factorial/factorial.ex
@@ -1,6 +1,6 @@
 defmodule Factorial do 
   def factorial(0), do: 1
-  def factorial(n), do: n * factorial(n-1)
+  def factorial(n) when n > 0, do: n * factorial(n-1)
 end
 
 arg = String.to_integer(List.first(System.argv))


code/mathematical_algorithms/src/factorial/factorial.erl
```diff
@@ -1,5 +1,5 @@
 -module(factorial).
--export([fac/1]).
+-export([factorial/1]).
 
-fac(0) -> 1;
-fac(N) -> N * fac(N-1).
\ No newline at end of file
+factorial(0) -> 1;
```
arnavb: Is there any way to just return 1 for a negative initial argument?

code/mathematical_algorithms/src/factorial/factorial.ex
```diff
@@ -1,7 +1,7 @@
 defmodule Factorial do 
   def factorial(0), do: 1
```
arnavb: Same here

code/mathematical_algorithms/src/factorial/factorial.erl
FrankKair: That would be 

```erlang
factorial(N) when N < 0 -> 1.
```

FrankKair: Do you think it should be handled like so?
Should the answer be yes, I can refactor the code to include the advent of negative numbers.

code/mathematical_algorithms/src/factorial/factorial.ex
FrankKair: Yep

```elixir
def factorial(n) when n < 0, do: 1
```

code/mathematical_algorithms/src/factorial/factorial.erl
arnavb: Yes, it should. You should probably handle all cases.

code/mathematical_algorithms/src/factorial/factorial.ex
arnavb: Yes, it should. You should probably handle all cases.

code/mathematical_algorithms/src/factorial/factorial.erl
OvermindDL1: Shouldn't `<0` cause a matcherror since factorial itself is undefined when `<0`?

arnavb: That works as well. I just want it to be handled.

OvermindDL1: Well a matcherror is thrown/raised when there is no matching function head, so technically it is already handled.  :-)

FrankKair: (This also applies to the **Elixir** implementation).

Currently:

```erlang
factorial(0) -> 1;
factorial(N) when N > 0 -> N * factorial(N - 1).
```

The way is defined right now, if you call `factorial(-1)`, there will be no matching function in this Erlang module, so an error will be thrown.

If we define:

```erlang
factorial(N) when N < 0 -> 1.
```

We'll be returning 1 for every negative number (which... is not the definition of factorial).

FrankKair: It's up to you all, but at least for me, it makes more sense to keep the mathematical definition of factorial, not allowing the user to call factorial of negative numbers.

arnavb: Ok, what I thought would happen was that if you called the function with a negative number, then the program would just not work. (Infinite loop). However, it seems that Elixir handles this very well! (Sorry for my incompetence; I don't know Elixir very well, it's just that I am going by general programming paradigms that _all_ cases should be handled).

FrankKair: Hey, @arnavb, no problem. Erlang and Elixir are not very common programming languages and they have some different concepts!

OvermindDL1: Heh yep, they are fully immutable functional languages that also have exceptions, and running through a set of matchers with no match ends up raising a `MatchError`.  :-)

They are languages designed to get real work done, so they have a lot of ways of doing things and are in general robust.  :-)

FrankKair:
@arnavb Is this ready to go, then?

arnavb:
@FrankKair Yup, just wait until @AdiChat merges this!

AdiChat:
Looks great 😍 

PR #3265, opened 2018-02-13
Binary Search Tree

varshika15:
A binary search tree (BST), sometimes also called an ordered or sorted binary tree, is a node-based binary tree data structure which has the following properties:
i) The left subtree of a node contains only nodes with keys less than the node’s key.
ii) The right subtree of a node contains only nodes with keys greater than the node’s key.
iii) The left and right subtree must each also be a binary search tree.
iv) There must be no duplicate nodes.
Whenever an element is to be searched, start searching from the root node. Then if the data is less than the key value, search for the element in the left subtree. Otherwise, search for the element in the right subtree.
**Fixes issue:**

**Changes:**

### DIFF
--- a/code/data_structures/src/tree/binary_tree/binary_tree/Binary Search Tree
+++ b/code/data_structures/src/tree/binary_tree/binary_tree/Binary Search Tree
@@ -1,0 +1,159 @@
+import java.util.Scanner;
+
+public class BinarySearchTree {
+	public static  Node root;
+	public BinarySearchTree(){
+		BinarySearchTree.root = null;
+	}
+	
+	public boolean find(int value){
+		Node current = root;
+		while(current!=null){
+			if(current.data==value){
+				return true;
+			}else if(current.data>value){
+				current = current.left;
+			}else{
+				current = current.right;
+			}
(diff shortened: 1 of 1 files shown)


YashovardhanSiramdas:
Make sure that you add the code in respective folder and save it with appropriate extension !

varshika15:
Sorry for the mistake! I am a newbie. Should I create a new folder of binary search tree and then add this file?

YashovardhanSiramdas:
Folder for binary search tree already exists. ```/code/data_structures/src/tree/binary_tree/binary_tree/tree```. Add your code in this path and save it with .java extension.

abdouskamel:
@varshika15 You should also name your file : binary_search_tree.java.

code/data_structures/src/tree/binary_tree/binary_tree/Binary Search Tree
```diff
@@ -0,0 +1,159 @@
+import java.util.Scanner;
```
arnavb: Add a file extension, use underscores, and all lowercase letters.

PR #3285, opened 2018-02-18
search_element.java

varshika15:
Whenever an element is to be searched, start searching from the root node. Then if the data is less than the key value, search for the element in the left subtree. Otherwise, search for the element in the right subtree.

**Fixes issue:**
Searches element in the binary search tree.

**Changes:**


code/data_structures/src/tree/binary_tree/binary_tree/tree/search_element.java
```diff
@@ -0,0 +1,159 @@
+import java.util.Scanner;
```
abdouskamel: Your code is actually badly organized. You should look at the java style guide.

PR #3289, opened 2018-02-20
Add and change search functions

iattempt:
**Fixes issue:**

**Changes:**

1. fix coding style
2. rename file
3. fix import location
4. change algorithm


code/search/src/exponential_search/exponential_search.py
```diff
@@ -33,4 +33,4 @@ def exponential_search(arr, element):
     while i < arrLength and arr[i] <= element:
         i *= 2
 
-    return _binary_search(arr, elment, i / 2, min(i, arrLength))
```
iattempt: fix typo

code/search/src/binary_search/binary_search.py
```diff
@@ -3,26 +3,29 @@
...
+def _binary_search_recursive_impl(arr, x, l, r):
+    if r > l:
         mid = l + int((r - l) / 2)
 
         if arr[mid] == x:
             return mid
 
         elif arr[mid] > x:
-            return binarySearchRecursive(arr, l, mid - 1, x)
+            return _binary_search_recursive_impl(arr, x, l, mid - 1)
```
arnavb: According to the [PEP 8](https://www.python.org/dev/peps/pep-0008/) style guide, I don't think you should be using a leading underscore for the name.

code/search/src/binary_search/binary_search.py
```diff
@@ -32,13 +35,11 @@ def binarySearchLoop(arr, l, r, x):
         else:
             l = mid + 1
 
-    return -1
+    return l if l >= 0 and l < len(arr) and arr[l] == x else -1
```
arnavb:
```Python
return l if l in range(0, len(arr)) # ...
```

code/search/src/exponential_search/exponential_search.py
```diff
@@ -0,0 +1,34 @@
...
+        if arr[mid] == element:
+            return mid
+
+        elif arr[mid] > element:
+            return _binary_search(arr, element, left, mid - 1)
+
+        else:
+            return _binary_search(arr, element, mid + 1, right)
+
+    return left if left >= 0 and left < len(arr) and arr[left] == element else -1
```
arnavb: Same here

code/search/src/binary_search/binary_search.py
iattempt: I think the method is private and [PEP 8 - Method names "Use one leading underscore only for non-public methods"](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables), thus I added leading underscore.

iattempt:
@arnavb Fixed, please review 😊.

AdiChat:
Looks great 😍 

Issue #3314, opened 2018-02-26
Read me for all folders in 'code' folder


varshika03:
[ ]  New algorithm
[x] Update to an existing algorithm
[ ]  Error
[ ]  Proposal to the Repository

I wish to add read me files for code categories as well as the algorithms within the categories for which they are missing.

vandana1212:
I would also like to work with my partner for RGSoC @varshika03 for the same issue.

arnavb:
By all means, go for it! 😄

nikitajuneja1:
I would also like to work with my partner for RGSoC'20 @vagishajain for the same issue. Please assign it to me!

PR #3319, opened 2018-02-26
Updated Readme.md of Search algorithms

vandana1212:
Issue #3314

Updated Read me file


code/search/src/README.md
```diff
@@ -1,2 +1,2 @@
-# cosmos
-Your personal library of every algorithm and data structure code that you will ever encounter
+# Searching algorithms
```
arnavb: There needs to be more data added for searching in this file. Currently, it is really less. Also, take what you removed and add it to the bottom of the file.

PR #3335, opened 2018-02-28
Add matrix transformation functions

iattempt:
**Fixes issue:**

**Changes:**

### DIFF
--- a/code/graph_algorithms/src/matrix_transformation/matrix_transformation.swift
+++ b/code/graph_algorithms/src/matrix_transformation/matrix_transformation.swift
@@ -1,0 +1,62 @@
+// Part of Cosmos by OpenGenus Foundation
+
+import Foundation
+
+class MatrixTransformation {
+    func rotate<T>(_ matrix: inout [[T]]) {
+        guard matrix.count > 0 && matrix.count == matrix[0].count else {
+            return
+        }
+
+        // 1 2 3      7 8 9      7 4 1
+        // 4 5 6  =>  4 5 6  =>  8 5 2
+        // 7 8 9      1 2 3      9 6 3
+        matrix.reverse()
+        symmetric(&matrix)
+    }
+
+    func reverse_rotate<T>(_ matrix: inout [[T]]) {
+        guard matrix.count > 0 && matrix.count == matrix[0].count else {
+            return
+        }
+
+        // 1 2 3      3 2 1      3 6 9
+        // 4 5 6  =>  6 5 4  =>  2 5 8
+        // 7 8 9      9 8 7      1 4 7
+        for i in 0..<matrix.count {
+            matrix[i].reverse()
+        }
+        symmetric(&matrix)
+    }
+
+    func symmetric<T>(_ matrix: inout [[T]]) {
+        guard matrix.count > 0 && matrix.count == matrix[0].count else {
+            return
+        }
+
+        // 1 2 3      1 4 7
+        // 4 5 6  =>  2 5 8
+        // 7 8 9      3 6 9
+        for i in 0..<matrix.count {
+            for j in i..<matrix.count {
+                (matrix[i][j], matrix[j][i]) = (matrix[j][i], matrix[i][j])
+            }
+        }
+    }
+
+    func reverse_symmetric<T>(_ matrix: inout [[T]]) {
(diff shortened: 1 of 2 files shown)


code/graph_algorithms/src/matrix_transformation/matrix_transformation.swift
```diff
@@ -0,0 +1,62 @@
...
+        // 1 2 3      3 2 1      3 6 9
+        // 4 5 6  =>  6 5 4  =>  2 5 8
+        // 7 8 9      9 8 7      1 4 7
+        for i in 0..<matrix.count {
+            matrix[i].reverse()
+        }
+        symmetric(&matrix)
+    }
+
+    func symmetric<T>(_ matrix: inout [[T]]) {
```
Vitorvgc: Maybe `transpose` would be better? The only problem is that the transpose is defined for any matrix and not only square ones.

Maybe `reflect` is more proper, as it is the reflection of the matrix over its main diagonal.

code/graph_algorithms/src/matrix_transformation/matrix_transformation.swift
```diff
@@ -0,0 +1,62 @@
...
+        // 4 5 6  =>  2 5 8
+        // 7 8 9      3 6 9
+        for i in 0..<matrix.count {
+            for j in i..<matrix.count {
+                (matrix[i][j], matrix[j][i]) = (matrix[j][i], matrix[i][j])
+            }
+        }
+    }
+
+    func reverse_symmetric<T>(_ matrix: inout [[T]]) {
```
Vitorvgc: If `reflect` is used to name the `symmetric` funtion, here it could be named `anti_reflect` or something like that to refer to the anti-diagonal.

Vitorvgc:
I find it a bit strange for these algorithms to be in the graphs folder. Although graphs can be represented as adjacency matrices, these algorithms aren't direct applications of graphs. I think that this would be more properly placed in the data structures folder.

iattempt:
I also have confused on this problem.

code/graph_algorithms/src/matrix_transformation/matrix_transformation.swift
iattempt: I think transpose is better. Maybe the problem would be fixed by add 'square' as prefix of file/folder name.

iattempt: Because It will be incremented complexity of code and time (maybe) if I implement for any matrix.

iattempt:
@OpenGenus/cosmos-maintainers Please review.

arnavb:
Wow, That's a new \@mention that I've never seen before!

code/graph_algorithms/src/matrix_transformation/matrix_transformation.swift
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        // 1 2 3      7 8 9      7 4 1
+        // 4 5 6  =>  4 5 6  =>  8 5 2
+        // 7 8 9      1 2 3      9 6 3
+        matrix.reverse()
+        square_transpose(&matrix)
+    }
+
+    func reverse_rotate<T>(_ matrix: inout [[T]]) {
```
arnavb: camelCase; See [our style guide](https://github.com/OpenGenus/cosmos/blob/master/guides/coding_style/swift/README.md#naming)

code/graph_algorithms/src/matrix_transformation/matrix_transformation.swift
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        // 1 2 3    1 2 0      1  4  3
+        // 4 5 6  x 0 1 0  =>  4 13  6
+        // 7 8 9    0 0 1      7 22  9
+        let ct = matrix.count
+        var res = [[T]](repeating: [T](repeating: 0, count: ct), count: ct)
+        for i in 0..<matrix.count {
+            for j in 0..<matrix.count {
+                for n in 0..<matrix.count {
```
arnavb: k is the usual name for  nested loop indices 3rd level.

iattempt:
@OpenGenus/cosmos-maintainers What do the others think?

AdiChat:
Looks great 😍 

PR #3343, opened 2018-03-01
Added Project Euler Problem #3 in JS

arnavb:
**Fixes issue:** N/A

**Changes:**
- Added Project Euler problem #3 in JS.


code/online_challenges/src/project_euler/problem_003/problem_003.js
```diff
@@ -0,0 +1,13 @@
+let num = 600851475143;
+let highestPrimeFactor = 0;
+let currentFactor = 2;
+
+while (num !== 1) {
```
abdouskamel: There are cases where num would never be equal to 1, for example num = 4.

code/online_challenges/src/project_euler/problem_003/problem_003.js
```diff
@@ -0,0 +1,13 @@
...
+  if (num % currentFactor === 0 && currentFactor > highestPrimeFactor) {
+    highestPrimeFactor = currentFactor;
```
abdouskamel: currentFactor could be not a prime, for example for num = 8, in the last iteration we get currentFactor = 4.

arnavb: You are right, I should reset the factor at some point.

arnavb:
@abdouskamel Could you review if the algorithm is correct now?

abdouskamel:
@arnavb It still gives 4 for entry 8. The problem is that there's the possibility to affect to `highestPrimeFactor` every divisor of `num` that we found. In fact, we should affect to the first variable a number only if it's a prime.

arnavb:
@abdouskamel You were entirely correct. However, I found another more efficient algorithm that achieves the same purpose. Please review now! Thanks for your effort!

arnavb:
@abdouskamel The checking for that could reduce the performance of the algorithm, so I will leave this as it is. The user should make sure that they pass the correct number.

@AdiChat please merge!

arnavb:
@AdiChat please look at my LGTM PR!

AdiChat:
Sure, I will take a look at it now :+1: 

Looks great 😍 

PR #3348, opened 2018-03-01
suffix_Array.c

Awani-Shukla:
**Fixes issue:**

**Changes:**sorted array of all suffixes of a given string in c.

### DIFF
--- a/code/string_algorithms/src/suffix_array/suffix_array.c
+++ b/code/string_algorithms/src/suffix_array/suffix_array.c
@@ -1,0 +1,61 @@
+#include <iostream>
+#include <cstring>
+#include <algorithm>
+using namespace std;
+
+
+struct suffix
+{
+	int index;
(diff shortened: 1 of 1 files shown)


code/string_algorithms/src/suffix_array/suffix_array.c
```diff
@@ -0,0 +1,61 @@
+#include <iostream>
+#include <cstring>
+#include <algorithm>
+using namespace std;
```
arnavb: C or C++? Pick a language? Currently, this file makes no sense at all.

dragon540: also don't use "using namespace std" you should std::cout(or whatever) .

arnavb: @dragon540 That depends on whether this is C or C++...

PR #3354, opened 2018-03-03
Create centroid_decomposition.cpp

KtlTheBest:
**Fixes issue:**

Code for Centroid Decomposition in C++ without documentation.

**Changes:**

### DIFF
--- a/centroid_decompostition.cpp
+++ b/centroid_decompostition.cpp
@@ -1,70 +1,92 @@
 #include <iostream>
-#include <list>
+#include <vector>
 
 using namespace std;
 
-struct Graph 
+class Graph 
 {
-	int V;
-	list<int> *adj;
-	int *size;
-	bool *deleted;
-
-	Graph (int num) : V(num) {
-		adj = new list<int>[V];
-		size = new int[V];
-		deleted = new bool[V];	
+private:
+	int ver;
+	vector<int>* adj;
+	vector<int>* centroidTree;
+	int* sizes;
+	bool* deleted;
+public:
+	Graph(int num) : ver(num) {
+		adj = new vector<int>(ver + 1);
+		centroidTree = new vector<int>(ver + 1);
+		sizes = new int[ver + 1];
+		deleted = new bool[ver + 1];	
 	}
+	
+	~Graph();
+	
+	void addEdge(int x, int y, vector<int> arr[]);
 
-	void addEdge (int x, int y);
+	void calcSizes(int cur, int par);
 
-	void calcSizes (int cur, int par);
+	int findCentroid(int cur, int par, int vertices);
 
-	int findCentroid (int cur, int par, int vertices);
-
-	void decomposeTree (int cur, int total);
+	void decomposeTree(int cur, int total);
 };
 
-void Graph::addEdge (int x, int y)
+Graph::~Graph()
 {
-	adj[x].push_back(y);
+	delete []adj;
+	delete []sizes;
+	delete []deleted;
 }
 
-void Graph::calcSizes (int cur, int par)
+void Graph::addEdge(int x, int y, vector<int> arr[])
 {
-	size[cur] = 1;
-	list<int>::iterator i;
-	for (i = adj[cur].begin(); i != adj[cur].end(); ++ i)
-	{
-		if (*i == par || deleted[*i] == true) continue;
-		calcSizes (*i, cur);
-		size[cur] += size[*i];
-	}
+	arr[x].push_back(y);
 }
 
-int Graph::findCentroid (int cur, int par, int vertices){
-	list<int>::iterator i;
-	for (i = adj[cur].begin(); i != adj[cur].end(); ++ i)
+void Graph::calcSizes(int cur, int par)
+{
+	sizes[cur] = 1;
+	for (int i = 0; i < adj[cur].size(); ++ i)
 	{
-		if (*i == par || deleted[*i] == true) continue;
-		if (size[*i] > vertices / 2)
+		int to = adj[cur][i];
+		if (!(to == par || deleted[to] == true))
 		{
-			return findCentroid (*i, cur, vertices);
+			calcSizes (to, cur);
+			sizes[cur] += sizes[to];
 		}
 	}
-	return cur;
 }
 
-void Graph::decomposeTree (int cur, int total)
+int Graph::findCentroid(int cur, int par, int vertices){
+	for (int i = 0; i < adj[cur].size(); ++ i)
+	{
+		int to = adj[cur].size();
+		if (!(to == par || deleted[to] == true))
+		{
+			if (sizes[to] > vertices / 2)
+			{
+				return findCentroid(to, cur, vertices);
+			}
+		}
+		return cur;
+	}
+}
+
+void Graph::decomposeTree(int cur, int total)
 {
 	calcSizes (cur, -1);
-	int centroid = findCentroid (cur, -1, size[cur]);
+	int centroid = findCentroid (cur, -1, sizes[cur]);
 	calcSizes (centroid, -1);
 	deleted[centroid] = true;
-	list<int>::iterator i;
-	for(i = adj[centroid].begin(); i != adj[centroid].end(); ++ i)
+	for (int i = 0; i < adj[centroid].size(); ++ i)
 	{
-		if (deleted[*i] == true) continue;
-		decomposeTree(*i, size[*i]);
+		int to = adj[centroid][i];
+		if (deleted[to] == true) continue;
+		decomposeTree(to, sizes[to]);
+		addEdge(cur, to, centroidTree);
 	}
+}
+
+int main()
+{
+	return 0;	
 }
--- a/centroid_decompostition.cpp
+++ b/centroid_decompostition.cpp
@@ -1,0 +1,70 @@
+#include <iostream>
+#include <list>
+
+using namespace std;
+
+struct Graph 
+{
+	int V;
+	list<int> *adj;
+	int *size;
+	bool *deleted;
+
+	Graph (int num) : V(num) {
+		adj = new list<int>[V];
+		size = new int[V];
+		deleted = new bool[V];	
+	}
+
+	void addEdge (int x, int y);
+
+	void calcSizes (int cur, int par);
+
+	int findCentroid (int cur, int par, int vertices);
+
+	void decomposeTree (int cur, int total);
+};
+
+void Graph::addEdge (int x, int y)
+{
+	adj[x].push_back(y);
+}
+
+void Graph::calcSizes (int cur, int par)
+{
+	size[cur] = 1;
+	list<int>::iterator i;
+	for (i = adj[cur].begin(); i != adj[cur].end(); ++ i)
+	{
+		if (*i == par || deleted[*i] == true) continue;
+		calcSizes (*i, cur);
+		size[cur] += size[*i];
+	}
+}
+
+int Graph::findCentroid (int cur, int par, int vertices){
+	list<int>::iterator i;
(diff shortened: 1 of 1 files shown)


centroid_decompostition.cpp
```diff
@@ -0,0 +1,70 @@
+#include <iostream>
+#include <list>
+
+using namespace std;
```
arnavb: Never do this. Also, put this file in the appropriate directory.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+struct Graph 
+{
+	int V;
```
arnavb: Don't use capital v, use a better name, such as vertices.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,70 @@
...
+	list<int> *adj;
+	int *size;
```
arnavb: Don't use pointers for any of these three variables, especially the list (Wrap everything in a `std::vector`). Also, a list seems overkill for this project. Use `std::vector` for that as well. Also, the variable names should be plural, otherwise this is misleading.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,70 @@
+#include <iostream>
+#include <list>
+
+using namespace std;
+
+struct Graph
```
arnavb: Use the class keyword.

Use encapsulation as well. Mark necessary variables `private`.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,70 @@
...
+	void addEdge (int x, int y);
+
+	void calcSizes (int cur, int par);
+
+	int findCentroid (int cur, int par, int vertices);
+
+	void decomposeTree (int cur, int total);
+};
+
+void Graph::addEdge (int x, int y)
```
arnavb: Remove spaces before parentheses here and in declarations.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+	adj[x].push_back(y);
+}
+
+void Graph::calcSizes (int cur, int par)
+{
+	size[cur] = 1;
```
arnavb: Risky, maybe check for Out of bounds, and throw an exception if so.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,70 @@
...
+	list<int>::iterator i;
```
arnavb: Use auto inside the loop. Even better: Use a range based for loop.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,70 @@
...
+	for (i = adj[cur].begin(); i != adj[cur].end(); ++ i)
+	{
+		if (*i == par || deleted[*i] == true) continue;
```
arnavb: Refactor:

```C++
for (const auto& elem : adj[cur]) // If you modify elem, remove the const
{
    if (!(elem == par || deleted[elem]))
    {
        calcSizes (elem, cur);
        size[cur] += size[elem]; // Do bounds checking here as well, unless it hinders performance.
    }
}
```

centroid_decompostition.cpp
```diff
@@ -0,0 +1,70 @@
...
+		calcSizes (*i, cur);
+		size[cur] += size[*i];
+	}
+}
+
+int Graph::findCentroid (int cur, int par, int vertices){
+	list<int>::iterator i;
```
arnavb: Same as function above.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+void Graph::decomposeTree (int cur, int total)
+{
+	calcSizes (cur, -1);
+	int centroid = findCentroid (cur, -1, size[cur]);
+	calcSizes (centroid, -1);
+	deleted[centroid] = true;
+	list<int>::iterator i;
+	for(i = adj[centroid].begin(); i != adj[centroid].end(); ++ i)
```
arnavb: Same as above.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,70 @@
...
+	{
+		if (deleted[*i] == true) continue;
+		decomposeTree(*i, size[*i]);
+	}
+}
```
arnavb: Add a main function, or make this a header file.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,70 @@
...
+		deleted = new bool[V];	
+	}
+
+	void addEdge (int x, int y);
+
+	void calcSizes (int cur, int par);
+
+	int findCentroid (int cur, int par, int vertices);
+
+	void decomposeTree (int cur, int total);
```
arnavb: This class needs  a destructor if it will use dynamic arrays so that no memory is leaked. A better option is std::vector.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,92 @@
...
+{
+private:
+	int ver;
+	vector<int>* adj;
+	vector<int>* centroidTree;
+	int* sizes;
+	bool* deleted;
+public:
+	Graph(int num) : ver(num) {
+		adj = new vector<int>(ver + 1);
```
arnavb: Any particular reason you are using ver + 1? (Don't immediately change anything, I just want to know if that is part of the algorithm)

centroid_decompostition.cpp
```diff
@@ -0,0 +1,92 @@
+#include <iostream>
+#include <vector>
+
+using namespace std;
+
+class Graph 
+{
+private:
+	int ver;
+	vector<int>* adj;
```
arnavb: @KtlTheBest Could you clarify what your goal is here?

Currently, you are trying to set up a system in which you have an array of vectors. A better way to express that would be using:

```C++
int ver_; // Add underscore to end for members
std::vector<std::vector<int>> adj_;
std::vector<std::vector<int> centroidTree_;
std::vector<int> sizes_;
std::vector<bool> marked_; // Deleted is misleading, you may also want a different container, but it doesn't matter here
```

In your constructor using the member initialize syntax:

```C++
Graph(int num) : 
    ver_(num)
    adj_(num+ 1, std::vector<int>), // Don't rely on ver being initialized, even though it is
    centroidTree_(num+ 1, std::vector<int>),
    sizes_(num+ 1),
    marked_(num + 1) // Make sure you really mean num + 1
{}
```

Now you can ditch the constructor.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,92 @@
...
+	void addEdge(int x, int y, vector<int> arr[]);
+
+	void calcSizes(int cur, int par);
+
+	int findCentroid(int cur, int par, int vertices);
+
+	void decomposeTree(int cur, int total);
+};
+
+Graph::~Graph()
```
arnavb: No need for this with vectors.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,92 @@
...
+{
+	delete []adj;
+	delete []sizes;
+	delete []deleted;
+}
+
+void Graph::addEdge(int x, int y, vector<int> arr[])
```
arnavb: Fix this function to:

```C++
void Graph::addEdge(int x, int y, std::vector<std::vector<int>>& graph) // Or whatever name. 
                                                                        // Make sure to pass by reference,
                                                                        // since you will modify the vector
{
    // If you want, you can do bounds checking:
    if (x >= graph.size() || x < 0)
    {
        throw std::out_of_range{"x coordinate out of bounds"}; // http://en.cppreference.com/w/cpp/error/out_of_range
    }
    graph[x].push_back(y);
```

centroid_decompostition.cpp
```diff
@@ -0,0 +1,92 @@
...
+{
+	arr[x].push_back(y);
+}
+
+void Graph::calcSizes(int cur, int par)
+{
+	sizes[cur] = 1;
```
arnavb: Bounds checking....

centroid_decompostition.cpp
```diff
@@ -0,0 +1,92 @@
...
+	for (int i = 0; i < adj[cur].size(); ++ i)
```
arnavb:
```C++
for (const auto& element : adj[cur])
{
    if (!(elem == par || marked_[elem]))
    {
        calcSizes (elem, cur);
        sizes[cur] += sizes[elem]; // Do bounds checking here too
    }
}
```

centroid_decompostition.cpp
```diff
@@ -0,0 +1,92 @@
...
+		if (!(to == par || deleted[to] == true))
+		{
+			calcSizes (to, cur);
+			sizes[cur] += sizes[to];
+		}
+	}
+}
+
+int Graph::findCentroid(int cur, int par, int vertices){
+	for (int i = 0; i < adj[cur].size(); ++ i)
```
arnavb: Use a range based loop here as well. I'll leave that as an exercise to you! 😄

centroid_decompostition.cpp
```diff
@@ -0,0 +1,92 @@
...
+	{
+		int to = adj[centroid][i];
+		if (deleted[to] == true) continue;
+		decomposeTree(to, sizes[to]);
+		addEdge(cur, to, centroidTree);
+	}
+}
+
+int main()
+{
```
arnavb: What I meant in my previous comment was for you to add an example in which you demonstrate each of the functions above working.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,92 @@
...
+	Graph(int num) : ver(num) {
+		adj = new vector<int>(ver + 1);
+		centroidTree = new vector<int>(ver + 1);
+		sizes = new int[ver + 1];
+		deleted = new bool[ver + 1];	
+	}
+	
+	~Graph();
+	
+	void addEdge(int x, int y, vector<int> arr[]);
```
arnavb: Make this function private.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+	void calcSizes(int cur, int par);
```
arnavb: This too.

arnavb:
@iattempt I want to get your review on this, as this user is new and needs help understanding OOP principles. I have given my input, but I feel that is by no means complete; please add something on!

centroid_decompostition.cpp
```diff
@@ -0,0 +1,92 @@
...
+	vector<int>* centroidTree;
+	int* sizes;
+	bool* deleted;
+public:
+	Graph(int num) : ver(num) {
+		adj = new vector<int>(ver + 1);
+		centroidTree = new vector<int>(ver + 1);
+		sizes = new int[ver + 1];
+		deleted = new bool[ver + 1];	
+	}
```
arnavb: Also, use 4 spaces, not tabs

centroid_decompostition.cpp
KtlTheBest: Since indexation starts from 0, you need additional space for Nth node

centroid_decompostition.cpp
KtlTheBest: Thank you for your thorough review, I'll make sure to correct my mistakes. And I'm sorry, I made new pull request for this (mistakenly?) but on the other pull request, the file is in proper folder.

centroid_decompostition.cpp
```diff
@@ -0,0 +1,92 @@
+#include <iostream>
```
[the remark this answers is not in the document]
KtlTheBest: I created new pull request (sorry), where file is in correct directory

centroid_decompostition.cpp
arnavb: I'm not saying you shouldn't use it, but modern C++ has many benefits, and this is undoubtedly (At least to me) easier to read.

PR #3362, opened 2018-03-04
Create centroid_decomposition.cpp

KtlTheBest:
**Fixes issue:**

Got rid of compilation error

**Changes:**

Made file as main function.
Deleted unnecessary spaces, changed list entries to vector, changed iterator to ranges, variable's name.

### DIFF
--- a/code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp
+++ b/code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp
@@ -83,17 +83,8 @@ void Graph::decomposeTree(int cur, int par, int total, int level, std::vector<st
     }
 }
 
-
-int main(){
-    Graph tree(7);
-    Level.resize(8);
-    tree.addEdge(1, 2, tree.adj_);
-    tree.addEdge(1, 3, tree.adj_);
-    tree.addEdge(2, 4, tree.adj_);
-    tree.addEdge(4, 5, tree.adj_);
-    tree.addEdge(4, 6, tree.adj_);
-    tree.addEdge(2, 7, tree.adj_);
-    tree.decomposeTree(1, -1, tree.ver_, 1, tree.adj_);
+void showOutput()
+{
     for (int i = 1; Level[i].empty() == false; ++i)
     {
         std::cout << "Centroids of level " << i << ": ";
@@ -109,5 +100,18 @@ int main(){
         }
         std::cout << '\n';
     }
+}
+
+int main(){
+    Graph tree(7);
+    Level.resize(8);
+    tree.addEdge(1, 2, tree.adj_);
+    tree.addEdge(1, 3, tree.adj_);
+    tree.addEdge(2, 4, tree.adj_);
+    tree.addEdge(4, 5, tree.adj_);
+    tree.addEdge(4, 6, tree.adj_);
+    tree.addEdge(2, 7, tree.adj_);
+    tree.decomposeTree(1, -1, tree.ver_, 1, tree.adj_);
+    showOutput();
     return 0;
 }
--- a/code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp
+++ b/code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp
@@ -1,66 +1,58 @@
 #include <iostream>
 #include <vector>
 
-using namespace std;
-
 class Graph 
 {
-private:
-	int ver;
-	vector<int>* adj;
-	vector<int>* centroidTree;
-	int* sizes;
-	bool* deleted;
-public:
-	Graph(int num) : ver(num) {
-		adj = new vector<int>(ver + 1);
-		centroidTree = new vector<int>(ver + 1);
-		sizes = new int[ver + 1];
-		deleted = new bool[ver + 1];	
-	}
-	
-	~Graph();
-	
-	void addEdge(int x, int y, vector<int> arr[]);
+	int ver_;
+	std::vector<std::vector<int>> adj_;
+	std::vector<std::vector<int>> centroidTree_;
+	std::vector<int> sizes_;
+	std::vector<bool> marked_;
+    
+	Graph(int num) : 
+        ver_(num),
+		adj_(num + 1, std::vector<int>),
+		centroidTree_(num + 1, std::vector<int>),
+		sizes_(num + 1),
+		marked_(num + 1),	
+    {}
 
-	void calcSizes(int cur, int par);
-
+	void addEdge(int x, int y, std::vector<std::vector<int>> &graph);
+	void calcSizes(int cur, int par, std::vector<std::vector<int>> graph);
 	int findCentroid(int cur, int par, int vertices);
-
 	void decomposeTree(int cur, int total);
 };
 
-Graph::~Graph()
+void Graph::addEdge(int x, int y, std::vector<std::vector<int>> &graph)
 {
-	delete []adj;
-	delete []sizes;
-	delete []deleted;
-}
-
-void Graph::addEdge(int x, int y, vector<int> arr[])
-{
+    if (x < 0 || x > ver_)
+    {
+        throw std::out_of_range{"x is out of boundaries"};
+    }
+    if (y < 0 || y > ver_)
+    {
+        throw std::out_of_range{"y is out of boundaries"};
+    }
 	arr[x].push_back(y);
 }
 
-void Graph::calcSizes(int cur, int par)
+void Graph::calcSizes(int cur, int par, std::vector<std::vector<int>> graph)
 {
-	sizes[cur] = 1;
-	for (int i = 0; i < adj[cur].size(); ++ i)
+	sizes_[cur] = 1;
+	for (const auto& to : graph[cur])
 	{
-		int to = adj[cur][i];
-		if (!(to == par || deleted[to] == true))
+		if (!(to == par || marked_[to] == true))
 		{
 			calcSizes (to, cur);
-			sizes[cur] += sizes[to];
+			sizes_[cur] += sizes_[to];
 		}
 	}
 }
 
-int Graph::findCentroid(int cur, int par, int vertices){
-	for (int i = 0; i < adj[cur].size(); ++ i)
+int Graph::findCentroid(int cur, int par, int vertices, std::vector<std::vector<int>> graph){
+	for (const auto& to : graph[cur])
 	{
-		int to = adj[cur].size();
-		if (!(to == par || deleted[to] == true))
+		if (!(to == par || marked_[to] == true))
 		{
 			if (sizes[to] > vertices / 2)
 			{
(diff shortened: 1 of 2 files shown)


KtlTheBest:
When I try to construct members (std::vector<std::vector<int>>) compilator throws errors. I don't understand why.

KtlTheBest:
@arnavb Can you please once again review my pull request?)

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.hpp
```diff
@@ -0,0 +1,81 @@
+#include <iostream>
+#include <vector>
+
+class Graph 
+{
+	std::vector<int> empt;
```
arnavb: call it `empty_`

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.hpp
```diff
@@ -0,0 +1,81 @@
...
+	int ver_;
+	std::vector<std::vector<int>> adj_;
+	std::vector<std::vector<int>> centroidTree_;
+	std::vector<int> sizes_;
+	std::vector<bool> marked_;
+    
+	Graph(int num) :
+        ver_(num),
```
arnavb: Indent this.

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.hpp
```diff
@@ -0,0 +1,81 @@
...
+class Graph 
+{
+	std::vector<int> empt;
+	int ver_;
+	std::vector<std::vector<int>> adj_;
+	std::vector<std::vector<int>> centroidTree_;
+	std::vector<int> sizes_;
+	std::vector<bool> marked_;
+    
+	Graph(int num) :
```
arnavb: None of these functions are public, As such, no one outside your class will be able to access them. Do this:

```C++
    std::vector<bool> marked_;

public:
    // All of your functions, including constructor
```

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.hpp
```diff
@@ -0,0 +1,81 @@
...
+{
+    if (x < 0 || x > ver_)
+    {
+        throw std::out_of_range{"x is out of boundaries"};
+    }
+    if (y < 0 || y > ver_)
+    {
+        throw std::out_of_range{"y is out of boundaries"};
+    }
+	graph[x].push_back(y);
```
arnavb: Unindent this.

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.hpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+void Graph::calcSizes(int cur, int par, std::vector<std::vector<int>> graph)
+{
+	sizes_[cur] = 1;
+	for (const auto& to : graph[cur])
+	{
+		if (!(to == par || marked_[to] == true))
```
arnavb: No need for the `== true`, which is implicit.

arnavb:
I am confused about one thing. When you pass graph as a parameter to all of your functions, is it supposed to represent your whole graph class?

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.hpp
KtlTheBest: @arnavb I wrote it as such, so code would be more understandable (when you are fast skimming through the code this way it will be easily readable instead without "== true"), but if it is needed to be as such I will delete it

KtlTheBest:
@arnavb There is a function `addEdge` which requires specifying which graph you are adding to, so I deemed it necessary to pass on which graph you are performing your actions on. But while writing this I understood what you meant...

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.hpp
arnavb: @KtlTheBest I feel that when I read `|| marked_[to]`, the code is checking whether the edge (?) is marked, versus checking whether it is equal to true. It's like saying this this:

```C++
while (!gameOver)
```
is clearer than:
```C++
while (gameOver == false)
```

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.hpp
```diff
@@ -0,0 +1,80 @@
...
+#include <vector>
+
+class Graph 
+{
+    int ver_;
+    std::vector<std::vector<int>> adj_;
+    std::vector<std::vector<int>> centroidTree_;
+    std::vector<int> sizes_;
+    std::vector<bool> marked_;
+    public:
```
arnavb: Unindent this one level.

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.hpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+void Graph::calcSizes(int cur, int par, std::vector<std::vector<int>> graph)
+{
+    sizes_[cur] = 1;
+    for (const auto& to : graph[cur])
+    {
+        if (!(to == par || marked_[to] == true))
+        {
+            calcSizes (to, cur, graph);
```
arnavb: I hope this doesn't cause infinite recursion...

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.hpp
```diff
@@ -0,0 +1,80 @@
...
+            sizes_[cur] += sizes_[to];
+        }
+    }
+}
+
+int Graph::findCentroid(int cur, int par, int vertices, std::vector<std::vector<int>> graph){
+    for (const auto& to : graph[cur])
```
arnavb: Are you sure you necessarily need a loop here? You are only accessing the first element of `graph[cur]`.

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.hpp
```diff
@@ -0,0 +1,80 @@
...
+    marked_[centroid] = true;
+    for (const auto& to : adj_[cur])
+    {
+        if(!(to == par || marked_[to] == true))
+        {
+            decomposeTree(to, cur, sizes_[to], graph);
+            addEdge(cur, to, centroidTree_);
+        }
+    }
+}
```
arnavb: Again, you are missing a use case. Add a main function (Or create another test file), in which you create a graph, and test every one of these member functions.

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.hpp
KtlTheBest: No, it doesn't. Since it is a tree, every node has one parent. We check if our node is parent or not and if it isn't and not marked, it is a go.

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.hpp
KtlTheBest: Yes, when we are in a subtree, we must find new centroid, and in case the current node we are in is not a centroid of current subtree, we must search for another centroid. Therefore, loop is necessary.

KtlTheBest:
@arnavb There is also code for centroid decomposition in Java, I mostly translated it to C++, so... if necessary, we have to change other file too. Also, there was code which implemented lists in it. Searching for strongly connected components in graph_algorithms. I will soon add test cases, so I believe this is the end of the line) Thank you for your reviews and support.
Btw, how can I pm you in case I have questions?

arnavb:
Just email me! (It's on my profile)

KtlTheBest:
@arnavb I finished writing Centroid Decomposition, with test case as well. Waiting for your approval.)

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp
```diff
@@ -0,0 +1,117 @@
...
+            more_than_1 = true;
+        }
+        std::cout << '\n';
+    }
+}
+
+int main(){
+    Graph tree(7);
+    Level.resize(8);
+    tree.addEdge(1, 2, tree.adj_);
```
arnavb: Since you are trying to pass member variables to the tree, I would just remove the parameter and directly modify these variables inside addEdge. Then you can make the variables private again.

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp
```diff
@@ -0,0 +1,117 @@
...
+    tree.addEdge(1, 3, tree.adj_);
+    tree.addEdge(2, 4, tree.adj_);
+    tree.addEdge(4, 5, tree.adj_);
+    tree.addEdge(4, 6, tree.adj_);
+    tree.addEdge(2, 7, tree.adj_);
+    tree.decomposeTree(1, -1, tree.ver_, 1, tree.adj_);
```
arnavb: Same here.

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp
```diff
@@ -0,0 +1,117 @@
...
+    showOutput();
```
arnavb: For `showOutput`, I would pass the Graph as an argument to it, and then iterate over it inside that function. Don't use the global variable `Level`, or keep track of it inside your graph class.

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp
KtlTheBest: @arnavb But there the moment where I add edges in another tree (Centroid Tree) using addEdge. If I modify it in the function, then how am I going to add edges to Centroid Tree?

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp
KtlTheBest: @arnavb I didn't quite understand what you mean by this. I add centroids in certain "levels" with help of decomposeTree. If I understood you correctly, then I have to decompose the graph again, second time to write centroids into Level[i].

arnavb:
@KtlTheBest You're right. But I would still figure out some way to get rid of the global variable.

KtlTheBest:
@arnavb I think I did it. Please review it.

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp
```diff
@@ -0,0 +1,117 @@
+#include <iostream>
+#include <vector>
+
+std::vector<std::vector<int>> Level;
```
arnavb: Is there any way you could move this variable to main?

KtlTheBest:
@arnavb Yes, I'm sorry for taking your time. Now Level is in main().

arnavb:
@KtlTheBest It's ok! You do have one error, reported by travis:

```C++
/home/travis/build/OpenGenus/cosmos/code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp
/home/travis/build/OpenGenus/cosmos/code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp: In member function ‘int Graph::findCentroid(int, int, int, std::vector<std::vector<int> >)’:
/home/travis/build/OpenGenus/cosmos/code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp:65:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
```

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp
```diff
@@ -0,0 +1,115 @@
...
+    {
+        if (!(to == par || marked_[to] == true))
+        {
+            if (sizes_[to] > vertices / 2)
+            {
+                return findCentroid(to, cur, vertices, graph);
+            }
+        }
+        return cur;
+    }
```
arnavb: Travis is warning that you could reach the end of this function. I would investigate and try to fix.

KtlTheBest:
@arnavb How do you think, is it okay to merge this pull request? It seems that Travis build is interrupted by other files.

code/graph_algorithms/src/centroid_decomposition/centroid_decompostition.cpp
KtlTheBest: @arnavb It was in wrong position. Fixed.

arnavb:
No, our Travis build is still in beta, so don't completely rely on its success. You just have to look at it to see where your file fails. @AdiChat will merge this!

KtlTheBest:
@arnavb It seems I can't merge the PR...)

arnavb:
@KtlTheBest Sometimes it takes a while. @AdiChat will merge when he gets to this; Have patience!

AdiChat:
Looks great 😍 

PR #3370, opened 2018-03-05
Added LGTM Support to Our Repo

arnavb:
**Fixes issue:** N/A yet...

**Changes:**
- Set up a maintainers file. This repository is not yet enabled, so @AdiChat has to do that from [https://lgtm.wu-boy.com/](https://lgtm.wu-boy.com/) (It's a fork) by making an account and enabling this repository from his profile.

UPDATE:

This works with my [test repository](https://github.com/arnavb/test-python-repository)! See this sample PR:

screen shot 2018-03-05 at 9 31 02 am

Unfortunately, the lack of documentation is still astounding.

UPDATE 2: Found some deprecated documentation. Going to apply it now. This link is: https://github.com/lgtmco/lgtm-docs/blob/master/content/customize.md


arnavb:
@AdiChat, there is only one way to find out if this works, and that is by merging this PR. If it doesn't work, then just revert it!

.lgtm
```diff
@@ -0,0 +1,2 @@
+pattern = "(?)LGTM"
```
arnavb: I have set up a basic pattern; This is just for a start. This will accept any string that ends with LGTM.

.lgtm
```diff
@@ -0,0 +1,2 @@
...
+self_approval_off = true
```
arnavb: Currently, each PR requires _two_ reviews from one of the people in the file below. This line prevents maintainers from approving themselves.

AdiChat:
Looks good :+1: 

I am not sure regarding the security implications of this as for a bot with write access, it requires access to private repositories and right to modify repositories of OpenGenus. Considering that the app has less than 10 users and lack of official support, it might not be safe to use. 
 
We need to think over this and other possibilities. :+1:

vaibhavsingh97:
IMO, it's not hard to make a custom bot for our org. I can setup one and can add regex to find "LGTM" and bot will add LGTM label. Also, we can add webhook to the slack to engage new contributors. Let me know your opinion 😀

arnavb:
@AdiChat Are you sure this app will be accessing the _private_ repositories of OpenGenus (It only works for the repositories you have enabled)? It is a very popular fork of the _official_ LGTM, so I don't think security should be an issue. They even have an official 

This is the official [LGTM Repository](https://github.com/go-gitea/lgtm) (The fork).
This popularity suggests that numerous users have tried using it, and have been able to succeed without any issues.
Even more to the benefit (Quoting README):

> If you find a security issue please contact security@gitea.io first.

So while this is your final decision, I suggest that you _do_  go forward with this, as it is only for cosmos, not the rest of your repos. Please let me know what you think, and I will close if you feel we should _not_ go forward with this.

PR #3413, opened 2018-03-16
Added singly linked list find operation in C++

Dhenderson10:
**Fixes issue:**

None.

**Changes:**

Added an algorithm to return true if an element is found in a linked list, false otherwise.

### DIFF
--- a/code/data_structures/src/list/singly_linked_list/operations/find/find.cpp
+++ b/code/data_structures/src/list/singly_linked_list/operations/find/find.cpp
@@ -58,22 +58,3 @@ bool find(LinkedList *list, int value)
     }
     return false;
 }
-
-int main()
-{
-    // Create linked list
-    LinkedList list; 
-    // Push in elements
-    push(&list, 1);
-    push(&list, 2);
-    push(&list, 3);
-    push(&list, 4);
-    push(&list, 5);
-    push(&list, 6);
-    // Find elements
-    std::cout << find(&list, 1) << "\n";  // Should be 1 - True
-    std::cout << find(&list, 3) << "\n";  // Should be 1 - True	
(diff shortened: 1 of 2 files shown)


code/data_structures/src/list/singly_linked_list/operations/find/find.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+// Part of Cosmos by OpenGenus Foundation
+struct Node
+{
+    Node *nextNode;
+    int data;
+
+    Node() //constructor
+    {
+        nextNode = NULL;
```
arnavb: Use `nullptr`.

code/data_structures/src/list/singly_linked_list/operations/find/find.cpp
```diff
@@ -0,0 +1,78 @@
...
+        data = 0;
+    }
+};
+
+// Singly Linked List
+struct LinkedList
```
arnavb: Why not use a class with some member functions?

code/data_structures/src/list/singly_linked_list/operations/find/find.cpp
```diff
@@ -0,0 +1,78 @@
...
+    }
+}
+
+// Return true if an element is found within a linked list
+// otherwise return false
+bool find(LinkedList *list, int value)
+{
+    Node *currentNode = new Node;
+    currentNode = list->head;
+    while (currentNode != NULL)
```
arnavb: while(currentNode) also works.

code/data_structures/src/list/singly_linked_list/operations/find/find.cpp
Dhenderson10: Just preference, doesn't matter to me.

PR #3423, opened 2018-03-21
invert bit in javascript

tbhaxor:
**Fixes issue:** NONE

**Changes:** invert bit in javascript added

### DIFF
--- a/code/bit_manipulation/src/invert_bit/invert_bit.js
+++ b/code/bit_manipulation/src/invert_bit/invert_bit.js
@@ -1,3 +1,3 @@
-const invertBit = (num, bit) => num ^= 1 << bit;
+const invertBit = (num, bit) => num ^ 1 << bit;
 
-console.log(invertBit(10, 5))  // 42
+console.log(invertBit(10, 5));  // 42
--- a/code/bit_manipulation/src/invert_bit/invert_bit.js
+++ b/code/bit_manipulation/src/invert_bit/invert_bit.js
@@ -1,0 +1,3 @@
+const invertBit = (num, bit) => num ^= 1 << bit;
+
+console.log(invertBit(10, 5))  // 42


code/bit_manipulation/src/invert_bit/invert_bit.js
```diff
@@ -0,0 +1,3 @@
+const invertBit = (num, bit) => num ^= 1 << bit;
```
arnavb: Why the assignment to `num`?

tbhaxor: idk but code is working :laughing:

arnavb: Does it work if you replace `^=` with `^`?

tbhaxor: no, wrong answer is coming ..

lets see what @AdiChat says

AdiChat: @tbhaxor This is related to JavaScript syntax. :+1: 

If you do not use the `return` keyword, then you need to assign the return value to some variable to achieve the same functionality. Therefore, if you want to remove `=` then, you must use `return` keyword.

arnavb: @AdiChat are you sure about that? As far as I know, [ES6 arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) without braces have an implicit return, so the explicit keyword isn't necessary. @tbhaxor I don't see an incorrect answer; When I remove the `=`, then I still get the same result. This is the code I am using (It's a diff):

```diff
-const invertBit = (num, bit) => num ^= 1 << bit;
+const invertBit = (num, bit) => num ^ 1 << bit;

console.log(invertBit(10, 5));  // 42
```

And here is a [JS Bin](http://jsbin.com/nofanizawo/edit?js,console) that illustrates a working example.

AdiChat: Yes, you are right. It looks like I was thinking of functions. :sweat_smile: 

I will reflect the change and merge this. :+1:

AdiChat:
Looks great 😍 

positive_opengenus_cosmos_1

arnavb:
Nice image choice! (Do you make an image every time, for the username?)

AdiChat:
Yes, I make a custom image for every user :smile:

arnavb:
😮

Issue #3431, opened 2018-03-22
add validate IPv6 in any language


tbhaxor:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

to know about IPv6 visit [this](https://en.wikipedia.org/wiki/IPv6) link

tbhaxor:
i will be coding in python3

kobra-cai:
I'll try to implement this in C if you don't mind

AdiChat:
@kobra-cai Sure, Go for it :+1:

skywalker212:
I would like to implement this in JavaScript

PR #3432, opened 2018-03-22
Create validate_ipv6.py

tbhaxor:
**Fixes issue:** #3431

**Changes:** added validate ipv6 in python


code/networking/src/validate_IP/validate_ipv6.py
```diff
@@ -0,0 +1,12 @@
+import socket
+
+def validateIPv6(ip):
+    try:
+        socket.inet_pton(socket.AF_INET6, ip)
+    except socket.error:
+        return False
+    return True
```
arnavb: I would move the `return True` inside the `try:` block.

AdiChat:
Looks great 😍 

Issue #3402, opened 2018-03-13
old codes are not supported in python 3


tbhaxor:
**This is a(n):**

- [ ] New algorithm
- [x] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

Python 2 will be obsoleted in ~2 years. Some of your use python 2 functions like
1. print
2. xrange
3. raw_input

Changes to be made
1. import print from future
```python
from __future__ import print_function
```
2. change xrange to range
```
try:
   xrange = range
except NameError:
   pass
```

3. change raw_input to range
```
try:
   input = raw_input
except NameError:
   pass
```

arnavb:
I think a better change might be to just convert the code? But yes, a good early change would be to add snippets such as this. However, this should be replaced with pure python3 code in the future.

tbhaxor:
may i take the responsibility, after march ?

arnavb:
@tbhaxor By all means, go for it! :+1:

tbhaxor:
i believe in updatig all code to py3

what do you think @AdiChat and @arnavb ?

mansiag:
I would like to work on this.

arnavb:
@mansiag @tbhaxor Sure, like I said, go for it! Python 2 [is getting old](https://pythonclock.org/)...

tbhaxor:
@mansiag lemme take this :D
I have commitrd changes just 2 3 directories are left

arnavb:
@tbhaxor You've gone through nearly every single file?

Wow... 👏 👏

tbhaxor:
@arnavb never doubt on my dedication :laughing:

tbhaxor:
@mansiag please allow me to open other Pull Request

mansiag:
@tbhaxor Can I work on this segment or should I delete this PR? I wont make any other changes. Thanks:smile:

tbhaxor:
@mansiag just do me favor please review all code in `code/data_structures` :smile:

mansiag:
@tbhaxor Thanks:smile:

PR #3442, opened 2018-03-26
Made changes to support Python 3

mansiag:
**Fixes issue:**
#3402 

**Changes:**
Made changes in right_view.py , maximum_height.py and minimum_height.py to support python 3.


mansiag:
@AdiChat Thanks a lot.:smile:

PR #3469, opened 2018-04-08
Added project euler problem7 solution in python

parulagg27:
**Fixes issue:**

**Changes:**

### DIFF
--- a/code/online_challenges/src/project_euler/problem_007/problem_007.py
+++ b/code/online_challenges/src/project_euler/problem_007/problem_007.py
@@ -1,15 +1,14 @@
 def main():
-	n = 10001
-	x = 2
-	list_of_primes = []
-	
-	while (len(list_of_primes) < n):
-		if all (x % prime for prime in list_of_primes):
-			list_of_primes.append(x)
-		x += 1
+    n = 10001
+    x = 2
+    list_of_primes = []
 
-	print (list_of_primes[-1])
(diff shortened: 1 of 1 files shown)


code/online_challenges/src/project_euler/problem_007/problem_007.py
```diff
@@ -0,0 +1,15 @@
...
+	list_of_primes = []
+	
+	while (len(list_of_primes) < n):
+		if all (x % prime for prime in list_of_primes):
+			list_of_primes.append(x)
+		x += 1
+
+	print (list_of_primes[-1])
+
+
```
mansiag: Remove this extra newline

parulagg27: I don't think this should be removed. According to PEP 8 style guide, we need to follow for python code in this repository, there should be 2 blank lines after class or function definition.

arnavb: @parulagg27 Looks like you're right (You also need this at the top of the file). However, I haven't really seen other files in Python follow this convention, so I would just remove the line.

parulagg27: Oh, alright.

AdiChat:
Looks great 😍 

positive_opengenus_cosmos

Issue #3473, opened 2018-04-09
Coding style of C++


iattempt:
**This is a(n):**

- [ ] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**
I'm fixing the coding style (Travis-CI errors) and I found some dangling style which need to be defined.

1.  Whether the 'namespace' body is indented.

arnavb:
@iattempt This is a very hard task, and it's good that someone is willing to do it! 😁

PR #3478, opened 2018-04-10
Update file to be more idiomatic C++

arnavb:
**Helps issue:** #3473 (Actually not really)

**Changes:**

- Get rid of `using namespace std`
- Some more style issues
- Indentation fix
- Braces to initialize classes
- Use `\n` instead of `endl` to remedy unnecessary stream flushing


AdiChat:
Looks great 😍 

positive_opengenus_arnavb

PR #3482, opened 2018-04-12
Fix coding style

iattempt:
**Fixes issue:**
#3473 

**Changes:**

1. Change some config of Uncrustify.
2. Fix coding style (add/remove spaces/indentation, or replace tabs with spaces, etc.).

Please write a comment for commit if you reviewed it, since this is big task.


third_party/uncrustify.cfg
```diff
@@ -702,14 +702,14 @@ indent_paren_open_brace         = false    # false/true
 indent_cs_delegate_brace        = false    # false/true
 
 # Whether the 'namespace' body is indented.
-indent_namespace                = false    # false/true
+indent_namespace                = true    # false/true
```
arnavb: I think this should NOT be true... Most large projects do NOT indent namespaces... (4ebad1)

third_party/uncrustify.cfg
```diff
@@ -719,7 +719,7 @@ indent_namespace_limit          = 0        # unsigned number
 indent_extern                   = false    # false/true
 
 # Whether the 'class' body is indented.
-indent_class                    = false    # false/true
+indent_class                    = true    # false/true
```
arnavb: This is also usually set to `false` in most C++ repos. Commit 92d0e124ae4f79330f222974a225eb1e08e54e63

iattempt: IMHO, methods should indent in class, either 2 spaces or 4 spaces because it should be more readable.
@OpenGenus/cosmos-maintainers What is your opinion on this?

arnavb:
@iattempt While you're fixing Travis CI and the likes, could you also fix the second C build? I believe the cause of the error is a typo: <code>libsdl1.2-dev<b>s</b></code> --> `libsdl1.2-dev`.

iattempt:
@AdiChat  @arnavb Please review

third_party/uncrustify.cfg
arnavb: @iattempt The class itself should be indented 4 spaces, **NOT** the access specifiers (`public`, `private`, `protected`).

iattempt: @arnavb is this what you mean for?
```
class Tree
{
    class Node
    {
public:  <----
        void value(int v)
        {
            // do something
        }
    }
}
```

arnavb: @iattempt This is what I mean:

```C++
class Tree
{
private: <---
    class Node
    {
    public:  <----
        void value(int v)
        {
            // do something
        }
    }
}
```

third_party/uncrustify.cfg
```diff
@@ -823,7 +823,7 @@ indent_col1_comment             = false    # false/true
 indent_label                    = 1        # number
 
 # Same as indent_label, but for access specifiers that are followed by a colon. Default=1
-indent_access_spec              = 1        # number
+indent_access_spec              = -4        # number
```
iattempt: @arnavb the access specifier indentation rule is specified here.

AdiChat:
Looks great 😍 

positive_opengenus

PR #3494, opened 2018-04-16
add implementations to validate ipv4 and ipv6 in C

kobra-cai:
**Fixes issue:**
#3431 
Added validate IPv6 and IPv4 (2 separate but really similar files) in C

**Changes:**
Implemented two separate but similar files called validate_ipv4.c and validate_ipv6 that validate both types respectively. They both, however, have very similar code with very small minor differences and I can try to do one of them in a completely different way if needed.


kobra-cai:
(FYI: CoralReefAtSea is my other github account (university related), made the commit of the implementations from the other account by accident.)

kobra-cai:
Sorry for the delay, made a new pull request: https://github.com/OpenGenus/cosmos/pull/3553

Issue #3411, opened 2018-03-15
Matrix Operations


nonotreally:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

I wish to add matrix operations under mathematical algorithms like
1)Addition of two matrices
2)Subtraction of two matrices
3)Product of two matrices
4)Checking equality of matrices

arnavb:
@nonotreally I would be careful with this as there are some implementations hovering around the repo. Otherwise, go for it!

tbhaxor:
@arnavb  can we use numpy ?

arnavb:
@tbhaxor I would try to avoid external dependencies, as the code may become non-portable otherwise. See what @AdiChat says.

tbhaxor:
@arnavb [this](https://github.com/tbhaxor/cosmos/blob/master/code/unclassified/src/magic_square/magic_square.py#L2) code is using numpy :smiley:

kobra-cai:
If you don't mind, I will do a pull request for the matrix multiplication and equality in C++

DeVasu:
I can code and implement all those Matrices Operations easily. but is someone else also working on this already?

nonotreally:
Yes I'm working on it...

uds5501:
@tbhaxor  Hey, just wanted to know that if you are still working with the python part or not. I could help with it if you may want to pass on the python implementation

nonotreally:
I'm only working on C and C++ part...

uds5501:
@nonotreally Great! I will help with python part

sakshikhachane:
I would like to work on this. Please assign me this.

PR #3498, opened 2018-04-17
Added matrix multiplication and equality implementations in c++ (#3411)

kobra-cai:
Fixes issue:
#3411

Changes:
Implemented matrix multiplication and equality in C++ using brute force and iterative approaches. The algorithms for them are in their respective functions void matrix_product() and bool matrix_equality() while the main is utilized as in interface with inputs being put in through cin.


arnavb:
@kobra-cai The location of the file is incorrect; Please move it to the correct folder!

kobra-cai:
Will do after Tuesday!

PR #3500, opened 2018-04-17
add_polynomials.cpp : wrote more C++ like and safety!

christianbender:
**Changes:**

* I wrote the project more C++ like. That means naming conventions for C++. And classes instead of structs.   
* I added a ```free()``` method for tidy up the heap. Because we use a linked list in the class Polynomial.   
* I wrote a destructor for tidy up the heap.


christianbender:
> However there's a lot of memory allocation happening here that should not ...

I have the project not written. I changed only the struct ```term``` into a class ```Term``` and added a destructor for cleaning up the heap.

OvermindDL1:
@chrisatanasian Yep, just noticing that the current file is... not even remotely what I'd find to be standard C++...  Nor is it efficient in the least...

Either way, Term would traditionally be a struct in C++ because it is POD, so I would not recommend making that change.

Instead of new/free an owned pointer should be used instead (that way things like the current implementations forgotten 'delete'/`free` could not actually be forgotten).

Though in general it would never be written the way that it is as it just starts scattering memory access all over, I'd easily have expected to be at least in a std::vector, not even with a term struct but just as a naked pair/tuple.  The code in this file as it is is easily 90% too large compared to the usual C++ implementation.  The fact it manually implements an intrusive linked list is amazingly inefficient, entirely not C++'ish, and is so difficult to read considering the actual simple implementation of the algorithm it is hiding inside of the mess....

I don't suppose you could clean all this up instead?

christianbender:
> Though in general it would never be written the way that it is as it just starts scattering memory access all over, I'd easily have expected to be at least in a std::vector, ...

You're right! I'd also use vector instead of own version of linked list.

> I don't suppose you could clean all this up instead?

Tomorrow I will see it. Then I used a vector

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -5,50 +5,56 @@
...
-  next = NULL;
-}
+class Term {
+  public:
+    Term(int c, int p) {
+      coeff = c;
+      pow = p;
+      next = NULL;
+    }
+    int coeff;
```
arnavb: If it's possible, could you make these variables private, have an underscore at the end, and have getters/setters if needed?

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -5,50 +5,56 @@
...
-  coeff = c;
-  pow = p;
-  next = NULL;
-}
+class Term {
+  public:
+    Term(int c, int p) {
+      coeff = c;
+      pow = p;
+      next = NULL;
```
arnavb: Please use `nullptr`.

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -5,50 +5,56 @@
...
-};
 
-term::term(int c, int p) {
-  coeff = c;
-  pow = p;
-  next = NULL;
-}
+class Term {
+  public:
+    Term(int c, int p) {
```
arnavb: I would prefer the use of a constructor initializer list for this constructor.

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -5,50 +5,56 @@
...
+      coeff = c;
+      pow = p;
+      next = NULL;
+    }
+    int coeff;
+    int pow;
+    Term* next;
```
arnavb: Can you consider using a smart pointer?

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -5,50 +5,56 @@
 #include <stddef.h>
 using namespace std;
```
arnavb: I would figure out a way to get rid of this line.

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -5,50 +5,56 @@
 #include <stddef.h>
```
arnavb: Replace with `cstddef`

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -5,50 +5,56 @@
...
-  term* head;
+class Polynomial {
+ private:
+  Term* head;
 
  public:
-  polynomial();
+  Polynomial();
+  ~Polynomial() {
+    free();
```
arnavb: Any reason why this is inline?

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -5,50 +5,56 @@
...
+  }
   void insert_term(int, int);
   void print();
-  friend polynomial operator+(polynomial, polynomial);
+  friend Polynomial operator+(Polynomial, Polynomial);
+  void free();
```
arnavb: Please mark implementation details as private.

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -5,50 +5,56 @@
...
+  void free(); 
 };
 
-polynomial::polynomial() { head = NULL; }
+// constructor 
+Polynomial::Polynomial() { head = NULL; }
 
-void polynomial::insert_term(int c, int p) {
+void Polynomial::insert_term(int c, int p) {
   if (head == NULL) {
```
arnavb: I would change all instances of `NULL` with `nullptr`.

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -57,8 +63,8 @@ void polynomial::insert_term(int c, int p) {
   }
 }
 
-void polynomial::print() {
-  term* t = head;
+void Polynomial::print() {
```
arnavb: This function is useless, TBH. Not every application uses a console, so it would be better if you returned the polynomial as a string. If you really want a print function, then overload the `<<` operator so it's not just restricted to stdout.

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -68,9 +74,9 @@ void polynomial::print() {
   cout << endl;
```
arnavb: Use newlines over `endl`.

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -68,9 +74,9 @@ void polynomial::print() {
...
 
-polynomial operator+(polynomial p1, polynomial p2) {
-  polynomial p;
-  term *t1 = p1.head, *t2 = p2.head;
+Polynomial operator+(Polynomial p1, Polynomial p2) {
+  Polynomial p;
+  Term *t1 = p1.head, *t2 = p2.head;
   while ((t1 != NULL) && (t2 != NULL)) {
     if (t1->pow > t2->pow) {
       p.insert_term(t1->coeff, t1->pow);
```
arnavb: C++ Style guide says to use camelCase for functions, BTW.

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -95,8 +101,24 @@ polynomial operator+(polynomial p1, polynomial p2) {
   return p;
 }
 
+void Polynomial::free() {
+  Term* t = head;
+  while (t) {
+    Term* tmp = NULL;
+    if (t->next) {
+      tmp = t->next;
```
arnavb: 4 spaces indentation.

christianbender:
@arnavb Thanks for assess. I will do the requested changes.

christianbender:
@arnavb Done.

AdiChat:
Nice work :+1: 

You must resolve the conflict and remove the binary file `a.out`. 

Kindly take a look into this and it will be good to be merged. :+1:

christianbender:
@AdiChat I removed the binary ```a.out```

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -2,101 +2,138 @@
...
-  next = NULL;
-}
+class Term {
+  private:
+    int coeff_;
+    int pow_;
+    Term* next_;
+  public:
+    Term(int c, int p): coeff_(c), pow_(p), next_(nullptr) {}
+    void set_coeff(int c) {coeff_ = c;}
```
arnavb: Please use camel case.

code/mathematical_algorithms/src/add_polynomials/add_polynomials.cpp
```diff
@@ -2,101 +2,138 @@
...
+  Polynomial();
+  ~Polynomial() {
+    free();
+  }
   void insert_term(int, int);
   void print();
-  friend polynomial operator+(polynomial, polynomial);
+  friend ostream& operator<< (ostream& stream, Polynomial poly);
+  friend Polynomial operator+(Polynomial, Polynomial);
+  void free();
```
arnavb: Can this function be private?

christianbender:
@arnavb Done.

AdiChat:
You need to resolve the conflict before we can merge this. :+1:

PR #3513, opened 2018-04-21
Affine.java : I fixed the bug of this program --- BUG FIX

christianbender:
**Fixes issue:**

* The program ```Affine.java``` ([here](https://github.com/OpenGenus/cosmos/blob/master/code/cryptography/src/affine_cipher/Affine.java) ) has a bug. The encryption process fails.  I fixed it. Now it run correct.  
```java
 private static void encrypt(int a, int b, String plaintext)
```
The method above has a bug.  

**Changes:**

* I changed the method above.


AdiChat:
Looks great 😍 

positive_opengenus_cosmos

PR #3519, opened 2018-04-22
tokenizer.cpp : more proper tokenizing.

christianbender:
**Changes:**

* Now the tokenizing process is proper. Multiple occurs of the delimiter will be overread.   
* Tidy up the heap. Set free the pointer ```ans``` with ```delete```

### DIFF
--- a/code/unclassified/src/tokenizer/tokenizer.cpp
+++ b/code/unclassified/src/tokenizer/tokenizer.cpp
@@ -2,8 +2,6 @@
 #include <cstring>
 // Part of Cosmos by OpenGenus Foundation
 
-using namespace std;
-
 char *myStrTok(char *input, char delim)
 {
 
@@ -46,14 +44,14 @@ int main()
 
     while (ans != nullptr)
     {
-        cout << ans << endl;
+        std::cout << ans << "\n";
         ans = myStrTok(nullptr, ' ');
     }
     char arr[] = "Hello world!";
     ans = myStrTok(arr, ' ');
-    cout << endl << ans << endl;
+    std::cout << std::endl << ans << "\n";
     ans = myStrTok(nullptr, ' ');
-    cout << endl << ans << endl;
+    std::cout << std::endl << ans << "\n";
 
     delete ans;
 
--- a/code/unclassified/src/tokenizer/tokenizer.cpp
+++ b/code/unclassified/src/tokenizer/tokenizer.cpp
@@ -2,6 +2,8 @@
 #include <cstring>
 // Part of Cosmos by OpenGenus Foundation
 
+using namespace std;
+
 char *myStrTok(char *input, char delim)
 {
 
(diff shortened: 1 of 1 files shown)


code/unclassified/src/tokenizer/tokenizer.cpp
```diff
@@ -2,6 +2,8 @@
 #include <cstring>
 // Part of Cosmos by OpenGenus Foundation
 
+using namespace std;
```
arnavb: No, please no. It is better practice to use `std::`.

code/unclassified/src/tokenizer/tokenizer.cpp
```diff
@@ -2,6 +2,8 @@
...
+
 char *myStrTok(char *input, char delim)
```
arnavb: If you plan on fixing this file, I would figure out a way to replace all usage of `char*` with `std::string`, since it's the best C++ practice.

code/unclassified/src/tokenizer/tokenizer.cpp
```diff
@@ -48,5 +55,7 @@ int main()
     ans = myStrTok(nullptr, ' ');
     cout << endl << ans << endl;
```
arnavb: Use `\n` over `endl`.

code/unclassified/src/tokenizer/tokenizer.cpp
```diff
@@ -48,5 +55,7 @@ int main()
...
 
+    delete ans;
```
arnavb: See, the usage of `std::string` would make this easier.

code/unclassified/src/tokenizer/tokenizer.cpp
christianbender: I must still use a pointer. Instead of ```char*``` I must use ```string*```. Because the function tidy up after its termination.

arnavb: @christianbender The point is, with `std::string`, you won't have to do tidying up at all! It might be a bit of a drastic change, but it is beneficial.

code/unclassified/src/tokenizer/tokenizer.cpp
arnavb: @christianbender I'm just saying that you really don't need pointers for tokenization. You might have to rewrite the algorithm, but if you don't want to, you can leave this as it is.

christianbender:
@arnavb I have done the requested changes except for using ```string```. I will do this.

arnavb:
@christianbender You don't have to make that change, but it is the better practice. In the end, it's up to you. 👍

code/unclassified/src/tokenizer/tokenizer.cpp
christianbender: @arnavb The problem is the function must return a string. But this must be a pointer of type ```std::string*``` .  I must use the heap because the function tidy up its stack.

christianbender:
@arnavb See #3525

PR #3525, opened 2018-04-23
Tokenizer string cpp

christianbender:
@arnavb as discussed

**Changes:**

* I used ```string``` instead of ```char*``` as return type.    
***important:***   I must use ```using namespace std;``` because my compiler don't recognize ```std::string```.  In both modes ```c++11``` and without ```c++11```.


AdiChat:
Looks great 😍 

positive_opengenus_cosmos

PR #3540, opened 2018-04-29
Added sphere-tetrahedron intersection.

RedBlight:
No brute-force here! Intersection volume and surface area are calculated using a complicated analytical method.

By the way, this is my first pull request ever on GitHub.

Codes are not formatted exactly as the guideline says, but me thinks tis perfectly readable.

### DIFF
--- a/code/computational_geometry/src/sphere_tetrahedron_intersection/LuVector.hpp
+++ b/code/computational_geometry/src/sphere_tetrahedron_intersection/LuVector.hpp
@@ -1,0 +1,884 @@
+/*
+N dimensional loop unrolled vector class.
+
+https://github.com/RedBlight/LoopUnrolledVector
+
+Refactored a little bit for "cosmos" repository.
+*/
+
+#ifndef LU_VECTOR_INCLUDED
+#define LU_VECTOR_INCLUDED
+
+#include <sstream>
+#include <string>
+#include <complex>
+#include <vector>
+#include <iostream>
+
+namespace LUV
+{
+    // LuVector.hpp
+
+    template<std::size_t N, class T>
+    class LuVector;
+
+    template<class T>
+    using LuVector2 = LuVector<2, T>;
+
+    template<class T>
+    using LuVector3 = LuVector<3, T>;
+
+    using LuVector2f = LuVector<2, float>;
+    using LuVector2d = LuVector<2, double>;
+    using LuVector2c = LuVector<2, std::complex<double>>;
+
+    using LuVector3f = LuVector<3, float>;
+    using LuVector3d = LuVector<3, double>;
+    using LuVector3c = LuVector<3, std::complex<double>>;
+	
+	// Added for cosmos
+	const double pi = 3.14159265358979323846;
+	//
+    
+    // LuVector_Op.hpp
+    
+    //////////////////////////////////////////////////////////////
+    // TEMPLATE LOOP INDEX VARIABLE
+    //////////////////////////////////////////////////////////////
+    template<std::size_t N>
+    struct LoopIndex
+    {
+
+    };
+
+    //////////////////////////////////////////////////////////////
+    // TEMPLATE LOOP UnrollING OPERATION FUNCTORS
+    //////////////////////////////////////////////////////////////
+
+    template<class T, class S>
+    struct OpCopy
+    {
+        inline void operator ()(T& lhs, const S& rhs)
+        {
+            lhs = rhs;
+        }
+    };
+
+    //////////////////////////////////////////////////////////////
+
+    template<class T, class S>
+    struct OpAdd
+    {
+        inline void operator ()(T& lhs, const S& rhs)
+        {
+            lhs += rhs;
+        }
+    };
+
+    template<class T, class S>
+    struct OpSub
+    {
+        inline void operator ()(T& lhs, const S& rhs)
+        {
+            lhs -= rhs;
+        }
+    };
+
+    template<class T, class S>
+    struct OpMul
+    {
+        inline void operator ()(T& lhs, const S& rhs)
+        {
+            lhs *= rhs;
+        }
+    };
+
+    template<class T, class S>
+    struct OpDiv
+    {
+        inline void operator ()(T& lhs, const S& rhs)
+        {
+            lhs /= rhs;
+        }
+    };
+
+    //////////////////////////////////////////////////////////////
+
+    template<class T, class S>
+    struct OpAbs
+    {
+        inline void operator ()(T& lhs, const S& rhs)
+        {
+            lhs = std::abs(rhs);
+        }
+    };
+
+    template<class T, class S>
+    struct OpArg
+    {
+        inline void operator ()(T& lhs, const S& rhs)
+        {
+            lhs = std::arg(rhs);
+        }
+    };
+
+    template<std::size_t N, class T, class S>
+    struct OpMin
+    {
+        inline void operator ()(LuVector<N, T>& result, const LuVector<N, T>& lhs, const LuVector<N, S>& rhs, const std::size_t& I)
+        {
+            result[I] = lhs[I] < rhs[I] ? lhs[I] : rhs[I];
+        }
+    };
+
+    template<std::size_t N, class T, class S>
+    struct OpMax
+    {
+        inline void operator ()(LuVector<N, T>& result, const LuVector<N, T>& lhs, const LuVector<N, S>& rhs, const std::size_t& I)
+        {
+            result[I] = lhs[I] > rhs[I] ? lhs[I] : rhs[I];
+        }
+    };
+
+    template<std::size_t N, class T, class S>
+    struct OpDot
+    {
+        inline void operator ()(T& result, const LuVector<N, T>& lhs, const LuVector<N, S>& rhs, const std::size_t& I)
+        {
+            result += lhs[I] * rhs[I];
+        }
+    };
+
+    //////////////////////////////////////////////////////////////
+
+    template<class T>
+    struct OpOstream
+    {
+        inline void operator ()(std::ostream& lhs, const T& rhs, const std::string& delimiter)
+        {
+            lhs << rhs << delimiter;
+        }
+    };
+    
+    // LuVector_Unroll.hpp
+    
+    //////////////////////////////////////////////////////////////
+    // TEMPLATE LOOP UnrollING
+    //////////////////////////////////////////////////////////////
+
+    // VECTOR OP SCALAR
+    template<std::size_t I, std::size_t N, class T, class S, template<class, class> class OP>
+    inline void Unroll(LuVector<N, T>& lhs, const S& rhs, OP<T, S> operation, LoopIndex<I>)
+    {
+        Unroll(lhs, rhs, operation, LoopIndex<I-1>());
+        operation(lhs[I], rhs);
+    }
+
+    template<std::size_t N, class T, class S, template<class, class> class OP>
+    inline void Unroll(LuVector<N, T>& lhs, const S& rhs, OP<T, S> operation, LoopIndex<0>)
+    {
+        operation(lhs[0], rhs);
+    }
+
+    // SCALAR OP VECTOR
+    template<std::size_t I, std::size_t N, class T, class S, template<class, class> class OP>
+    inline void Unroll(S& lhs, const LuVector<N, T>& rhs, OP<T, S> operation, LoopIndex<I>)
+    {
+        Unroll(lhs, rhs, operation, LoopIndex<I-1>());
+        operation(lhs, rhs[I]);
+    }
+
+    template<std::size_t N, class T, class S, template<class, class> class OP>
+    inline void Unroll(S& lhs, const LuVector<N, T>& rhs, OP<T, S> operation, LoopIndex<0>)
+    {
+        operation(lhs, rhs[0]);
+    }
+
+    // VECTOR OP VECTOR
+    template<std::size_t I, std::size_t N, class T, class S, template<class, class> class OP>
+    inline void Unroll(LuVector<N, T>& lhs, const LuVector<N, S>& rhs, OP<T, S> operation, LoopIndex<I>)
+    {
+        Unroll(lhs, rhs, operation, LoopIndex<I-1>());
+        operation(lhs[I], rhs[I]);
+    }
+
+    template<std::size_t N, class T, class S, template<class, class> class OP>
+    inline void Unroll(LuVector<N, T>& lhs, const LuVector<N, S>& rhs, OP<T, S> operation, LoopIndex<0>)
+    {
+        operation(lhs[0], rhs[0]);
+    }
+
+    // SCALAR = VECTOR OP VECTOR (could be cross-element)
+    template<std::size_t I, std::size_t N, class T, class S, template<std::size_t, class, class> class OP>
+    inline void Unroll(T& result, const LuVector<N, T>& lhs, const LuVector<N, S>& rhs, OP<N, T, S> operation, LoopIndex<I>)
+    {
+        Unroll(result, lhs, rhs, operation, LoopIndex<I-1>());
+        operation(result, lhs, rhs, I);
+    }
+
+    template<std::size_t N, class T, class S, template<std::size_t, class, class> class OP>
+    inline void Unroll(T& result, const LuVector<N, T>& lhs, const LuVector<N, S>& rhs, OP<N, T, S> operation, LoopIndex<0>)
+    {
+        operation(result, lhs, rhs, 0);
+    }
+
+    // VECTOR = VECTOR OP VECTOR (could be cross-element)
+    template<std::size_t I, std::size_t N, class T, class S, template<std::size_t, class, class> class OP>
+    inline void Unroll(LuVector<N, T>& result, const LuVector<N, T>& lhs, const LuVector<N, S>& rhs, OP<N, T, S> operation, LoopIndex<I>)
+    {
+        Unroll(result, lhs, rhs, operation, LoopIndex<I-1>());
+        operation(result, lhs, rhs, I);
+    }
+
+    template<std::size_t N, class T, class S, template<std::size_t, class, class> class OP>
+    inline void Unroll(LuVector<N, T>& result, const LuVector<N, T>& lhs, const LuVector<N, S>& rhs, OP<N, T, S> operation, LoopIndex<0>)
+    {
+        operation(result, lhs, rhs, 0);
+    }
+
+    // OSTREAM OP VECTOR
+    template<std::size_t I, std::size_t N, class T, template<class> class OP>
+    inline void Unroll(std::ostream& lhs, const LuVector<N, T>& rhs, const std::string& delimiter, OP<T> operation, LoopIndex<I>)
+    {
+        Unroll(lhs, rhs, delimiter, operation, LoopIndex<I-1>());
+        operation(lhs, rhs[I], delimiter);
+    }
+
+    template<std::size_t N, class T, template<class> class OP>
+    inline void Unroll(std::ostream& lhs, const LuVector<N, T>& rhs, const std::string& delimiter, OP<T> operation, LoopIndex<0>)
+    {
+        operation(lhs, rhs[0], delimiter);
+    }
+    
+    // LuVector_Body.hpp
+    
+    template<std::size_t N, class T>
+    class LuVector
+    {
+    public:
+        LuVector() = default;
+
+        ~LuVector() = default;
+
+        template<class... S> 
+        LuVector(const S&... elem) : _{ static_cast<T>(elem) ...}
+        {
+
+        }
+
+        LuVector(const LuVector<N, const T>& other) : _{ other._ }
+        {
+
+        }
+
+        template<class S> 
+        LuVector(const LuVector<N, S>& other)
+        {
+            Unroll(*this, other, OpCopy<T, S>(), LoopIndex<N-1>());
+        }
+
+        template<class S> 
+        LuVector(const S& scalar)
+        {
+            Unroll(*this, scalar, OpCopy<T, S>(), LoopIndex<N-1>());
+        }
+
+        template<class S>
+        inline LuVector<N, T>& operator =(const LuVector<N, S>& other)
+        {
+            Unroll(*this, other, OpCopy<T, S>(), LoopIndex<N-1>());
+            return *this;
+        }
+
+        inline const T& operator [](const std::size_t& idx) const
+        {
+            return _[idx];
+        }
+
+        inline T& operator [](const std::size_t& idx)
+        {
+            return _[idx];
+        }
+
+        inline std::string ToString() const
+        {
+            std::stringstream ss;
+            ss << "(";
+            Unroll(ss, *this, ",", OpOstream<T>(), LoopIndex<N-2>());
+            ss << _[N-1] << ")";
+            return std::string(ss.str());
+        }
+
+    private:
+        T _[N];
+    };
+    
+    // LuVector_Overload.hpp
+    
+    //////////////////////////////////////////////////////////////
+    // OPERATOR OVERLOADS
+    //////////////////////////////////////////////////////////////
+
+    // OSTREAM << VECTOR
+    template<std::size_t N, class T>
+    std::ostream& operator <<(std::ostream& lhs, const LuVector<N, T>& rhs)
+    {
+        lhs << "(";
+        Unroll(lhs, rhs, ",", OpOstream<T>(), LoopIndex<N-2>());
+        lhs << rhs[N-1] << ")";
+        return lhs;
+    }
+
+    // VECTOR += SCALAR
+    template<std::size_t N, class T, class S>
+    inline LuVector<N, T>& operator +=(LuVector<N, T>& lhs, const S& rhs)
+    {
+        Unroll(lhs, rhs, OpAdd<T, S>(), LoopIndex<N-1>());
+        return lhs;
+    }
+
+    // VECTOR -= SCALAR
+    template<std::size_t N, class T, class S>
+    inline LuVector<N, T>& operator -=(LuVector<N, T>& lhs, const S& rhs)
+    {
+        Unroll(lhs, rhs, OpSub<T, S>(), LoopIndex<N-1>());
+        return lhs;
+    }
+
+    // VECTOR *= SCALAR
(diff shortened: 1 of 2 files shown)


code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+#include <string>
+#include <complex>
+#include <vector>
+#include <iostream>
+
+namespace LUV
+{
+	// LuVector.hpp
+	
+	using namespace std;
```
arnavb: Please please please avoid this at ALL costs.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+			lhs = rhs;
+		}
+	};
+
+	//////////////////////////////////////////////////////////////
+
+	template< class T, class S >
+	struct OP_ADD
+	{
+		inline void operator ()( T& lhs, const S& rhs )
```
arnavb: You don't spaces before/after the parentheses. Same for the templates. You don't need spaces around them.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+	{
+
+	};
+
+	//////////////////////////////////////////////////////////////
+	// TEMPLATE LOOP UNROLLING OPERATION FUNCTORS
+	//////////////////////////////////////////////////////////////
+
+	template< class T, class S >
+	struct OP_COPY
```
arnavb: Also, don't use all caps for struct names. Use PascalCase and avoid underscores.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+	};
+
+	//////////////////////////////////////////////////////////////
+
+	template< class T, class S >
+	struct OP_ABS
+	{
+		inline void operator ()( T& lhs, const S& rhs )
+		{
+			lhs = abs( rhs );
```
arnavb: Please avoid `using namespace std` and fully qualify the name.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+			lhs = arg( rhs );
+		}
+	};
+
+	template< size_t N, class T, class S >
+	struct OP_MIN
+	{
+		inline void operator ()( LuVector< N, T >& result, const LuVector< N, T >& lhs, const LuVector< N, S >& rhs, const size_t& I )
+		{
+			result[ I ] = lhs[ I ] < rhs[ I ] ? lhs[ I ] : rhs[ I ];
```
arnavb: No spaces with the brackets.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+			result += lhs[ I ] * rhs[ I ];
+		}
+	};
+
+	//////////////////////////////////////////////////////////////
+
+	template< class T >
+	struct OP_OSTREAM
+	{
+		inline void operator ()( ostream& lhs, const T& rhs, const string& delimeter )
```
arnavb: Minor typo, but I think this isn't the correct spelling of delimiter.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+	// VECTOR OP VECTOR
+	template< size_t I, size_t N, class T, class S, template< class, class > class OP >
+	inline void UNROLL( LuVector< N, T >& lhs, const LuVector< N, S >& rhs, OP< T, S > operation, LOOP_INDEX< I > )
+	{
+		UNROLL( lhs, rhs, operation, LOOP_INDEX< I - 1 >() );
+		operation( lhs[ I ], rhs[ I ] );
+	}
+
+	template< size_t N, class T, class S, template< class, class > class OP >
+	inline void UNROLL( LuVector< N, T >& lhs, const LuVector< N, S >& rhs, OP< T, S > operation, LOOP_INDEX< 0 > )
```
arnavb: Please don't use ALL_CAPS; stick to camel case for function names.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+	}
+	
+	// LuVector_Body.hpp
+	
+	template< size_t N, class T >
+	class LuVector
+	{
+	private:
+		T _[ N ];
+
```
arnavb: Public interface goes before private; No spaces around brackets.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+	public:
+		LuVector( )
```
arnavb: How about:

```C++
LuVector() = default;
```

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+	{
+		LuVector< N, T > result( lhs );
+		return result -= rhs;
+	}
+
+	// VECTOR * VECTOR
+	template< size_t N, class T, class S >
+	inline LuVector< N, T > operator *( const LuVector< N, T >& lhs, const LuVector< N, S >& rhs )
+	{
+		LuVector< N, T > result( lhs );
```
arnavb: We use 4 spaces, not tabs.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: I thought using it in a namespace would limit it's scope and there wouldn't be any problems but a quick search proved otherwise. I'll not use it anymore.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: Sure, don't need them but... Tell me, isn't it more readable? Isn't it looking more sexy :)

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: Ah, they were related to experimental template stuff, I made them all caps to denote that they are the template unrolling related staff, but will fix it.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: Sure.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: What do you mean, could you elaborate lil bit more?

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: Ah, thanks for pointing out. No, not a typo, I knew it as 'delimeter' :) English is not my main language and I have never noticed that it was actually 'delimiter'.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: Ah, I got it, spaces inside square brackets... Well as I said it's much more readable, for me. Is this a strict rule that I must obey?

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: How about nothing at all? Isn't it default, by default?

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: Seriously? That's absolutely barbaric!
jk :)
Why though?

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: OK.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: "Public interface goes before private;"
Why?

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: By the way, what do you say about this: https://github.com/OpenGenus/cosmos/issues/3541

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
arnavb: @RedBlight We want consistency and according to our style guide, we only use spaces around operators, before opening braces, and right before opening parentheses for logical keywords (not functions).

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
arnavb: @RedBlight we have a style guide.....

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
arnavb: @RedBlight According to our style guide, the person looking at the code wants to most likely see the public interface first. They shouldn't care about implementation details.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
arnavb: @RedBlight No, since you have defined a constructor, you have to specify the `= default` or otherwise the constructor will not exist.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
arnavb: @RedBlight Since we don't have the option of a separate directory for just includes, an ugly option may have to be considered: Directly including .cpp files using relative paths. Yes, I know, I know, you're probably screaming at the mere thought of doing this, but there really isn't any more choice. Most of the algorithms in this repo were meant to be simple 1 file things, but now that we are expanding, we should make the appropriate changes. For now, there isn't much you can do.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+    {
+        V += testIntersection.GetSurfaceArea( R ) * S;
+    }
+    cout << V << ", " << testIntersection.GetVolume( U ) << endl; // must be almost equal
+    
+    return 0;
+}
+	
+	
+
```
arnavb: All these newlines?

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+    vol = testIntersection.GetVolume( 30 ); // must be 1000
+    cout << srf << ", " << vol << endl;
+    
+    // Numerical integral of surface area to obtain volume
+    int N = 10000;
+    double L = 0;
+    double U = 13;
+    double S = ( U - L ) / (double)N;
+    double V = 0;
+    for( double R = L; R < U; R += S )
```
arnavb: Space should be after for, not after (. Use lowercase for 1 letter variables.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+    
+    SphereTetrahedronIntersection testIntersection( vecTetA, vecTetB, vecTetC, vecTetD, vecSphCenter );
+    
+    // @ radius = 1 
+    double srf = testIntersection.GetSurfaceArea( 1 ) / pi; // must be 4
+    double vol = testIntersection.GetVolume( 1 ) / pi; // must be 4/3
+    cout << srf << ", " << vol << endl;
+    
+    // @ radius = 30
+    srf = testIntersection.GetSurfaceArea( 30 ); // must be 0
```
arnavb: Why did you reassign the variables? Couldn't you have just printed what the function returns? This way, you could get rid of the variables altogether and just print out the values above as well.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+    vol = testIntersection.GetVolume( 30 ); // must be 1000
+    cout << srf << ", " << vol << endl;
+    
+    // Numerical integral of surface area to obtain volume
+    int N = 10000;
+    double L = 0;
+    double U = 13;
+    double S = ( U - L ) / (double)N;
```
arnavb: Prefer a static_cast over a C style one.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+	template< class T >
+	inline LuVector< 3, T > PlaneNormalPN(  const LuVector< 3, T >& vec, const LuVector< 3, T >& v1, const LuVector< 3, T >& n ) 
+	{
+		return Dot( n, vec - v1 ) > 0 ? n : -n;
+	}
+
+};
+#endif
+
+const double pi = 3.14159265358979323846;
```
arnavb: I'd prefer that you namespace this variable in something called `Constants` or whatnot. (And maybe mark it constexpr). This may reduce the chance for a naming conflict.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+	
+	double GetSurfaceArea( double radius )
+	{
+		return GetSolidAngle( radius ) * radius * radius;
+	}
+	
+	double GetVolume( double radius )
+	{
+		double volume = 0;
+		for( int idx = 0; idx < 24; ++idx )
```
arnavb: Spaces go right after the keyword.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+				int idm = idp + 1;
+				
+				obsTetSgn.push_back(
+					LUV::Dot( obsTet[idp].prdA, dirN[idf] )
+					* LUV::Dot( obsTet[idp].prdF, dirL[ids] )
+					* LUV::Dot( obsTet[idp].prdD, dirU[ids] )
+				);
+				
+				obsTetSgn.push_back(
+					- LUV::Dot( obsTet[idm].prdA, dirN[idf] )
```
arnavb: This looks more confusing. If you mean the unary minus operator, then it should be attached to its number. The operations should be on the right hand side, not the left.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+		}
+		else if( radius > lenA )
+		{
+			return VolumeWedge( dirA, dirB, dirC, radius ) - (
+				VolumeCap( radius, LUV::Length( prpA - posA ) )
+				- VolumeCone( prlE, lenA )
+				) * angBAC / ( 2.0 * pi );
+		}
+		else
+		{
```
arnavb: Our style guide says that single line branches do not require braces.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+		if( radius > lenC )
+		{
+			return VolumeTetrahedron( posR, posA, posB, posC );
+		}
+		else if( radius > lenB )
+		{
+			return ( VolumeWedge( dirA, dirC, dirF, radius ) - (
+				VolumeCap( radius, LUV::Length( prpA - posA ) )
+				- VolumeCone( prlE, lenA )
+				) * angFAE / ( 2.0 * pi ) )
```
arnavb: I get why you would want to separate for readability, but this looks _really_ confusing. Could you rewrite this in a cleaner way?

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
```diff
@@ -0,0 +1,1237 @@
...
+		{
+			return SolidAngleWedge( dirA, dirC, dirF )
+				- SolidAngleCap( angERA ) * angFAE / ( 2.0 * pi );
+		}
+		else if( radius > lenA )
+		{
+			return solidAngleFull
+				- SolidAngleCap( angERA ) * angBAC / ( 2.0 * pi );
+		}
+		else
```
arnavb: No need for the else.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: "...Directly including .cpp files using relative paths..."
Actually, seperating the vector class into a seperate header file is not that bad.
I'll just add another file as "LuVector.hpp" containing the vector class, and then include it in "sphere_tetrahedron_intersection.cpp" if that's OK?

In that case, can we treat LuVector as a black box and leave it's formatting as it is? Since it will just be used as a header-only library in the main code file.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: Ah, of course, I remembered why I added them :)

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: It feels more comfortable to have a few newlines at the ready at the end of the files. Is it OK, or should I remove them before commit?

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: Will do.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: For no reason. It's just a simple test of the functionality, I haven't given much thought about it.

I might have unconsciously thought repeating this in a single line might make it uncomfortably long to read:
"testIntersection.GetSurfaceArea(...)"

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: Yes unary.

"The operations should be on the right hand side, not the left."

You mean this:
- A
* B
* C;

must be this:
-A*
B*
C;

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: It is itself confusing and complicated, see the math:
https://www.quora.com/What-are-some-equations-that-you-made-up/answer/Mahmut-Akku%C5%9F
but I'll try to write it more cleaner. I'm also open to suggestions.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: Again, this was just a quick demonstration of the functionality, and not something in the core functionality of the algorithm.
I think whatever that looks simpler and less verbose is better in this.

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
RedBlight: Cancer of the C++ :)
Even more cancerous thing is defining it like "pi = atan(1.0)*4;" :D
https://stackoverflow.com/questions/1727881/how-to-use-the-pi-constant-in-c
I mean, do they even...?

I'm planning to refactor the intersection codes, I'll put it in a better place.
By the way, constexpr is not necessary if it's not an expression like "atan(1.0)*4", does it?

RedBlight:
So, summing up all together for now:
 * Style guidelines has to be strictly obeyed, code must be reformatted according to it.
 * "using namespace" is a no-no.
 * 4 spaces instead of tabs.
 * Vector class codes will be seperated to another file.
 * Intersection codes will be refactored and written more clear.

Just one last question, do I have to reformat vector class codes, if it's going to be used as a lumped element in the main .cpp file? Since it would be used just like a 3rd party library in the main.cpp file. No one cares about what internal code style convention 3rd party libraries use, as long as their interface use a reasonable style.

By the way my next commit might come up to a month later. I guess it's OK right?

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
arnavb: Well you can't use it if it has a function call anyway, so I mean why not just write it out?

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
arnavb: No, I mean:

```
-A *
B *
C;
```

code/computational_geometry/src/sphere_tetrahedron_intersection/sphere_tetrahedron_intersection.cpp
arnavb: Yeah, even I personally agree with you, but still, it is what the style guide says.

arnavb:
@RedBlight 

For the 4 things, you're correct.

As for your question, if the file is in this repository, then it must follow the guide. Now if it was a git submodule or if it was dynamically linked, that would be different.

code/computational_geometry/src/sphere_tetrahedron_intersection/LuVector.hpp
```diff
@@ -0,0 +1,884 @@
...
+    using LuVector2f = LuVector<2, float>;
+    using LuVector2d = LuVector<2, double>;
+    using LuVector2c = LuVector<2, std::complex<double>>;
+
+    using LuVector3f = LuVector<3, float>;
+    using LuVector3d = LuVector<3, double>;
+    using LuVector3c = LuVector<3, std::complex<double>>;
+	
+	// Added for cosmos
+	const double pi = 3.14159265358979323846;
```
arnavb: Please fix the indentation

AdiChat:
Looks great 😍 

positive_opengenus_first

PR #3544, opened 2018-05-01
stack.cpp : some changes and improvements

christianbender:
**Changes:**

* Removed shell commands like ```cls``` and ```pause```. Because not platform-independent.  
* Added a destructor for tidy up the heap.  
* Renamed the ```struct node``` in ```struct Node``` according to the [code style conventions](https://github.com/OpenGenus/cosmos/tree/master/guides/coding_style/c%2B%2B)  
* Changed the program logic in the ```main```. Changed the ```while (true)```.  And removed the ```exit(0)```.

### DIFF
--- a/code/data_structures/src/stack/stack/stack.cpp
+++ b/code/data_structures/src/stack/stack/stack.cpp
@@ -6,17 +6,17 @@
 using namespace std;
 
 template <typename T>
-struct node
+struct Node
 {
     T n;
-    node *next;
+    Node *next;
 };
 
 template <class T>
 class Stack
 {
 private:
-    node<T> *top;
+    Node<T> *top;
 public:
     Stack();
     ~Stack();
@@ -34,8 +34,8 @@ template <class T> Stack<T>::~Stack()
 {
     if (top)
     {
-        node<T> * ptr = top;
-        node<T> * tmp = nullptr;
+        Node<T> * ptr = top;
+        Node<T> * tmp = nullptr;
         while (ptr)
         {
             tmp = ptr;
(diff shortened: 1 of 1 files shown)


code/data_structures/src/stack/stack/stack.cpp
```diff
@@ -6,19 +6,20 @@
...
-    node *next;
+    Node *next;
 };
 
 template <class T>
 class Stack
 {
 private:
-    node<T> *top;
+    Node<T> *top;
```
arnavb: Should be `Node<T>* top`.

code/data_structures/src/stack/stack/stack.cpp
```diff
@@ -29,9 +30,24 @@ template <class T> Stack<T>::Stack()
...
 
+template <class T> Stack<T>::~Stack()
+{
+    if (top)
+    {
+        Node<T> * ptr = top;
+        Node<T> * tmp = nullptr;
+        while (ptr)
+        {
+            tmp = ptr;
```
arnavb: You can make tmp a local variable.

code/data_structures/src/stack/stack/stack.cpp
christianbender: See my commit ```renamed struct node```

christianbender:
@arnavb Done. I made the variable ```tmp``` local.

code/data_structures/src/stack/stack/stack.cpp
arnavb: @christianbender I meant that the star should be next to the `<T>`,  not `top`. Same goes for `next` above.

christianbender:
@arnavb Done. I changed the formatting of the pointer type ```Node<T>*```.

AdiChat:
Looks great 😍 

positive_opengenus_cosmos

PR #3568, opened 2018-05-16
Internship screening, Addition of queue DS

crownedlake:
**Fixes issue:**

**Changes:**

### DIFF
--- a/code/data_structures/src/queue/queue.c
+++ b/code/data_structures/src/queue/queue.c
@@ -1,0 +1,94 @@
+#include<stdio.h>
+#include<stdlib.h>
+
+struct queue{
+    int rear, front,  size;
+    unsigned capacity;
+    int* array;
+};
+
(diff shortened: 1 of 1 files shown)


code/data_structures/src/queue/queue.c
```diff
@@ -0,0 +1,94 @@
+#include<stdio.h>
+#include<stdlib.h>
+
+struct queue{
```
arnavb: do a `typedef` so that the `struct` prefix can be avoided.

code/data_structures/src/queue/queue.c
```diff
@@ -0,0 +1,94 @@
...
+    int rear, front,  size;
+    unsigned capacity;
+    int* array;
+};
+
+struct queue* createQueue(unsigned capacity)
```
arnavb: Follow our C style guide.

PR #3574, opened 2018-05-18
string_addition.cpp : changed formatting and added assert(...)

christianbender:
**Changes:**

* I changed formatting and added ```assert(...)```  
  * Removed ```using namespace std;```  
  * Renamed a variable ```digit``` into ```sum```.


AdiChat:
Looks great 😍 

positive_opengenus_cosmos

PR #3586, opened 2018-05-28
Create runningkey_cipher.cpp

Technist:
Creating the C++ runningkey cipher file. _Plaintext_ and _key_ example was taken from runningkey_cipher.py.


arnavb (dismissed):
Finally! Something that doesn't have blatant issues!

Issue #3602, opened 2018-06-11
Addition of avl tree in python


yashshah1:
**This is a(n):**

- [ ] New algorithm
- [x ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

Support for avl tree in python

yashshah1:
Is this not needed?

divyeshgaur:
I would like to work on this.

thetaniyagupta:
I have made a PR for this issue. Have a look.

abhijeet010304:
I would Like to work on it in C++ and Python.. Please assign it to me...as a part of GSSoC20

PR #3603, opened 2018-06-12
Added AVL Tree in python #3602

yashshah1:
**Fixes issue:**

**Changes:**

Added the support for AVL Tree in python


code/data_structures/src/tree/binary_tree/avl_tree/avl_tree.py
```diff
@@ -0,0 +1,217 @@
...
+	my_tree.delete(3)
+	print "Delete 3"
+	my_tree.inorder_traverse()
+	print
+	print
+	print "Delete 4"
+	my_tree.delete(4)
+	my_tree.inorder_traverse()
+		
+
```
arnavb: Please remove all these extra newlines.

PR #3618, opened 2018-06-23
Create chans_algorithm.cpp

pankaj1181998:
**Fixes issue:**

**Changes:**

### DIFF
--- a/code/computational_geometry/src/chans_algorithm/chans_algorithm.cpp
+++ b/code/computational_geometry/src/chans_algorithm/chans_algorithm.cpp
@@ -1,265 +1,266 @@
-#include <iostream>
-#include <stdlib.h>
-#include <vector>
-#include <algorithm>			// For qsort() algorithm
-#include <utility>			    // For pair() STL
-#define RIGHT_TURN -1  			// CW
-#define LEFT_TURN 1  			// CCW
-#define COLLINEAR 0  			// Collinear
+	#include <iostream>
+	#include <stdlib.h>
+	#include <vector>
+	#include <algorithm>			// For qsort() algorithm
+	#include <utility>			    // For pair() STL
+	#define RIGHT_TURN -1  			// CW
+	#define LEFT_TURN 1  			// CCW
+	#define COLLINEAR 0  			// Collinear
 
-using namespace std ;
+	using namespace std ;
 
-/*
-Class to handle the 2D Points!
-*/
-class Point
-{
-    public:
-	
-	int x;
-	int y;
-	Point (int newx=0,int newy=0)
-	{
-		x=newx;
-		y=newy;
-	}
 	/*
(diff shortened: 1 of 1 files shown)


code/computational_geometry/src/chans_algorithm/chans_algorithm.cpp
```diff
@@ -0,0 +1,249 @@
+    #include <iostream>
+    #include <stdlib.h>
+    #include <vector>
+    #include <algorithm>	// For qsort() algorithm
+    #include <utility>		// For pair() STL
+    #define RIGHT_TURN -1  // CW
+    #define LEFT_TURN 1  // CCW
+    #define COLLINEAR 0  // Collinear
+    using namespace std;
```
arnavb: What's with the indentation here? Also, avoid `using namespace std` at all costs.

code/computational_geometry/src/chans_algorithm/chans_algorithm.cpp
```diff
@@ -0,0 +1,249 @@
...
+    /*
+    Class to handle the 2D points!
+    */
+    class point
```
arnavb: Capital `P` please.

code/computational_geometry/src/chans_algorithm/chans_algorithm.cpp
```diff
@@ -0,0 +1,249 @@
...
+        int y;
+        point (int newx=0,int newy=0){
+            x=newx;
+            y=newy;
+        }
+        /*
+            Overloaded == operator to check for equality between 2 objects of class point
+        */
+        friend bool operator== (const point& p1,const point& p2){
+            return (p1.x==p2.x && p1.y==p2.y);
```
arnavb: No need for parentheses.

code/computational_geometry/src/chans_algorithm/chans_algorithm.cpp
```diff
@@ -0,0 +1,249 @@
...
+        }
+        /*
+            Overloaded != operator to check for non-equality between 2 objects of class point
+        */
+        friend bool operator!= (const point& p1,const point& p2){
+            return (!(p1.x==p2.x && p1.y==p2.y));
```
arnavb: How about `return !(p1 == p2);`.

code/computational_geometry/src/chans_algorithm/chans_algorithm.cpp
pankaj1181998: Tried removing curly braces but there it's giving error (precedence of operators).

code/computational_geometry/src/chans_algorithm/chans_algorithm.cpp
pankaj1181998: the operator "=="  is overloaded here in this function.  So that they can compare them using "==" operator in the further code.

Issue #3623, opened 2018-06-29
Code implementation in C++ for balanced_expression in stack data structure


nitishbahl:
**This is a(n):**

- [ ] New algorithm
- [x] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

Currently the implementation of [balanced_expression](https://github.com/OpenGenus/cosmos/tree/master/code/data_structures/src/stack/balanced_expression) problem under [Stack](https://github.com/OpenGenus/cosmos/tree/master/code/data_structures/src/stack) data structure is available only in Java, this should be implemented in other programming languages such as C,C++,Python,Scala etc.
I wish to add code implementation for it in C++, support for other languages is welcomed.
@AdiChat @iattempt

saurabhgoyaliitr:
Hi, I want to contribute the C++ implementation.

deesims:
Can I do python implementation ?

abhijeet010304:
I wish to contribute to it as a part of GSSoC 20 in C++. Please assign this to me .

PR #3624, opened 2018-06-29
Fixes issue #3263-Code implementation in C++ for balanced_expression in stack data structure

bahl24:
**Fixes issue:**

Fixes issue [#3263](https://github.com/OpenGenus/cosmos/issues/3623)-Code implementation in C++ for balanced_expression in stack data structure.

**Changes:**

A C++ implementation for [balanced_expression](https://github.com/OpenGenus/cosmos/tree/master/code/data_structures/src/stack/balanced_expression) was added. This expands the language base for the [Stack data structure](https://github.com/OpenGenus/cosmos/tree/master/code/data_structures/src/stack) for which currently only Java implementation is available.


nitishbahl:
The code takes a string of bracket as input from stdin and prints YES if the string is balanced and NO if string is not balanced. 
Input format-
1. First line takes in number of test cases as t.
2. Next t lines takes in the strings to be tested.
As I am a newbie, please give a review. @AdiChat

code/data_structures/src/stack/balanced_expression/balanced_expression.cpp
```diff
@@ -0,0 +1,70 @@
+#include <iostream>
+#include <vector>
+#include <cstring>
+#include <stack>
+
+using namespace std;
```
arnavb: Please avoid this.

code/data_structures/src/stack/balanced_expression/balanced_expression.cpp
```diff
@@ -0,0 +1,70 @@
+#include <iostream>
```
arnavb: Alphabetical order.

code/data_structures/src/stack/balanced_expression/balanced_expression.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <vector>
+#include <cstring>
+#include <stack>
+
+using namespace std;
+
+bool isOpen(char c)
+{
+    if(c == '(' || c == '{' || c == '[')
```
arnavb: How about: `return c == '(' || c == '{' || c == '[';`? You could also consider marking this function `inline`.

code/data_structures/src/stack/balanced_expression/balanced_expression.cpp
```diff
@@ -0,0 +1,70 @@
...
+        return true;
+    return false;
+}
+
+bool isBalanced(char popped, char now)
+{
+    if(popped == '(' && now ==')')
```
arnavb:
```C++
return (popped == '(' && now ==')') ||
    (popped == '[' && now ==']') ||
    (popped == '{' && now =='}');
```

code/data_structures/src/stack/balanced_expression/balanced_expression.cpp
```diff
@@ -0,0 +1,70 @@
...
+    if(popped == '{' && now =='}')
+        return true;
+    return false;
+}
+
+int main()
+{
+    /* The code takes a string str as input from stdin and prints "YES" 
+    if str is a balanced expression else it prints "NO".*/
+        stack<char> s;
```
arnavb: Fix your indentation here please.

code/data_structures/src/stack/balanced_expression/balanced_expression.cpp
```diff
@@ -0,0 +1,70 @@
...
+        string str;
+        cin>>str;
```
arnavb: Add spaces around operators please.

code/data_structures/src/stack/balanced_expression/balanced_expression.cpp
```diff
@@ -0,0 +1,70 @@
...
+        int no = 0;
+        for(int i = 0; i < str.length(); i++)
```
arnavb: I think a range based for loop might be clearer, unless `i` is used somewhere in the loop.

code/data_structures/src/stack/balanced_expression/balanced_expression.cpp
```diff
@@ -0,0 +1,70 @@
...
+                if(isBalanced(popped, now))
+                    continue;
+                else
+                {
+                    no = 1;
+                    break;
+                }
+            }
+        }
+        if(s.size() != 0)
```
arnavb: ```if (!s.empty())```

code/data_structures/src/stack/balanced_expression/balanced_expression.cpp
bahl24: As `i` is being used inside the loop, a range based loop will lead in more execution time or might be insufficient if `str` is quite long. Plus using `str.length()` makes viewer exactly understand the code.

bahl24:
@arnavb I have reviewed the changes suggested by you and have made the changes wherever applicable. I have also justified my reasons few changes suggested.

code/data_structures/src/stack/balanced_expression/balanced_expression.cpp
bahl24: As the return type is `bool` so we need to return `true` or `false` and not `1` or `0`. I have checked that the code fails in that case.

code/data_structures/src/stack/balanced_expression/balanced_expression.cpp
bahl24: As the return type is `bool` so we need to return `true` or `false` and not `1` or `0`. I have checked that the code fails in that case.

arnavb: @bahl24 Sorry what? Last I checked, the or operator returns a `bool` value. `1` and `0` evaluate to `true` and `false` respectively anyway. This shouldn't be a problem, unless the return type was that of an integer (it isn't). Could you clarify what test cases this fails in?

bahl24: Sample input- `(})`
It fails to output "NO" when we use `return (popped == '(' && now ==')') ||
    (popped == '[' && now ==']') ||
    (popped == '{' && now =='}');`
But, it shouldn't be a problem as the latest code passes any sample input given and is more easy to understand.

arnavb: Actually, it does output `NO`. Check this and modify as you desire: https://wandbox.org/permlink/ygACUgdFsVMLdG8Q

As for your second point, I personally disagree, since it is repetitive, but it is a matter of opinion so I'm not going to contradict you there. However, the code I gave _definitely_ works, as both snippets are logically equivalent, as shown in the wandbox above (I used the sample input you gave).

bahl24:
@arnavb  @AdiChat I have incorporated all the changes suggested by @arnavb . Sorry for the trouble, but now the code is working correctly.

code/data_structures/src/stack/balanced_expression/balanced_expression.cpp
```diff
@@ -0,0 +1,62 @@
+#include <iostream>
+#include <stack>
+
+using namespace std;
```
arnavb: Just remove usage of `using namespace std`, and you're good to go!

bahl24: @arnavb @AdiChat Done.

bahl24:
@arnavb @AdiChat any problem with the code?

PR #3630, opened 2018-07-04
Add files via upload

brunaostii:
I add a new type of sorting algorith called timsort implemented in c.

### DIFF
--- a/code/sorting/src/tim_sort/Timsort.c
+++ b/code/sorting/src/tim_sort/Timsort.c
@@ -1,0 +1,101 @@
+#include<stdio.h>
+#include<limits.h>
+#include<stdlib.h>
+
+
+const int RUN = 32;
+
+
+/*InsertionSort*/
+void 
+insertionSort(int *A , int left, int right) 
+{
+
+    int i, j;
+    for (int i = left + 1; i <= right; i++) {
+        int temp = A[i];
+        int j = i - 1;
+        while (A[j] > temp && j >= left) {
+            A[j+1] = A[j];
+            j--;
+        }
+        A[j+1] = temp;
+    }
+}
+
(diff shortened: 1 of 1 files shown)


code/sorting/src/tim_sort/Timsort.c
```diff
@@ -0,0 +1,101 @@
...
+
+/*InsertionSort*/
+void 
+insertionSort(int *A , int left, int right) 
+{
+
+    int i, j;
+    for (int i = left + 1; i <= right; i++) {
+        int temp = A[i];
+        int j = i - 1;
```
christianbender: Double declaration of ```j```. See line 14.

code/sorting/src/tim_sort/Timsort.c
```diff
@@ -0,0 +1,101 @@
...
+            A[k] = L2[j];
+            j++;
+        }
+    }
+
+    free(L1);
+    free(L2);
+}
+
+int
```
christianbender: Instead of this function you could write it as makro:
```c
#define min(a, b) (((a) < (b)) ? (a) : (b)) 
```

code/sorting/src/tim_sort/Timsort.c
```diff
@@ -0,0 +1,101 @@
...
+int 
+min(int a, int b) 
+{
+    if(a < b) return (a);
+    else return (b);
+}
+
+/*TimSort*/
+void 
+timSort(int *A, int tamanho)
```
christianbender: Instead of ```tamanho``` write ```len``` or ```size```. Because we are international :blush:

brunaostii: thank you

Issue #3634, opened 2018-07-09
Adding C++ for Prefix to Postfix


CodeSeeker99:
**This is a(n):**

- [ ] New algorithm
- [x] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

Currently, the prefix to postfix algorithm is available in python only, I'd like to add C++ support.

amit-s19:
assign me this issue pls.

rucha174:
Is this issue still open?

anmolkaur18:
Assign this issue to me please @CodeSeeker99

rudrakshi99:
I want to work on this issue as a part of GSSoC'20 .Plzz assigns it to me.

rudrakshi99:
@arnavb Sir, please review my PR.

PR #3635, opened 2018-07-09
Adding Prefix to Postfix for C++

CodeSeeker99:
**Fixes issue:**

#3634 

**Changes:**

Added .cpp file for prefix to postfix algorithm.
Tested speed is about as much as the infix to postfix algorithm.
This is my first commit, feedback is always **appreciated**.


CodeSeeker99:
image

I don't understand. Where is the issue?

PR #3637, opened 2018-07-11
Create Huffman.java

codebloddedharshit:
Java code for Huffman Coding Algorithm

**Fixes issue:**

**Changes:**


code/greedy_algorithms/src/huffman_coding/Huffman.java
```diff
@@ -0,0 +1,134 @@
+import java.util.PriorityQueue;
+import java.util.Scanner;
+import java.util.Comparator;
+
```
christianbender: You must split your classes into seperate files. Or write the classes as inner classes of the class ```Huffman```. 
Otherwise you get an error of your compiler.

christianbender: @codebloddedharshit I have tested your project. It works just as it

PR #3638, opened 2018-07-16
shuffle_an_array.py : removed argument and write more randomly

christianbender:
**Fixes issue:**

* No seed. Permutations often the same by each program start.  

**Changes:**

* Removed useless argument ```n``` use ```len()``` instead.  
* Seed the random generator by each call of the function.


AdiChat:
Looks great 😍 
Do join us at Discourse and send [me](https://discourse.opengenus.org/u/aditya/summary) a message so we can discuss regarding a few things. 

positive_opengenus

PR #3707, opened 2018-10-02
Added Naive bayes example using Iris dataset (k-fold cross validation…

danielmg97:
Just added an example of Naive Bayes using the Iris dataset which is also included. As an extra did k-fold cross validation which might come in handy. The example is made on python 3 using sklearn.


AdiChat:
Looks great 😍 

You have made a wonderful contribution. Join us [**at our community forum**](https://discourse.opengenus.org/signup) :family_man_woman_girl_boy: to get connected and access to community benefits

positive_opengenus

Issue #3713, opened 2018-10-02
Add code for digit dp in other languages


abhaygupta97:
This is a:
- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

Digit dp code is currently present only in c++. Add code for digit dp in other languages such as ```Java```, ```Python```, ```Swift```,  ```JavaScript```, ```GoLang```, ```Matlab``` and others.

nghiaxlee:
I would like to add Java implementation.

shamamaz:
I would like to add the Python implementation.

Meenakshi2000:
i would too lie to work on it in java

pragyakapoor:
Hey, can I work on this in Python?

jonathlan:
Hello !
First timer in hackoberfest here.
I will try this in JavaScript.

jonathlan:
Hi there!
Just wondering if there is anybody available to take a look at my PR https://github.com/OpenGenus/cosmos/pull/6569? 
Would love to know your comments.

Thanks!

priyakant007:
Hi, 
I have added a Python solution for Digit DP. Could you please review it and let me know if there is anything I should change or improve? If everything looks good, please consider merging it. Thank you!

PR #3714, opened 2018-10-02
Adding code for digit-dp using python

abhaygupta97:
**Fixes issue:**
[#3713](https://github.com/OpenGenus/cosmos/issues/3713#issue-365903711)

**Changes:**
Adding python solution for the digit-dp dynamic programming problem.


sibasish14:
@abhaygupta97 make it PEP8 compliant.

abhaygupta97:
@sibasish14 made the changes

sibasish14:
@abhaygupta97  It's still not PEP8 compliant. Check your code here- http://pep8online.com/
Then make changes accordingly.

sibasish14:
.Also run your code once after making the changes. Ensure that it works.

abhaygupta97:
@sibasish14 thanks for the feedback. I made the suggested changes. The code is working now and it was working earlier also. All PEP8 issues are due to use of tabs which I used as it was given in [python style guide](https://github.com/OpenGenus/cosmos/tree/master/guides/coding_style/python).

abhaygupta97:
@AdiChat please review this pull request.

code/dynamic_programming/src/digit_dp/digit_dp.py
```diff
@@ -0,0 +1,95 @@
...
+
+	def digit_dp_helper(num, pos, count, f):
+		"""A recursive helper function for digit_dp function."""
+		if count > k:
+			return 0
+
+		if pos == len(num):
+			if count == k:
+				return 1
+			return 0
```
Vitorvgc: This can be simplified to `return count == k`

code/dynamic_programming/src/digit_dp/digit_dp.py
```diff
@@ -0,0 +1,95 @@
...
+	"""Check if n has digit d in it k times."""
+	count = 0
+	for c in str(n):
+		if int(c) == d:
+			count += 1
+
+	if count == k:
+		return 1
+	else:
+		return 0
```
Vitorvgc: This can be simplified to `return count == k`

code/dynamic_programming/src/digit_dp/digit_dp.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def main():
+	"""Main function."""
+	a = 100
+	b = 100000000
+	d = 3
+	k = 5
+	ans = digit_dp(b, d, k) - digit_dp(a, d, k) + check_num(a, d, k)
```
Vitorvgc: I think that to check a range `[a, b]`, we can simply compute `digit_dp(b, d, k) - digit_dp(a-1, d, k)`. This way we could get rid of the `check_num` function and make the code simpler.

sibasish14: @Vitorvgc are you sure we need to do it this way? I wrote this code in C++, and the reason why I included the check() was because `a` and `b` are strings. And we cannot subtract an integer from a string in C++. Hence, had to add a check() to make sure I'm not missing out `a`.

But I don't know how it works in python. So please explain.

Vitorvgc: I see, it is not necessary, I just think this way makes it simpler. Python also helps here. This change is easy to do here because the main function `digit_dp` already receives an integer, and inside of it there is a helper function `digit_dp_helper`, and only this helper function works with the number as a string. In the code we can see that in line 68 it is made a cast from int to string. So this way when calling the main function `digit_dp` we can work with integers, allowing to do `a-1`, and this function will call the recursive one that actually does the job, passing the number as string.

In your C++ code, you can achieve this by simply making the function receive an integer, and what is does is just cast it to string and call the DP function that works with the string. For example, you can change your `solve` function to receive an integer and then pass the string to the `digit_dp` function like this:

```c++
long solve(long num)
{
    memset(dp, -1, sizeof(dp));
    long result = digit_dp(to_string(num), 0, 0, 0);
    return result;
}
```

This way you could do:

```c++
int main() {
    long a = 100, b = 100000000;
    long ans = solve(b) - solve(a-1);
    cout << ans << endl;
}
```

And this will achieve the same result.

sibasish14: I have a reason for using string. What if the number was a 2000 digit number?? It won't fit in any integer type. Hence used string.

Vitorvgc: Yes, the drawback of using integer in this case is overflow problems. So if it needs to support arbitrary precision integers I think there is no way to avoid an extra function, either your check function or a "BigInteger" structure. Python makes things easier because it already supports arbitrary precision integers.

sibasish14: Agree! @abhaygupta97 make the required changes.

abhaygupta97:
@Vitorvgc thanks for the review. I have made all the suggested changes.

abhaygupta97:
@AdiChat kindly merge the pull request. Or do I need to make some other changes?

PR #3723, opened 2018-10-03
Update dijkstra.js

timhuff:
`Array(noOfVertices).fill(Array(noOfVertices).fill(0))` does not create an nxn matrix of zeroes. It creates an array of n references to the same array of n zeroes (that is to say, only 2 arrays are ever being created, rather than n+1). I also put a default value on the `Graph#dijkstra` method so it can be invoked without an argument.

**Fixes issue:**

**Changes:**

### DIFF
--- a/code/graph_algorithms/src/dijkstra_shortest_path/dijkstra.js
+++ b/code/graph_algorithms/src/dijkstra_shortest_path/dijkstra.js
@@ -6,9 +6,9 @@
  */
 
 class Graph {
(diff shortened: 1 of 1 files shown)


AdiChat:
Looks great 😍 

You have made a wonderful contribution. Join us [**at our community forum**](https://discourse.opengenus.org/signup) :family_man_woman_girl_boy: to get connected and access to community benefits

positive_opengenus

PR #3754, opened 2018-10-06
Added: collatz_conjecture_sequence.c

YashovardhanSiramdas:
**Changes:**
Added a C program to find collatz conjecture sequence which starts with given input "n"


arnavb:
djldk 2

PR #3775, opened 2018-10-08
Add solution to Project Euler problem 023 in C++

Vitorvgc:
**Fixes issue:**
None

**Changes:**
Add description and C++ solution to the [Non-abundant sums](https://projecteuler.net/problem=23) problem of Project Euler


code/online_challenges/src/project_euler/problem_023/problem_023.cpp
```diff
@@ -0,0 +1,43 @@
+#include <iostream>
+#include <vector>
+#include <algorithm>
```
arnavb: Alphabetical order.

PR #3791, opened 2018-10-13
Rename linearRegression.js to linear_regression.js

sailakshmi17:
**Changes:**

Changed the filename as per naming convention


arnavb:
djldk 2

PR #3794, opened 2018-10-13
Sum the numbers in an array - elixir

MariliaMJ:
**Changes:**
A file containing a simple sum of numbers in an array using elixir.


MariliaMJ:
This is not building, but I did not understand why. If anyone could help me with it, much appreciated.

PR #3800, opened 2018-10-14
fix cheking on password strength

MichelHanna:
**Fixes issue:**

checking on password strength is incorrect.
**Changes:**

alter checking on password strength to return true if password is strong and false if password is weak.


MichelHanna:
@AdiChat you may forgot to merge this pull request.

arnavb:
djldk 2

PR #3805, opened 2018-10-14
Add Digit_dp in python

shamamaz:
**Fixes issue:**
#3713 

**Changes:**
Adds code for digit dp in python.

### DIFF
--- a/code/dynamic_programming/src/digit_dp/digit_dp.py
+++ b/code/dynamic_programming/src/digit_dp/digit_dp.py
@@ -1,0 +1,75 @@
+# Part of Cosmos by OpenGenus Foundation
+# By: Shamama Zehra (https://github.com/shamamaz)
+
+# How many numbers x are there in the range a to b
+# where the digit d occurs exactly k times in x?
+# dp[pos][count][f] = Number of valid numbers <= b from this state
+# pos = current position from left side (zero based)
+# count = number of times we have placed the digit d so far
+# f = the prefix of the number we are building is smaller than prefix in b [0 = no, 1 = yes]
+import numpy as np
+
+def digit_dp(num, pos, count, f):
+    global dp
+    global k
+    global d
+
+    if (count > k):
+        return 0
+
+    if (pos == len(str(num))):
+        if (count == k):
+            return 1
+        return 0
+
+    if (dp[pos][count][f] != -1):
+        return dp[pos][count][f]
+
+    result = 0
+
+    if (f == 0):
+        string_num = str(num)
+        limit = int(string_num[pos])
+    else:
+        limit = 9
+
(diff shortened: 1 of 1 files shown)


code/dynamic_programming/src/digit_dp/digit_dp.py
```diff
@@ -0,0 +1,75 @@
...
+# How many numbers x are there in the range a to b
+# where the digit d occurs exactly k times in x?
+# dp[pos][count][f] = Number of valid numbers <= b from this state
+# pos = current position from left side (zero based)
+# count = number of times we have placed the digit d so far
+# f = the prefix of the number we are building is smaller than prefix in b [0 = no, 1 = yes]
+import numpy as np
+
+def digit_dp(num, pos, count, f):
+    global dp
```
arnavb: Do you _really_ need globals? Why not just pass values around as parameters?

code/dynamic_programming/src/digit_dp/digit_dp.py
```diff
@@ -0,0 +1,75 @@
...
+    numb = str(numb)
+    for i in range (0, len(numb)):
+        if (int(numb[i]) == d):
+            count += 1
+
+    if (count == k):
+        return 1
+    return 0
+
+if __name__ == "__main__":
```
arnavb: Please create a separate main function and call it here. Also, PLEASE get rid of the global variables.

code/dynamic_programming/src/digit_dp/digit_dp.py
```diff
@@ -0,0 +1,75 @@
...
+
+def check(numb):
+    count = 0
+    numb = str(numb)
+    for i in range (0, len(numb)):
+        if (int(numb[i]) == d):
+            count += 1
+
+    if (count == k):
+        return 1
```
arnavb: Are you trying to represent a boolean? Then why not do `return count == k`?

code/dynamic_programming/src/digit_dp/digit_dp.py
shamamaz: With this one, because this function is called recursively and edits the same array, using globals is a fairly direct approach. The alternative would be to pass in the dp array to the function each time which seems memory intensive

code/dynamic_programming/src/digit_dp/digit_dp.py
shamamaz: No need for caps! Thanks for the suggestion, it makes sense with k and d.

sibasish14:
@shamamaz a PR (https://github.com/OpenGenus/cosmos/pull/3714) already exists for this one. You'll have to wait and see if the other one is being merged. Please try other issues till then :)

PR #3826, opened 2018-10-19
made changes

archanacoder:


arnavb:
@archanacoder Please do not get rid of the working code already present for bubble sort. In addition, this code is directly copy/pasted from https://www.geeksforgeeks.org/bubble-sort/. Plagiarism will not be tolerated in our community.

In the end, please make sure to see if implementations are already present for an algorithm. If so, then consider refactoring it for improvement! We value your contributions, but we want to make sure it really is _your_ contribution in the first place.

PR #3831, opened 2018-10-20
fixed NullPointerException and fixed some formatting issues

A-Au:
- Fixed a NullPointerException that can be thrown when deleting a node
- Fixed the case where you delete the first item in the list
- Renamed file to match Java coding standards
- Minor formatting changes


arnavb:
djldk 2

PR #3834, opened 2018-10-21
Add palindrome check in JS

IvanFon:
Adds a palindrome checker function in Javascript.


arnavb:
Actually, never mind. This isn't the correct folder for palindrome stuff. It's actually [this](https://github.com/OpenGenus/cosmos/tree/master/code/string_algorithms/src/palindrome_checker). Unfortunately, there is already a JS implementation, meaning this will have to be closed. On the other hand, you could add recursive and iterative implementations to the JS code.

IvanFon:
Sorry about that, didn’t see the other folder! If I add additional implementations I’ll open another PR.

Thanks!

PR #3835, opened 2018-10-21
Add duplicate removal in Go

IvanFon:
Adds duplicate removal string algorithm in Go.

### DIFF
--- a/code/string_algorithms/src/remove_dups/remove_dups.go
+++ b/code/string_algorithms/src/remove_dups/remove_dups.go
@@ -1,0 +1,30 @@
+package main
+
+import (
+  "fmt"
+  "regexp"
+  "strings"
+)
+
+var re, _ = regexp.Compile("[^a-z]")
+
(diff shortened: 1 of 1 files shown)


code/string_algorithms/src/remove_dups/remove_dups.go
```diff
@@ -0,0 +1,30 @@
+package main
+
+import (
```
arnavb: Same problems here:
- Gofmt your code.
- Fix naming conventions.

code/string_algorithms/src/remove_dups/remove_dups.go
```diff
@@ -0,0 +1,30 @@
...
+  "fmt"
+  "regexp"
+  "strings"
+)
+
+var re, _ = regexp.Compile("[^a-z]")
```
arnavb: Is it possible to either not make this global or make it constant?

PR #3837, opened 2018-10-21
Add leap year finder in Go

IvanFon:
Adds leap year finder in Go.

### DIFF
--- a/code/unclassified/src/leap_year/leap_year.go
+++ b/code/unclassified/src/leap_year/leap_year.go
@@ -2,7 +2,7 @@ package main
 
 import "fmt"
 
-func isLeapyear(year int) bool {
+func IsLeapYear(year int) bool {
 	if year%400 == 0 {
 		return true
 	}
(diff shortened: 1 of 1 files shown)


code/unclassified/src/leap_year/leap_year.go
```diff
@@ -0,0 +1,43 @@
+package main
```
arnavb: Please format this code according to `gofmt`, as seen [here](https://play.golang.org/p/ud8YUyAikXP).

code/unclassified/src/leap_year/leap_year.go
```diff
@@ -0,0 +1,43 @@
...
+
+import "fmt"
+
+func isLeapyear(year int) bool {
```
arnavb: Also, these functions should start with an uppercase letter, since I assume they are public functions.

arnavb:
capture

Issue #3840, opened 2018-10-21
Add Linked List in Haskell


gustavoaca1997:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

Implementation of a Linked List in Haskell

PR #3842, opened 2018-10-21
linked list haskell

gustavoaca1997:
**Fixes issue:**

#3840

**Changes:**

Implementation of a Linked List in Haskell.

### DIFF
--- a/code/data_structures/src/linked_list/LinkedList.hs
+++ b/code/data_structures/src/linked_list/LinkedList.hs
@@ -1,3 +1,5 @@
+{-/* Part of Cosmos by OpenGenus Foundation */-}
+
 module LinkedList where
 import Data.Maybe
 {-| A node for a linked list of type `a`-}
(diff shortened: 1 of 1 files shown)


code/data_structures/src/linked_list/LinkedList.hs
```diff
@@ -0,0 +1,52 @@
+{-/* Part of Cosmos by OpenGenus Foundation */-}
```
arnavb: Why not just use single line comments? (`--`)

gustavoaca1997: I though that it is suppose to be block comments, as `/*` is for block comments in C. Sorry about that. Thanks for reviewing.

arnavb: Sure, but no need to apologize! This isn't a mistake, more like a preferred style by the community here.

PR #3858, opened 2018-10-25
Add average fn in Nim

Jay9596:
Add average of list  implementation in Nim

### DIFF
--- a/code/unclassified/src/average/average.nim
+++ b/code/unclassified/src/average/average.nim
@@ -1,0 +1,7 @@
+func average(arr: seq[int]): int =
+    for i in arr:
+      result += i
+  
+# echo average(@[1,2,3,4,5])
+# echo @[1,2,3,4,79,3,6,59,92].average
+  


code/unclassified/src/average/average.nim
```diff
@@ -0,0 +1,7 @@
+func average(arr: seq[int]): int =
+    for i in arr:
+      result += i
```
arnavb: This function doesn't compute the average. It only computes the sum of the sequence. Please fix this.

Jay9596: Sorry my bad. I made the changes and also added a `isMainModule`

code/unclassified/src/average/average.nim
```diff
@@ -0,0 +1,9 @@
+func average(arr: seq[int]): float =
+    for i in arr:
+        result += i.float()
+    result / arr.len().float()
+
+when isMainModule:
+
```
arnavb: Remove this newline.

PR #3859, opened 2018-10-25
Add palindrome check in Nim

Jay9596:
Add palindrome checker implementation in Nim

### DIFF
--- a/code/utility/src/palindrome/palindrome_check/palindrome.nim
+++ b/code/utility/src/palindrome/palindrome_check/palindrome.nim
@@ -1,14 +1,15 @@
 func is_palindrome(str: string): bool =
     let length = str.len - 1
     for i in 0..(length mod 2):
-      if str[i] != str[length - i]:
-        return false
+        if str[i] != str[length - i]:
+            return false
     true
-  
-func test() =
+
+from unicode import reversed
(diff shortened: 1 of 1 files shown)


code/utility/src/palindrome/palindrome_check/palindrome.nim
```diff
@@ -0,0 +1,14 @@
+func is_palindrome(str: string): bool =
+    let length = str.len - 1
+    for i in 0..(length mod 2):
+      if str[i] != str[length - i]:
+        return false
+    true
+  
+func test() =
+    assert("madam".is_palindrome)
+    assert(not is_palindrome("nimrod"))
```
arnavb: Please use UFCS here too. Also I would make this is a `main` function, called `when isMainModule`. Then this should be good to merge. You may also consider adding recursive and string-reversing implementations of this algorithm in this same file.

code/utility/src/palindrome/palindrome_check/palindrome.nim
```diff
@@ -0,0 +1,14 @@
...
+
+# Run test
+# test()
+
```
arnavb: Add exactly _one_ newline to the end of the file.

arnavb:
djldk 2

PR #3896, opened 2018-10-30
Project Euler 1 ao 5 - (Apredendo)

leofagomes:
**Fixes issue:**

**Changes:**


arnavb:
@leofagomes Sorry, these problems are already implemented and we only accept English code here! Sorry! You can always try improving some of the many solutions here, however.

PR #3898, opened 2018-10-31
Add documentation for Levenshtein distance

TwoTau:
Added a README for the [Levenshtein distance algorithm](https://en.wikipedia.org/wiki/Levenshtein_distance)

**Fixes issue:**
#331

**Changes:**
- Added an explanation and psuedocode for the Levenshtein distance string similarity algorithm


arnavb:
djldk 2

PR #3920, opened 2019-01-11
Added karatsuba multiplication in C++

Aftaab99:
**Changes:**
Added Karatsuba multiplication of n-bit integers in C++


code/mathematical_algorithms/src/karatsuba_multiplication/karatsuba_multiplication.cpp
```diff
@@ -0,0 +1,96 @@
+#include <iostream>
+using std::string;
+using std::cout;
+using std::max;
+using std::endl;
```
arnavb: While this isn't really _bad_, our codebase uses `std::` everywhere. For the sake of consistency and as our [style guide](https://github.com/OpenGenus/cosmos/tree/master/guides/coding_style/c%2B%2B) says, please follow this rule. Also, use `\n` instead of `endl`.

arnavb:
djldk 2

PR #3924, opened 2019-01-13
Create m_coloring.cpp

Xayie:
**Fixes issue:**

**Changes:**


Xayie:
A solution to the m coloring problem using C++.

Issue #3928, opened 2019-01-20
SyntaxError in linkedlist.py


rahuldecoded:
**This is a(n):**

- [ ] New algorithm
- [x ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

![screenshot 166](https://user-images.githubusercontent.com/15686569/51445506-dff9cc80-1d2b-11e9-9906-6442f4c76d79.png)

rahuldecoded:
I'm taking this and improving the code in OOP style.

PR #3929, opened 2019-01-20
 Resolved SyntaxError and modified the code  

rahuldecoded:
**Fixes issue:**

#3928

**Changes:**

1. Modified the code
2. Added proper docstrings

### DIFF
--- a/code/data_structures/src/linked_list/linked_list.py
+++ b/code/data_structures/src/linked_list/linked_list.py
@@ -1,52 +1,90 @@
 ''' simple prpgrame for single linked list in python '''
 
 
-Head=None   ''' head node, initilase Null or none '''
-Tail=None   ''' tail node, initilase Null or none '''
+class Node:
+    '''
 
-'''  object class '''
+    This is class to create a node
 
-class A:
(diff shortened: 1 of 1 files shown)


code/data_structures/src/linked_list/linked_list.py
```diff
@@ -1,52 +1,90 @@
...
+            current.next = newNode
+
+
+    def printLinkedList(self):
+        '''
+            The function to traverse and print the data elements present in each node
+        '''
+
+        if self.head == None:
+            print("No Liked List exists")
```
arnavb: Would you consider raising an exception instead?

code/data_structures/src/linked_list/linked_list.py
```diff
@@ -1,52 +1,90 @@
...
+                current = current.next
+            current.next = newNode
+
+
+    def printLinkedList(self):
+        '''
+            The function to traverse and print the data elements present in each node
+        '''
+
+        if self.head == None:
```
arnavb:
```suggestion
        if self.head is None:
```

Issue #713, opened 2017-10-06
Sum of left leaves in Binary Tree 


nimit95:
I would like to implement it in C++

AdiChat:
Sure, Go for it 👍 

Be sure to check existing implementations 👍

PinkiePie1189:
Hi! I like this project a lot and I would really like to help :D. Not sure how is the implementation policy here tho. Should I just write a .cpp with a function sum(int tree[]) in it and make a PR with it?

AdiChat:
@PinkiePie1189 Glad you liked it 😍 You can make this great. 

There is room for every style of implementation, just make sure there are no duplicate files. Browse through our code base and submit a pull request for any problem which you think should belong here 👍 

You can take a look at our [Contribution guidelines](https://github.com/OpenGenus/cosmos/wiki/contribute) as well 👍

rstalekar:
I will do this in C

rstalekar:
Yes , I too didn't understand this , does this mean we need to sum all the leaves of the left subtree from the root or is it sum of  the left leaves of the left subtree from the root.
Please clarify.

AdiChat:
In my opinion, he meant the sum of all leaves having the property of being the left child of its parent node 👍

PinkiePie1189:
@AdiChat if that's the case,doesn't that mean that we only need to all sum even numbered nodes,which makes the implentation trival?

AdiChat:
@PinkiePie1189 Yes, you are right 👍 

In array representation (root at index 1), the solution will be the sum of all even numbered nodes.
In pointer representation, we need to traverse the tree and check if the left child is a leaf or not.

In both cases, the implementation is trivial 😄

nimit95:
@PinkiePie1189 you are right but the array representation would take extra memory in the binary tree.

rstalekar:
Ok I am implementing in C.

rstalekar:
I have created pull request, please review.

PR #3980, opened 2019-03-11
Python implementation for sum of left leaves in Binary Tree

u-wut-m8:
**Fixes issue:**

Issue #713 

**Changes:**

Added Python solution to Issue #713


u-wut-m8:
Yes, it is formatted using the latest version of black.

arnavb:
djldk 2

PR #3981, opened 2019-03-12
Add Python one liner to find greatest digit in a number.

u-wut-m8:
**Fixes issue:**

Add Python one liner to find greatest digit in a number.

**Changes:**

### DIFF
--- a/code/mathematical_algorithms/src/greatest_digit_in_number/greatest_digit_in_number.py
+++ b/code/mathematical_algorithms/src/greatest_digit_in_number/greatest_digit_in_number.py
@@ -1,17 +1,2 @@
 #!/usr/bin/python3
-
-
-def greatest_digit(n):
-    max_digit = int(n % 10)
-
-    while n:
-        if max_digit < n % 10:
-            max_digit = n % 10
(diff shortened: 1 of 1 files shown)


u-wut-m8:
Ok. Thank you for our response.
I will keep that in mind when making further pull requests.

Vaibhavi24:
We could include this as a short-solution of the original program @arnavb

arnavb:
@u-wut-m8 You could create a function in the file called `greatest_digits_with_builtins` that contains your code (and uses a parameter instead).

code/mathematical_algorithms/src/greatest_digit_in_number/greatest_digit_in_number.py
```diff
@@ -13,5 +13,10 @@ def greatest_digit(n):
     return max_digit
 
 
+def greatest_digits_with_builtins(num):
+    print("Greatest digit is: {}".format(int(max([*str(num)]))))
```
arnavb: Return the value instead.

arnavb:
djldk 2

PR #3986, opened 2019-03-15
Create BresenhamCircleDrawingAlgorithm.cpp

piyush26c:
Add implementation of Bresenham's Circle Drawing Algorithm in C++.

### DIFF
--- a/code/computational_geometry/src/BresenhamCircleDrawingAlgorithm.cpp
+++ b/code/computational_geometry/src/BresenhamCircleDrawingAlgorithm.cpp
@@ -1,0 +1,83 @@
+#include <iostream>
+#include <graphics.h>
+
+class BresenhamCircle
+{
+
+public:
+    BresenhamCircle(int radius_) : radius_(radius_) { }
+
+    void getRadiusCenter();
+
+    void drawBresenhamCircle();
+
+    void displayBresenhmCircle(int xc_, int yc_, int x, int y);
+
+private:
+    int radius_;
+    int xc_;
+    int yc_;
+
+};
+
+void BresenhamCircle::drawBresenhamCircle()
+{
+    int x = 0, y = radius_;
+    int decesionParameter = 3 - 2 * radius_;
+    displayBresenhmCircle(xc_, yc_, x, y);
+    while (y >= x)
+    {
+        x++;
+        if (decesionParameter > 0)
+        {
+            y--;
+            decesionParameter = decesionParameter + 4 * (x - y) + 10;
+        }
+        else
+            decesionParameter = decesionParameter + 4 * x + 6;
+        displayBresenhmCircle(xc_, yc_, x, y); //displaying all the Eight Pixels of (x,y)
+        delay(30);
+    }
+}
+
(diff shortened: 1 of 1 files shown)


piyush26c:
Is this code been reviewed?

code/computational_geometry/src/BresenhamCircleDrawingAlgorithm.cpp
```diff
@@ -0,0 +1,82 @@
...
+    putpixel(xc_+y, yc_+x, WHITE);
+    putpixel(xc_-y, yc_+x, WHITE);
+    putpixel(xc_+y, yc_-x, WHITE);
+    putpixel(xc_-y, yc_-x, WHITE);
+}
+
+int main()
+{
+    int gd = DETECT, gm;
+    initgraph(&gd, &gm, NULL);
```
arnavb: I’m not saying to not use graphics functions I’m saying to include a more modern one in the repository itself. Then you can include this specific file locally. For example you can use SDL2.

piyush26c: Sorry as before I was not knowing about SDL, but when I mined about SDL we can just use their inbuilt functions to draw a line, circle and many more and the ALGORITHM OF LINE AND CIRCLE DRAWING which students refer usually is graphics .h when they learn for the very first time. I have to give an implementation of ALGORITHM in C++ for early bird users which they can understand easily.

piyush26c: If this is not accepted then will it be acceptable if I write only the function which performs the circle and line drawing which would be then included locally in SDL, or both the files(function template and one including graphics) should I pull the request?

arnavb: Well, to be honest, I'd rather you at least include the `graphics.h` file in the repository itself, just so the code works. The idiomatic alternative would be to use a more modern graphics library. However, if this proves to be too troublesome, then just include `graphics.h` in the repository.

piyush26c: Sorry for inconvenient replies as i was not able to figure out the requirement and your instructions sir, i will make the appropriate changes...

arnavb: No, its fine. Ideally we would want to be able to link the file as well, but what you had in your most recent PR was good.

piyush26c: To be honest i tried my best to include graphics.h in repository but i don't know why it is not happening as only single file is being included each time you can also check my commits...

arnavb:
@piyush26c I think it would be easier if you learned how to use git from the command line. Here's a basic guide: https://product.hubspot.com/blog/git-and-github-tutorial-for-beginners (there are others online). The online committing system in GH is fairly limited.

Also, don't give up! Your code is good, we are on the final stages of merging! Don't let things like this discourage you.

piyush26c:
Alright, i will reopen this request

Issue #3934, opened 2019-02-01
Duplicate codes for linked list


FR4NKESTI3N:
there is data_structures/src/list/singly_linked_list, but there is also a code for same problem in data_structures/linked_list/
One of them ought to be removed.

rucha174:
Looks like this has been fixed

FR4NKESTI3N:
@rucha174 Sorry, I had written wrong file path. Both directories still exist in repo.

rucha174:
Is the filename correct?

arnavb:
@FR4NKESTI3N @rucha174 If either of you would like to resolve this issue, then by all means go for it! Make sure to reference this issue however.

rucha174:
I have sent a PR. Please review it.

code-blooded:
Why is this issue still open ? @arnavb @rucha174

arnavb:
@code-blooded The old directory isn't completely gone yet. We moved from one file to removing the whole directory, which is pretty much a duplicate of the other one.

nisarg1499:
@arnavb Still same code of linked list is seen in both the paths.
Is this issue still open?

rucha174:
@arnavb Hello! Should I make the required changes in the code?

nitashadhingra:
under GSSoc i would like to contribute @AdiChat

shreeya13garg:
I want to contribute in it..I think I would be able to remove the duplicate directory.

arnavb:
@shreeya13garg / @nitashadhingra One of you can create a PR and fix this.

shreeya13garg:
I have removed the singly linked list from  data_structures/src/list/singly_linked_list to remove duplicates. Should I push it on Github?

shreeya13garg:
I have cloned the cosmos repository and committed the changes i.e removing duplicate files now I need permission to push it on github

shreeya13garg:
My PR in which I have added the modified code-:https://github.com/shreeya13garg/cosmos_PR 
If there is any error let me know.

arnavb:
@shreeya13garg That isn't how you create a pull request. I think you should follow this [guide](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request) and become proficient with git.

hanzohasashi33:
Can I take up this issue if it is still open ? @arnavb

arnavb:
[...]

PR #3990, opened 2019-03-18
Remove the directory

rucha174:
**Fixes issue:**
This fixes issue #3934

**Changes:**
cpp and c files from this (cosmos/code/data_structures/src/linked_list/)  directory were better compared to the other directory ()code/data_structures/src/list/singly_linked_list/. So replace the code. python java and js code seemed fine, changed nothing in them. This directory is redundant.


arnavb:
djldk 2

arnavb:
@rucha174 Could you open a new PR getting completely rid of the old directory (and picking better files if there are two in the same programming language?)

PR #3991, opened 2019-03-18
Replaced the contents for cpp file

rucha174:
**Changes:**
Replaced the contents of code/data_structures/src/list/singly_linked_list/singly_linked_list.cpp with code/data_structures/src/linked_list/linked_list.cpp because better code found


arnavb:
djldk 2

PR #3992, opened 2019-03-18
Add Implementation of Bresenham Circle Drawing Algorithm

piyush26c:


piyush26c:
Please review this implementation and as per suggestion, I had included graphics.h header file.

code/computational_geometry/src/BresenhamCircleDrawingAlgorithm.cpp
```diff
@@ -0,0 +1,82 @@
+#include <iostream>
```
arnavb: Fix folder location.

Issue #4000, opened 2019-03-25
Wrong code at fast_inverse_sqrt.cpp   


Vaibhavi24:
**This is a(n):**

- [ ] New algorithm
- [x] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

The code at          cosmos/code/mathematical_algorithms/src/fast_inverse_sqrt/fast_inverse_sqrt.cpp  fails at testcases like 100.00 etc. Submit correct code.

arnavb:
Good catch! If anyone wants to take on this issue, please say so here.

ratulSharker:
Yes, i'm willing to address the issue, about to submit a PR about this.

ratulSharker:
I checked the code [source](https://stackoverflow.com/questions/1349542/john-carmacks-unusual-fast-inverse-square-root-quake-iii), the optimization they discussed is all about `float`, but here `double` is used. There is a magic number `0x5f375a86`, which i believe needed to be updated to work for `double`.

hsayy:
I've Created a PR. 
The algorithm has been updated with the new magical number `0x5fe6eb50c7b537a9` ([Source](https://stackoverflow.com/questions/11644441/fast-inverse-square-root-on-x64))
for double now gives correct output for test case 100.00.

arnavb:
Fixed in #4039

PR #4039, opened 2019-04-22
Update fast_inverse_sqrt.cpp

hsayy:
The Algorithm now works for double values i.e. 100.00
Magical number for double is 0x5fe6eb50c7b537a9 , [Source](https://stackoverflow.com/questions/11644441/fast-inverse-square-root-on-x64 "Fast Inverse Squareroot on x64")

**Fixes issue:**

#4000

**Changes:**

Magical number for double uas been updated.
Support for double type is added.


arnavb:
djldk

PR #4041, opened 2019-04-23
Static memory allocation

shreyagupta30:
**Fixes issue:**


code/languages/c/README.md
```diff
@@ -0,0 +1,53 @@
+# Introduction to C
+C is a general-purpose, imperative computer programming language, supporting structured programming, lexical variable scope and recursion, while a static type system prevents many unintended operations. By design, C provides constructs that map efficiently to typical machine instructions, and it has therefore found lasting use in applications that were previously coded in assembly language. Such applications include operating systems, as well as various application software for computers ranging from supercomputers to embedded systems. **It is a Turing Language**
```
arnavb: Turing-Complete language.

shreyagupta30: Hi, I have made the required changes.

code/languages/c/static_memory_allocation/README.md
```diff
@@ -0,0 +1,27 @@
...
+static int a = 32;
+```
+Technically, extern and static are two separate classes of variables in C.
+
+```c
+extern int a; /* Declaration */
+int a; /* Definition */
+```
+
+It's somewhat confusing to say that static memory is allocated at compile time, especially if we start considering that the compilation machine and the host machine might not be the same or might not even be on the same architecture.
```
Vaibhavi24: This text has been taken from https://stackoverflow.com/questions/8385322/difference-between-static-memory-allocation-and-dynamic-memory-allocation . Pls give proper credits.

shreyagupta30: I have made the changes as told.

code/languages/c/static_memory_allocation/example.c
```diff
@@ -0,0 +1,27 @@
+#include <stdio.h>
+
+void func(void);
+ 
+static int count = 5; /* global variable */
+
+int  
+main()
```
aditisrinivas97: Change to `int main(void)` - This makes it more clear. In C (unlike cpp), if a function signature doesn’t specify any argument, it means that the function can be called with any number of parameters or without any parameters. So this means,  if you define it as `int main()` you can call main with no parameters or any number of parameters.

shreyagupta30: done!

Issue #4061, opened 2019-05-05
selection_sort.php does not work


parthamittra:
**This is a(n):**

- [ ] New algorithm
- [ ] Update to an existing algorithm
- [x ] Error
- [ ] Proposal to the Repository

**Details:**
selection_sort.php in cosmos/code/sorting/src/selection_sort/selection_sort.php does not work.  When I run it on the cmd line 'php selection_sort.php' I get the following results
parthamittra@system76-pc:~/OpenGenus/cosmos/code/sorting/src/selection_sort$ php selection_sort.php
Array
(
    [0] => 1
    [1] => 3
    [2] => 43
    [3] => 2
    [4] => 14
    [5] => 53
    [6] => 12
    [7] => 87
    [8] => 33
)
Array
(
    [0] => 1
    [1] => 3
    [2] => 43
    [3] => 43
    [4] => 14
    [5] => 53
    [6] => 53
    [7] => 87
    [8] => 33
)

I have already solved the issue and written new code .  I just have no rights to create new branch for the pull request

parthamittra:
please assign this issue to me as I have already written the code and committed to my local repository.  I just have no rights on the remote repository to create my remote dev branch.

arnavb:
Done.

arnavb:
Just create a PR and I'll review, then merge.

parthamittra:
will do

parthamittra:
@arnavb   I don't have rights to the repository so I cannot push my remote dev branch which has the fix in it for selection_sort.php.

arnavb:
@parthamittra Did you fork this repository and create your branch there? Once you do that, you create a PR here.

parthamittra:
Got it.   I already forked and worked on it there.

parthamittra:
pull request is created from my forked branch

PR #4064, opened 2019-05-07
selection_sort.php fixed

yuvag:
**Fixes issue:**

#4061

**Changes:**

Changed the way array was traversed through the loop.


arnavb:
@yuvag This is late, but could you do what Vaibhavi said?

PR #4072, opened 2019-05-12
Tuple in python

shreyagupta30:
**Fixes issue:**

**Changes:**

### DIFF
--- a/code/languages/python/tuple/README.md
+++ b/code/languages/python/tuple/README.md
@@ -29,7 +29,7 @@ string and list reversal mechanism works for tuples as well:
 
 ```python
 >>> t[::-1]
-('corge', 'quux', 'qux', 'baz', 'bar', 'foo')
+('corge', 'quux', 'aux', 'baz', 'bar', 'foo')
 ```
 
 ### Why use a tuple instead of a list?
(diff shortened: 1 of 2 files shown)


code/languages/python/tuple/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Tuple is a collection of Python objects much like a list. The sequence of values stored in a tuple can be of any type, and they are indexed by integers. The important difference between a list and a tuple is that tuples are immutable. 
+
+Tuples are immutable, and usually, they contain a sequence of heterogeneous elements that are accessed via unpacking or indexing (or even by attribute in the case of named tuples). Lists are mutable, and their elements are usually homogeneous and are accessed by iterating over the list.
+
+### Defining and Using Tuples
+
+* Tuples are identical to lists in all respects, except for the following properties:
+
+* Tuples are defined by enclosing the elements in parentheses (()) instead of square brackets ([]).
```
yedhink: Add these points as sub-lists of the above sentence.

code/languages/python/tuple/README.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+string and list reversal mechanism works for tuples as well:
+
+```python
+>>> t[::-1]
+('corge', 'quux', 'qux', 'baz', 'bar', 'foo')
+```
+
+### Why use a tuple instead of a list?
```
yedhink: Tuple unpacking is a very much essential concept that relates with tuple. Do show an example of tuple unpacking. Also tuples can be created without explicitly having those parens. Do mention that tuples can be created that way, with an example.

code/languages/python/tuple/README.md
shreyagupta30: Done!

AdiChat:
Looks great 😍 

message

PR #4073, opened 2019-05-14
Added initial script for metadata generation with only location

yedhink:
Added script for generating metadata by taking into consideration the categories, and generating metadata with only location of category.


AdiChat:
Looks great 😍 

positive_opengenus

PR #4080, opened 2019-05-23
Implementation of Union-Find DS

saddhu1005:
**Fixes issue:**

python3 code and documentation for implementation of Union-Find data structure

**Changes:**

Added the implementation code(in python3) and documentation(detailed doc in iq.opengenes.org) for Union-Find Data Structure.
Created a function to solve the problem of finding a cycle in an undirected graph using the disjoint set (union-find) functions.


agzuniverse:
You should add a main function, called when `if __name__ == '__main__'`.

saddhu1005:
> 
> 
> You should add a main function, called when `if __name__ == '__main__'`.

Added ```main``` function in the code.

PR #4115, opened 2019-08-06
Updated Implementation of Fibonacci in C++

adevasia:
**Changes:**
Added a upper limit check
Put namespace into the function
Renamed functions and variables to useful names
Changed array to vector
Made changes to documentation to follow the Style Guide

### DIFF
--- a/code/dynamic_programming/src/fibonacci/fibonacci.cpp
+++ b/code/dynamic_programming/src/fibonacci/fibonacci.cpp
@@ -1,0 +1,28 @@
+from sys import argv
+
+
+def fibonacci(n):
+    """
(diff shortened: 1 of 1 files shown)


code/dynamic_programming/src/fibonacci/fibonacci.cpp
```diff
@@ -0,0 +1,47 @@
...
+    int n_fib = get_fib(number);
+
+    cout << "Fibonacci number is: " << n_fib << endl;
+
+    return 0;
+}
+
+int get_fib(int num)
+{
+    using namespace std;
```
nisarg1499: Why to keep namespace in individual functions?
Instead I would suggest you to keep it global.

Issue #3650, opened 2018-08-13
Add code for Largest Independent Set


skywalker212:
**This is a:**

- [ ] New algorithm
- [x] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

https://github.com/OpenGenus/cosmos/tree/master/code/dynamic_programming/src/longest_independent_set
The repo folder exists but there is no code or problem description, I am working on this solution.

PR #4201, opened 2019-10-08
#3650 Added code for longest independent set problem

antrix190:
**Fixes issue:**
#3650 Added code for longest independent set problem

**Changes:**
Adds code for longest independent set problem which did not exist earlier.


antrix190:
@arnavb please review.

PR #4213, opened 2019-10-12
Added efficient algo to update(add numbers) to an array

arvindh75:
**Fixes issue:**

**Changes:**


Babatunde13:
You can use the Gauss sum to reduce time and space complexity.
e.g def gauss(N):
        return N * (N + 1)

This is a python code though.

PR #4217, opened 2019-10-13
Added factorial program in C++ and Java

iamrajiv:


code/divide_conquer/src/factorial/factorial.cpp
```diff
@@ -0,0 +1,11 @@
+#include<iostream> 
+using namespace std; 
+  
+// function to find factorial of given number 
+
+unsigned int factorial(unsigned int n) 
+{ 
+    if (n == 0)
```
ellial: I think you should add brackets, an if statement only includes code that is on the same line or enclosed in brackets in the following lines

iamrajiv:
hey @ellial I have updated the code please take a look

PR #4286, opened 2019-10-23
addition_using_bits.cpp

oggy420:
**Fixes issue:**

 no issues

**Changes:**

added a comment :)


achen9:
Hi @oggy420, just wanted to follow-up on this. It appears your comment is about an "if-else" statement and doesn't seem relevant to the algorithm. Could you explain a bit more on why you want to add this comment?

achen9:
Hi @AdiChat, it's been 1 week since I reached out to @oggy420 but there has been no response. I recommend not merging this request and closing this issue since the comment does not seem relevant to explaining or enhancing understanding of the algorithm. Without feedback from @oggy420, this issue can't be moved forward much more.

PR #4325, opened 2019-12-01
Make and Makefile

kshitizsaini113:
**Fixes issue:**

**Changes:**

Added make and makefile concept code in the cosmos


kshitizsaini113:
I have modified as required.Now you may consider the changes

PR #4330, opened 2019-12-08
Added is_empty() function to the stack

### LINKED ISSUE #4271: Adding an is_empty() function to stack in python
**This is a(n):**

- [ ] New algorithm
- [x ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**
The current implementation of stack in python doesn't have a generic is_empty() function, but calls the same assert condition in both peek() and pop() function.
Hence, there is no abstracted way of determining if the stack is really empty without peeking, or popping.

utsavlathia:
**Fixes issue:**

fixes #4271 

**Changes:**

This adds the is_empty() function to the stack algorithm. This also allows this function to be used in other parts of the stack if required. 

## Complexity
O(1). This is a constant time check.

### DIFF
--- a/code/data_structures/src/stack/stack/stack.py
+++ b/code/data_structures/src/stack/stack/stack.py
@@ -14,6 +14,11 @@ class Stack(object):
     def push(self, data):
         self.stack_arr.append(data)
 
+    def is_empty(self):
+        if len(self.stack_arr) > 0:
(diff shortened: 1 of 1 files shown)


AdiChat:
Looks great 😍 

og

PR #4336, opened 2020-01-12
Create Arithmetic_Encoding

etiganou:
-Implementation of Arithmetic Encoding Algorithm. Decimal Library is included because we need precision in computations with large floats.

-Encoding and decoding are both implemented


etiganou:
Required tests for big inputs

PR #4408, opened 2020-03-02
edited the comment in continue.cpp

BhardwajShweta25:
**Fixes issue:**

#include <iostream>

int main()
{
   for (int j=0; j<=8; j++)
   {
      **if (j==5)**
      {
	    /* The continue statement is encountered when
	     * **the value of j is equal to 4.**
	     */
	    continue;
       }

       /* This print statement would not execute for the
	* loop iteration **where j ==4**  because in that case
	* this statement would be skipped.
	*/
       std::cout << j;
   }
   return 0;
}
/*
0 1 2 3 4 6 7 8
*/
**Changes:**

#include <iostream>

int main()
{
   for (int j=0; j<=8; j++)
   {
      if (j==5)
      {
	    /* The continue statement is encountered when
	     * the value of **j is equal to 5**.
	     */
	    continue;
       }

       /* This print statement would not execute for the
	* loop iteration where **j ==5**  because in that case
	* this statement would be skipped.
	*/
       std::cout << j;
   }
   return 0;
}
/*
0 1 2 3 4 6 7 8
*/

### DIFF
--- a/code/language/cpp/loop/continue.cpp
+++ b/code/language/cpp/loop/continue.cpp
@@ -7,13 +7,13 @@ int main()
       if (j==5)
       {
 	    /* The continue statement is encountered when
-	     * the value of j is equal to 4.
+	     * the value of j is equal to 5.
 	     */
 	    continue;
        }
 
        /* This print statement would not execute for the
-	* loop iteration where j ==4  because in that case
+	* loop iteration where j ==5  because in that case
 	* this statement would be skipped.
 	*/
        std::cout << j;


AdiChat:
Looks great 😍 

You have made a wonderful contribution. 
Join us [**at our community forum**](https://discourse.opengenus.org/signup) :family_man_woman_girl_boy: to get connected and access to community benefits

If you would like to [join our internship program](https://discourse.opengenus.org/t/internship-guidelines-at-opengenus/2335/) and contribute to [OpenGenus IQ](https://iq.opengenus.org/), let us know and we will be happy to have you onboard 

og (18)

PR #4433, opened 2020-03-04
Create quicksort algorithm

richabh456:
**Fixes issue:**

**Changes:**


richabh456:
i have added quicksort algorithm taking pivot element along with example.

PR #4489, opened 2020-03-10
unset range of bits

anushkajha:
**Fixes issue:**

the code added unset a  range of bits .

sample --
Input
Enter a number
31
Enter range 
1 
3
New number generated after clearing bits from 1 to 3 : 17

explaniation-
if input is 31 (binary form is 11111)'
and range is 1 to 3
then it would change it to 
10001
and gives an output 
17


anushkajha:
i have added the code for this in c and c++

anushkajha:
i have made the required changes please review the pull request

PR #615, opened 2017-10-05
I have done the necessary correction.

RajatGarg97:

PR #4493, opened 2020-03-10
Create problem (B) Collecting Packages.cpp

rudrakshi99:
Add codes for Codeforces Round #615 (Div. 3), problem (B) Collecting Packages.cpp


rudrakshi99:
Sir, I deleted the duplicate files and fix the location.

PR #4494, opened 2020-03-10
problem (C) Product of Three Numbers.cpp

rudrakshi99:
Add codes of Codeforces Round #615 (Div. 3), problem (C) Product of Three Numbers.cpp


rudrakshi99:
Sir, I deleted the duplicate files and fix the location.

PR #4497, opened 2020-03-10
Create Utopian Tree.c

Srashtimittal:
Eighth Contribution:"cosmos/code/online_challenges/src/hackerrank/Utopian_Tree/Utopian_Tree.c"


Srashtimittal:
@AdiChat Sir, in this I did not replace any existing  implementation instead I add this new code in hackerrank folder. Also, I will make readme file of this code.

PR #4499, opened 2020-03-10
Create Find Digits.c

Srashtimittal:
Tenth Contribution:"cosmos/code/online_challenges/src/hackerrank/Find_Digits/Find_Digits.c"


Srashtimittal:
@AdiChat Sir, in this I did not replace any existing  implementation instead I add this new code in hackerrank folder. Also, I will make readme file of this code.

PR #4527, opened 2020-03-13
Create tug_of_war.py

Harikrishnan6336:
**Fixes issue:**

PROBLEM : Given a set of n numbers , divide the set in two subsets of n/2 sizes each such that the difference of the sum of two subsets is as minimum as possible

The solution tries every possible subset of half size. We initialize current set as empty and then build it considering that there are two possibilities for each element, either be a part of current set, or be a part of the other subset. We consider both possibilities for every element and when a subset of half size is obtained we check whether this solutions is better than the best solution so far and update as needed

**Changes:**

I have added a new folder for tug_of_war algorithm in backtracking folder.In that folder I have added the python implementation of the same

### DIFF
--- a/code/backtracking/src/tug_of_war/tug_of_war.py
+++ b/code/backtracking/src/tug_of_war/tug_of_war.py
@@ -1,0 +1,105 @@
+# Python program for Tug of War using Backtracking
+
+# Checks possible solution by calling recursion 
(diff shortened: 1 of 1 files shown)


Harikrishnan6336:
@AdiChat  Sir Please review this

Issue #4659, opened 2020-03-22
Added Sudoku for NxN matrix 


aryangulati:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

added a new algorithm

Aayushi1506:
please assign this to me

Aayushi1506:
Thanks

PR #4660, opened 2020-03-22
Added Sudoku for NxN matrix #4659

aryangulati:
**Fixes issue:**
Added Sudoku for NxN matrix #4659

**Changes:**

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp

### DIFF
--- a/code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
+++ b/code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
@@ -1,4 +1,4 @@
-#include<bits/stdc++.h>
+#include<iostream>
 
 using namespace std;
 
--- a/code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
+++ b/code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
@@ -1,4 +1,3 @@
-
 #include<bits/stdc++.h>
 
 using namespace std;
--- a/code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
+++ b/code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
@@ -7,18 +7,18 @@ const int n = 9;
 int grid[n][n];
 
 void input_grid() {
-    int i, j;
-    for (i = 0; i < n; i++) {
-        for (j = 0; j < n; j++) {
+    //int i, j;
+    for (int i = 0; i < n; i++) {
+        for (int j = 0; j < n; j++) {
             cin >> grid[i][j];
         }
     }
 }
 
 void print_grid() {
-    int i, j;
-    for (i = 0; i < n; i++) {
-        for (j = 0; j < n; j++) {
+    //int i, j;
+    for (int i = 0; i < n; i++) {
+        for (int j = 0; j < n; j++) {
             cout << grid[i][j] << " ";
         }
         cout << '\n';
@@ -26,8 +26,7 @@ void print_grid() {
 }
 
 bool can_be_placed(int row, int col, int num) {
-    int ii, jj;
-    ii = 0, jj = col;
+   int ii = 0, jj = col;
     // row check
     while (ii < n) {
         if (grid[ii][jj] == num) {
--- a/code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
+++ b/code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
@@ -1,6 +1,7 @@
 #include<iostream>
 
-using namespace std;
+using std ::cout;
+using std ::cin;
 
 const int n = 9;
 
--- a/code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
+++ b/code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
@@ -1,0 +1,96 @@
+
+#include<bits/stdc++.h>
+
+using namespace std;
+
+const int n = 9;
+
+int grid[n][n];
+
+void input_grid() {
+    int i, j;
+    for (i = 0; i < n; i++) {
+        for (j = 0; j < n; j++) {
+            cin >> grid[i][j];
+        }
+    }
+}
+
+void print_grid() {
+    int i, j;
+    for (i = 0; i < n; i++) {
+        for (j = 0; j < n; j++) {
+            cout << grid[i][j] << " ";
+        }
+        cout << '\n';
+    }
+}
+
+bool can_be_placed(int row, int col, int num) {
+    int ii, jj;
+    ii = 0, jj = col;
+    // row check
+    while (ii < n) {
+        if (grid[ii][jj] == num) {
+            return false;
+        }
+        ii++;
+    }
+    // col check
+    ii = row; jj = 0;
+    while (jj < n) {
+        if (grid[ii][jj] == num) {
+            return false;
+        }
+        jj++;
+    }
+    // 3*3 check
+    int startrow, startcol;
+    startrow = (row / 3) * 3;
+    startcol = (col / 3) * 3;
+    for (ii = startrow; ii < startrow + 3; ii++) {
+        for (jj = startcol; jj < startcol + 3; jj++) {
+            if (grid[ii][jj] == num) {
+                return false;
+            }
+        }
+    }
+    return true;
+}
+
+void sudoku_solver(int row, int col) {
+    // cout<<row<<" "<<col<<'\n';
+    if (row == n && col == 0) {
+        print_grid();
+        return ;
+    }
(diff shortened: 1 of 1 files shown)


code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
```diff
@@ -0,0 +1,96 @@
+
+#include<bits/stdc++.h>
```
arnavb: #1915

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+using namespace std;
```
arnavb: Don't use this.

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
```diff
@@ -0,0 +1,96 @@
...
+            if (grid[ii][jj] == num) {
+                return false;
+            }
+        }
+    }
+    return true;
+}
+
+void sudoku_solver(int row, int col) {
+    // cout<<row<<" "<<col<<'\n';
```
arnavb: Remove this.

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
```diff
@@ -0,0 +1,96 @@
+
```
arnavb:
```suggestion
```

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
```diff
@@ -0,0 +1,96 @@
...
+using namespace std;
+
+const int n = 9;
+
+int grid[n][n];
+
+void input_grid() {
+    int i, j;
+    for (i = 0; i < n; i++) {
```
arnavb: Use pre-increment instead of post-increment. Also declare loop variables inside the loop itself. For example:

```suggestion
    for (int i = 0; i < n; ++i) {
```

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
```diff
@@ -0,0 +1,96 @@
...
+        for (j = 0; j < n; j++) {
+            cout << grid[i][j] << " ";
+        }
+        cout << '\n';
+    }
+}
+
+bool can_be_placed(int row, int col, int num) {
+    int ii, jj;
+    ii = 0, jj = col;
```
arnavb: Declaration and assignment can happen on the same lines.

arnavb:
Also can you clarify how this is different from [this already existing file](https://github.com/OpenGenus/cosmos/blob/master/code/backtracking/src/sudoku_solve/sudokusolve.cpp)?

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
aryangulati: It's universal header file

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
aryangulati: It's a myth it doesn't increase run time

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
arnavb: @aryangulati It's also bad practice. Use `std::` everywhere.

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
aryangulati: It's for user so they can comment it out and check for there case for more clarity

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
aryangulati: It's doesn't make an effect declaring again again we can declare it once it's easy

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
arnavb: @aryangulati See [this](https://stackoverflow.com/a/31816096/6525260) and [this](https://stackoverflow.com/questions/25311011/how-does-include-bits-stdc-h-work-in-c/33764128#33764128). Regardless of whether it affects performance, it definitely not portable.

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
aryangulati: It's just to show users so they can get confusion

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
aryangulati: Waiting will send a proof if u want

arnavb: This repository isn't a programming contest repository. It is a general repository of algorithms for all audiences. Thus, we follow software engineering best practices.

aryangulati:
It's different sudoku as it can print NxN as mentioned

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
arnavb: I think this:

```cpp
int ii = 0;
int jj = col;
```

will be clearer than both of these.

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
arnavb: @aryangulati But it's not following _best practices_. Sure it may not have a major effect on this program specifically, but when someone comes along and see this, they may think its normal and use it in their own code. Reusing loop variables is _never_ a good idea. Each loop should have its own code, especially in C++.

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
arnavb: Also, more reasons not to use this file: https://www.geeksforgeeks.org/bitsstdc-h-c/

aryangulati: If u are looking for this too it's has equal no. Of advantages and disadvantages but for more clarity and some issues suggest by you will change

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
aryangulati: I agree with all but not this because no one ever take my loop and use it whosoever know basic would be able to understand the logic behind is SAME

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
aryangulati: Will do

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
arnavb: I disagree, because it's definitely clearer when you can immediately see where a variable comes from, but that's your opinion.

code/backtracking/src/sudoku_solve/sudoku_solveNxN.cpp
aryangulati: https://isocpp.org/wiki/faq/coding-standards#using-namespace-std
either of the way is fine

aryangulati: also most of the code use this due to decrease the complexity you can see many in the organisation ,too see this 
https://github.com/OpenGenus/cosmos/blob/master/code/backtracking/src/sudoku_solve/sudokusolve.cpp

arnavb: @ The established coding standard in this repository is to not use using namespace std. Even the link you put:

> Look, the whole point of namespaces is to prevent namespace collisions between two independently developed piles of code. The using-directive (that’s the technical name for using namespace XYZ) effectively dumps one namespace into another, which can subvert that goal. The using-directive exists for legacy C++ code and to ease the transition to namespaces, **but you probably shouldn’t use it on a regular basis, at least not in your new C++ code.**

So please remove the using directive. (More info here: https://stackoverflow.com/q/1452721/6525260). In the end it comes down to consistency and best standards.

aryangulati:
please add labels of GSOC to it !!

PR #4676, opened 2020-03-22
Create counting_sort.coffee

Harikrishnan6336:
**Fixes issue:**

**Changes:**

Added counting sort in CoffeeScript


code/sorting/src/counting_sort/counting_sort.coffee
```diff
@@ -0,0 +1,36 @@
+counting_sort = (input) ->
```
arnavb: Rename to countingSort.

PR #4718, opened 2020-03-23
Update class for Python3

PSAB:
**Fixes issue:**
Remove 'object' from class parameter to meet Python 3 standards


code/data_structures/src/stack/stack/stack.py
```diff
@@ -4,7 +4,7 @@
 
 
 # Part of Cosmos by OpenGenus Foundation
-class Stack(object):
+class Stack():
```
arnavb:
```suggestion
class Stack:
```

Issue #4742, opened 2020-03-24
Moving zeroes to the end of array


HariniJeyaraman:
**This is a(n):**

- [ ] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

HariniJeyaraman:
@arnavb @AdiChat Please assign this issue to me.

Jahnavi-Jonnalagadda:
@arnavb I would like to work on this issue in Python. Could you please assign this to me.

15aarushiGhadiya:
I want to work on this issue .please assign this issue to me in java @arnavb

PR #4746, opened 2020-03-24
Moving zeroes to end of array

HariniJeyaraman:
**Fixes issue:**

#4742

**Changes:**

 Moving zeroes to the end of array in C++


HariniJeyaraman:
@arnavb Please review the code

move_zeroes_to_end.cpp
```diff
@@ -0,0 +1,39 @@
+#include <iostream>
+#include <stdlib.h>
```
arnavb: Move this code to the correct folder. Use `cstdlib`. In all for loops, use preincrement.

move_zeroes_to_end.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+    for (int i = 0; i < n; i++)
+        std::cout << arr[i] << " ";
+}
+
+int main() {
+    int n;
+    std::cout << "Enter the size of array : \n";
+    std::cin >> n;
+    int a[n], arr[n];
```
arnavb: This is non standards compliant C++. Use `std::vector` instead.

arnavb:
I would put this code in the unclassified algorithms folder, since I can't think of a better place. Let me know if you have a better idea.

arnavb:
Accidently closed.

arnavb:
The filename should match the folder name exactly.

HariniJeyaraman:
@arnavb Please review

PR #4747, opened 2020-03-24
added power of 2 code in php

abhijeet010304:
**Fixes issue:**

Created code to check a number is power of two or not.

**Changes:**

Added code to check a number is power of two or not  .

### DIFF
--- a/code/bit_manipulation/src/power_of_2/power_of_2.php
+++ b/code/bit_manipulation/src/power_of_2/power_of_2.php
@@ -1,0 +1,17 @@
+<?php
+function is_Power($x)
+{
+
+
+ if (($x & ($x - 1)) != 0)
+		return "$x is not power of 2";
+ else
+    {
+		return "$x is power of 2";
+    }
+
+}
+
+print_r(is_Power(32)."\n");
+print_r(is_Power(31)."\n");
+?>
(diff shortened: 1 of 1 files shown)


code/bit_manipulation/src/power_of_2/power_of_2.php
```diff
@@ -0,0 +1,17 @@
+<?php
+function is_Power($x)
```
arnavb: use camel case. Also beautify your code somewhere.

code/bit_manipulation/src/power_of_2/power_of_2.php
```diff
@@ -0,0 +1,17 @@
...
+ else
+    {
+		return "$x is power of 2";
+    }
+
+}
+
+print_r(is_Power(32)."\n");
+print_r(is_Power(31)."\n");
+?>
```
arnavb: Don't include a closing tag.

code/bit_manipulation/src/power_of_2/power_of_2.php
abhijeet010304: Ok

code/bit_manipulation/src/power_of_2/power_of_2.php
abhijeet010304: OK Sir I have corrected it please review!

PR #4767, opened 2020-03-25
Create simpson.cpp

kichloo:
Added the implementation of Simpson's 1/3 rule in C language.

**Fixes issue:**
Added another rule (Simpson's 1/3 rule) for integration.

**Changes:**
Added the file simpson.cpp


simpson.cpp
```diff
@@ -0,0 +1,45 @@
+#include<stdio.h>
+#include<conio.h>
```
arnavb: Don't use this. Reformat this file using clang-format with 4 spaces. Place this file into the correct directory.

simpson.cpp
```diff
@@ -0,0 +1,45 @@
...
+float f(float x)
+{
+    return(1/(1+x));
```
arnavb: No need for extra parentheses on the outside of the expression.

simpson.cpp
```diff
@@ -0,0 +1,45 @@
...
+}
+void main()
```
arnavb:
```suggestion
int main()
```

simpson.cpp
```diff
@@ -0,0 +1,45 @@
...
+{
+    int i,n;
```
arnavb: Declare i in the loop itself.

simpson.cpp
```diff
@@ -0,0 +1,45 @@
...
+    float x0,xn,h,y[20],so,se,ans,x[20];
```
arnavb: Declare these variables when they are first used, not at the top of the function.

simpson.cpp
```diff
@@ -0,0 +1,45 @@
...
+    scanf("%f%f%f",&x0,&xn,&h);
+    n=(xn-x0)/h;
+    if(n%2==1)
+    {
+        n=n+1;
+    }
+    h=(xn-x0)/n;
+    printf("\n Refined value of n and h are:%d %f\n",n,h);
+    printf("\n Y values: \n");
+    for(i=0; i<=n; i++)
```
arnavb: Pre-increment in all for loops.

simpson.cpp
```diff
@@ -0,0 +1,45 @@
...
+            so=so+y[i];
+        }
+        else
+        {
+            se=se+y[i];
+        }
+ 
+    }
+    ans=h/3*(y[0]+y[n]+4*so+2*se);
+    printf("\n Final integration is %f",ans);
```
arnavb: Use `cout` and `cin` in C++.

simpson.cpp
kichloo: @arnavb This is my first time contributing here. So, could you help regarding the correct directory?

kichloo:
@arnavb Please check #4774 as I did it with the required changes

Issue #4759, opened 2020-03-25
segregate odd even numbers in an array


HariniJeyaraman:
**This is a(n):**

- [ ] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

HariniJeyaraman:
@arnavb Please assign this issue to me. I'm a participant of GSSoC20 and can implement this in C,C++,Python and Java

shortCircuit81:
please assign this to me.
I'm a participant of GSSoC20 and can implement this in C,C++and Java

SifR-01:
I can work on this in C,C++,Java,if possible plz assign

PR #4784, opened 2020-03-25
segregate even odd numbers in an array

HariniJeyaraman:
**Fixes issue:**

#4759

**Changes:**

Segregate even and odd numbers in an array in C++


HariniJeyaraman:
@arnavb Please review this code

code/unclassified/src/segregate_even_odd/segregate_even_odd.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+int main() {
+    int n, val;
+    std::vector<int> a;
+    std::cout << "Enter the number of elements in array :  ";
+    std::cin >> n;
+    std::cout << "Enter the values of the elements :  ";
+    for (int i = 0; i < n; ++i) {
+        std::cin >> val;
+        a.push_back(val);
```
arnavb:
```suggestion
    int n;
    std::cout << "Enter the number of elements in array :  ";
    std::cin >> n;
    std::vector<int> a(n);
    std::cout << "Enter the values of the elements :  ";
    for (int i = 0; i < n; ++i) {
        int val;
        std::cin >> val;
        a[i] = val;
```

Do this for all future PRs with vectors.

HariniJeyaraman: Done ! Will do.

code/unclassified/src/segregate_even_odd/segregate_even_odd.cpp
```diff
@@ -0,0 +1,47 @@
...
+int main() {
+    int n;
+    std::cout << "Enter the number of elements in array :  ";
+    std::cin >> n;
+    std::vector<int> a(n);
+    std::cout << "Enter the values of the elements :  ";
+    for (int i = 0; i < n; ++i) {
+        int val;
+        std::cin >> val;
+        a.push_back(val);
```
arnavb: This line needs to be changed still.

HariniJeyaraman: I cannot write a[ i ] = val; Since a is a vector

HariniJeyaraman: I will get an error if i change this line

arnavb: `std::vector` supports the `[]` operator: https://en.cppreference.com/w/cpp/container/vector/operator_at. Otherwise this code adds past the end of the array and you end up with a vector twice the expected size.

arnavb: Also anywhere you need the size of the array, use `.size()` instead of passing `n` around.

HariniJeyaraman: Got it. Made the changes requested

HariniJeyaraman:
@arnavb Please review again

Issue #4623, opened 2020-03-19
Longest non-palindromic substring in C++


shweta0699:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

shweta0699:
I want to work on this issue as a part of GSSoC'20. Please assign it to me.

Srashtimittal:
@arnavb Please assign me this issue.

Srashtimittal:
@AdiChat  @arnavb Please view my pull request related to this issue.

shweta0699:
Please review my pr too.
Fixes #5541

PR #4813, opened 2020-03-26
Create longest non-palindromic substring in C++

Srashtimittal:
Issue fixed: #4623


code/sorting/src/bubble_sort/bubble_sort.cpp
```diff
@@ -1,50 +1,42 @@
...
+#include <iostream>
+using namespace std;
+int recbub(int arr[],int n)
+{   
+   int i,temp;
+   if(n==1)
+   {
+   return n;
+   }
+    else
```
arnavb: Why did you remove the old code?

Srashtimittal: @arnavb I write bubble sort code using recursion, that's why I remove the old one .
or Can I make another file for this ?

Srashtimittal: @arnavb Please view my other pull requests also.

PR #4817, opened 2020-03-26
changed fibonacci_search.c

shreeya13garg:
**Fixes issue:**

Had to write code for fibonacci_search.c

**Changes:**

Changed code for Fibonacci_search.c


arnavb:
Why is the new code better than the old?

shreeya13garg:
What are the problems in old code which needs to be solved or it the old
code perfect?As this issue is open so I was expecting some changes so if
you can tell me the changes then I can work on it.

PR #4835, opened 2020-03-27
fixed fibonacci_search.c

shreeya13garg:
**Fixes issue:**

Change fibonacci_search.c

**Changes:**

Rewrite that file


arnavb:
@shreeya13garg What makes your code better than the previous? What do you fix? I think that code is fine as is.

PR #4836, opened 2020-03-27
Add maxdiff

wokira:
**Fixes issue:**

Added code to this problem and documented it according to style guide.

**Changes:**

### DIFF
--- a/rgsoc_02.c
+++ b/rgsoc_02.c
@@ -1,3 +1,4 @@
+/*MAXDIFF problem On CodeChef */
 #include <stdio.h>
 
 int main(void) 
--- a/rgsoc_02.c
+++ b/rgsoc_02.c
@@ -1,0 +1,38 @@
+#include <stdio.h>
+
+int main(void) 
+{
+    /* Input N and K */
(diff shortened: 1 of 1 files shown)


wokira:
I have added Java, Python and C implementations of the CodeChef: MAXDIFF problem here.

maxdiff.java
```diff
@@ -0,0 +1,37 @@
+import java.util.*;
```
arnavb: Reformat with 4 spaces and move to correct directory.

maxdiff.py
```diff
@@ -0,0 +1,14 @@
+t=int(input())
+for i in range(t):
+    n,k=map(int,input().split())
+    weights=list(map(int,input().split()))
```
arnavb: Reformat with black.

rgsoc_02.c
```diff
@@ -0,0 +1,39 @@
+/*MAXDIFF problem On CodeChef */
+#include <stdio.h>
```
arnavb: Same filename.

PR #4837, opened 2020-03-27
changed topological_sort.py

shreeya13garg:
**Fixes issue:**

change topological_sort.py

**Changes:**

Changed


code/sorting/src/topological_sort/Topological.py
```diff
@@ -0,0 +1,43 @@
...
+        print(stack)
+print("Enter the size of the directed graph")
+n=int(input()) 
+print("Enter the edges of the directed graph")
+g= Graph(n) 
+for i in range(0,n):
+    val=str(input())
+    arr=val.split(" ")
+    edge1=int(arr[0])
+    edge2=int(arr[1])
```
arnavb:
```suggestion
    arr=val.split(" ")
    edge1, edge2 = map(int, val.split(" "))
```

code/sorting/src/topological_sort/Topological.py
```diff
@@ -0,0 +1,43 @@
...
+    g.addEdge(edge1,edge2)
+
+print("Topological Sort of the given graph-:")
+g.topologicalSort()
```
arnavb: Reformat this file using black.

code/sorting/src/topological_sort/Topological.py
```diff
@@ -0,0 +1,43 @@
+from collections import defaultdict
```
arnavb: The old filename was better.

code/sorting/src/topological_sort/Topological.py
```diff
@@ -0,0 +1,43 @@
+from collections import defaultdict 
+
+class Graph: 
+    def __init__(self,vertices): 
+        self.graph = defaultdict(list) 
+        self.V = vertices 
+  
+    
+    def addEdge(self,u,v):
```
arnavb: Rename all these functions so they use underscore case.

Issue #4466, opened 2020-03-07
Disjoint Set Union in c and c++


shubhammantri1:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

I want to write code for the implementation of Disjoint Set Union datastructure.
Please assign it to me.

Kuljeet-123:
@arnavb  @shubhammantri1  I would like to work on this issue.... can you please assign me this.

Salonii02:
I would like to work on this issue
can you please assign me
@arnavb @shubhammantri1

arnavb:
I have assigned this issue to all three of you. Just decide which programming languages you would like to work on.

Salonii02:
I would be fine with c++.

PR #4842, opened 2020-03-27
Created Disjoint Set Union in Cpp #4466

aryangulati:
**Fixes issue:**
Created Disjoint Set Union

**Changes:**

Added Disjoint Set Union

### DIFF
--- a/code/graph_algorithms/src/Disjoint_Set_Union_algo.cpp
+++ b/code/graph_algorithms/src/Disjoint_Set_Union_algo.cpp
@@ -1,4 +1,4 @@
-#include<bits/stdc++.h>
+#include<iostream>
 
 using namespace std;
 
--- a/code/graph_algorithms/src/Disjoint_Set_Union_algo.cpp
+++ b/code/graph_algorithms/src/Disjoint_Set_Union_algo.cpp
@@ -1,0 +1,55 @@
+#include<bits/stdc++.h>
+
+using namespace std;
+
+#define int long long int
+#define ld long double
+#define F first
+#define S second
+#define pb push_back
+
+const int N = 100001;
+
+int parent[N], sz[N];
+
+// gives superparent of the component
+int get_parent(int x) {
+    if (x == parent[x]) {
+        return x;
+    }
+    else {
+        // path compression
+        return parent[x] = get_parent(parent[x]);
+    }
+}
+
+// joins two disjoint components
+void unite(int x, int y) {
+    int root_x = get_parent(x);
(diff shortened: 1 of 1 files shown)


code/graph_algorithms/src/Disjoint_Set_Union_algo.cpp
```diff
@@ -0,0 +1,55 @@
+#include<bits/stdc++.h>
```
arnavb: #1915
Move code to its own folder
Don't use using namespace std

code/graph_algorithms/src/Disjoint_Set_Union_algo.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+using namespace std;
+
+#define int long long int
+#define ld long double
+#define F first
+#define S second
+#define pb push_back
```
arnavb: Don't use these defines except for the first one; use `lli` not `int`

code/graph_algorithms/src/Disjoint_Set_Union_algo.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+const int N = 100001;
+
+int parent[N], sz[N];
+
+// gives superparent of the component
+int get_parent(int x) {
```
arnavb: camel case.

code/graph_algorithms/src/Disjoint_Set_Union_algo.cpp
```diff
@@ -0,0 +1,55 @@
...
+}
+
+void init() {
+    for (int i = 0; i < N; i++) {
+        parent[i] = i;
+        sz[i] = 1;
+    }
+}
+
+int32_t main()
```
arnavb: int

code/graph_algorithms/src/Disjoint_Set_Union_algo.cpp
```diff
@@ -0,0 +1,55 @@
...
+{
+    init();
+    int n, m;
+    cin >> n >> m;
+    for (int i = 0; i < m; i++) {
```
arnavb: Pre increment.

code/graph_algorithms/src/Disjoint_Set_Union_algo.cpp
aryangulati: basically for sake of using long long int i use int32_t instead it makes int as long long int so its easy and too applicable!

Issue #4030, opened 2019-04-13
Infix to prefix converter in c++ and java 


robotjellyzone:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

This will convert any infix to prefix expressions

robotjellyzone:
Please assign it to me!!

sonali623:
@arnavb @AdiChat is this issue still open? I would like to work on it as a part of gssoc 2020.

Kuljeet-123:
@arnavb @AdiChat I want to implement this in Python as a part of GSSOC'20.Please assign this to me.

nikitajuneja1:
Could you please assign this to me? I will implement the solution in c++.

PR #4850, opened 2020-03-27
Modified code for infix to prefix converter in cpp

Ayush7-BIT:
**Fixes issue:**
#4030

**Changes:**

Modified the code in cpp"


arnavb:
@Ayush7-BIT What's wrong with the existing code?

Ayush7-BIT:
@arnavb all were combined in one na..so i modified some also..is there anything wrong?

arnavb:
Your modified code provides no benefit over the existing one. Why would I remove someone else's previous contribution?

Ayush7-BIT:
Oh that u r saying..ok i will modify it😀..

PR #4791, opened 2020-03-26
Codechef TREEROOT code in C, JAVA and Python with README (Issue #4773)

ka1shi:
**Fixes issue:**
#4773 

**Changes:**
Added C, JAVA and python solution for Codechef Root of the Problem (TREEROOT) Problem.
Added Readme file for that problem.

PR #4855, opened 2020-03-27
TREEROOT.cpp

shivani789:
#4791 

C++ code for the problem


arnavb:
@shivani789 This PR doesn't contain the correct code.

PR #4870, opened 2020-03-28
initial commit

saidrishya:
**Fixes issue:**

Changes:
Added implementation of balanced expressions in python

### DIFF
--- a/code/data_structures/src/stack/balanced_expression/balanced_expression.py
+++ b/code/data_structures/src/stack/balanced_expression/balanced_expression.py
@@ -1,0 +1,18 @@
+def check(expression): 
+      
+    open_tup = tuple('({[') 
+    close_tup = tuple(')}]') 
+    map = dict(zip(open_tup, close_tup)) 
+    queue = [] 
+  
+    for i in expression: 
+        if i in open_tup: 
+            queue.append(map[i]) 
+        elif i in close_tup: 
+            if not queue or i != queue.pop(): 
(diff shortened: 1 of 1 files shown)


code/data_structures/src/stack/balanced_expression/balanced_expression.py
```diff
@@ -0,0 +1,18 @@
+def check(expression): 
+
```
arnavb:
```suggestion
```

code/data_structures/src/stack/balanced_expression/balanced_expression.py
```diff
@@ -0,0 +1,18 @@
+def check(expression): 
+      
+    open_tup = tuple('({[')
```
arnavb: Reformat this file using black python.

code/data_structures/src/stack/balanced_expression/balanced_expression.py
```diff
@@ -0,0 +1,18 @@
...
+    close_tup = tuple(')}]') 
+    map = dict(zip(open_tup, close_tup)) 
+    queue = [] 
+  
+    for i in expression: 
+        if i in open_tup: 
+            queue.append(map[i]) 
+        elif i in close_tup: 
+            if not queue or i != queue.pop(): 
+                return "Unbalanced"
```
arnavb: Function should be called `is_balanced` and should return `True` or `False`.

Issue #847, opened 2017-10-07
Add Cycle Sort


ahhmarr:
Add the code for Cycle sort in any language ( C, C++, Java, Go, Python or any other)

The code should be placed at [code/sorting/cycle_sort](https://github.com/OpenGenus/cosmos/tree/master/code/sorting/cycle_sort)

Note: multiple contributors can work on this issue as it has multiple parts (languages)

Your pull request will be reviewed by maintainers instantly.

For contribution guidelines, see [this](https://github.com/OpenGenus/cosmos/wiki/contribute)

If you need any help, let us know.

ahhmarr:
#848 
for JavaScript

parthsharma2:
#860 implemented in python

amirkarimi:
#865 Go implementation

Jecoms:
I'll implement this in C#

jenkassim:
I'll work on the C++ implementation

j3r3mias:
#930 implemented in C.

PR #4882, opened 2020-03-28
Added code for cycle sort in C#

Ayush7-BIT:
**Fixes issue:**

#847

**Changes:**

Added code for cycle sort in C#


arnavb:
@Ayush7-BIT Already exists [here](https://github.com/OpenGenus/cosmos/blob/master/code/sorting/src/cycle_sort/cycle_sort.cs).

Issue #4975, opened 2020-03-30
Codechef OJUMPS solution in C


ritish099:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

This morning Chef wants to jump a little. In a few minutes he will arrive at the 
point 0. Then he will perform a lot of jumps in such a sequence: 1-jump, 2-jump, 
3-jump, 1-jump, 2-jump, 3-jump, 1-jump, and so on.

1-jump means that if Chef is at the point x, he will jump to the point x+1.

2-jump means that if Chef is at the point x, he will jump to the point x+2.

3-jump means that if Chef is at the point x, he will jump to the point x+3.

Before the start Chef asks you: will he arrive at the point a after some number of 
jumps?

PR #4976, opened 2020-03-30
Codechef OJUMPUS solution in C

ritish099:
**Fixes issue:#4975**

**Changes:**

Added Codechef OJUMPUS solution in C
Added README for Codechef OJUMPUS Problem

### DIFF
--- a/code/online_challenges/src/codechef/OJUMPS/OJUMPS.c
+++ b/code/online_challenges/src/codechef/OJUMPS/OJUMPS.c
@@ -1,0 +1,17 @@
+#include<stdio.h>
+
+int main() 
+{
+    long long int a;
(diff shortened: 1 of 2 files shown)


ritish099:
Sir @arnavb , Check it out!

code/online_challenges/src/codechef/OJUMPS/OJUMPS.c
```diff
@@ -0,0 +1,17 @@
+#include<stdio.h>
+
+int main() 
+{
+    long long int a;
+    scanf("%lld", & a);
+    int b = a % 6;
```
hanzohasashi33: I think the b should also be of type long long int as in some huge testcases, b can be overflowed.

Issue #4988, opened 2020-03-31
Codechef GCD2 Solution in C


ritish099:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

Frank explained its friend Felman the algorithm of 
Euclides to calculate the GCD of two numbers. Then 
Felman implements it algorithm

int gcd(int a, int b)
{
	if (b==0)
		return a;
	else
		return gcd(b,a%b);
}
and it proposes to Frank that makes it but with a 
little integer and another integer that has up to 250 
digits.
Your task is to help Frank programming an efficient 
code for the challenge of Felman.

PR #4989, opened 2020-03-31
Codechef GCD2 Solution in C

ritish099:
**Fixes issue:#4988**

**Changes:**

Added Codechef GCD2 Solution in C
Added README for Codechef GCD2 Problem


ritish099:
Sir @arnavb , Check it out !

PR #5027, opened 2020-04-01
Made Changes

shreetanu:
**Fixes issue:**

**Changes:**

Find Your Gift problem's solution added.

### DIFF
--- a/code/online_challenges/src/codechef/GIFTSRC/GIFTSRC.cpp
+++ b/code/online_challenges/src/codechef/GIFTSRC/GIFTSRC.cpp
@@ -9,10 +9,10 @@ int main() {
 	{
 	    x=y=0;
 	    cin>>n;
-	    for(i=1;i<=n;i++)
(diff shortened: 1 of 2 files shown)


code/online_challenges/src/codechef/GIFTSRC/GIFTSRC.cpp
```diff
@@ -0,0 +1,46 @@
+#include <iostream>
+using namespace std;
```
arnavb, line 2: Don't use this.

code/online_challenges/src/codechef/GIFTSRC/GIFTSRC.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+int main() {
```
arnavb, line 4: Reformat this file with clang format and 4 spaces. loop variables should be declared in the loop itself. All other variables should be declared closest to where they are used.

Issue #5106, opened 2020-04-03
Codechef NUMGAME2 in C


ritish099:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

Alice and Bob play the following game.They choose a 
number N to play with.The runs are as follows :

1.Bob plays first and the two players alternate.

2.In his/her turn ,a player can subtract from N any 
prime number(including 1) less than N.The number 
thus obtained is the new N.

3.The person who cannot make a move in his/her turn 
loses the game.

Assuming both play optimally,who wins the game ?

Ankit152:
please assign this issue to me. i will do it in C++.

PR #5107, opened 2020-04-03
Codechef NUMGAME2 Solution in C

ritish099:
**Fixes issue:#5106**

**Changes:**

Added Codechf NUMWGAME2 solution in C
Added README for codechef NUMGAM2 problem


ritish099:
Sir @arnavb , Check it out.

Issue #5138, opened 2020-04-04
Codechef LEBOMBS solution in C


ritish099:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

The Little Elephant from the Zoo of Lviv currently is on the military mission. 
There are N enemy buildings placed in a row and numbered from left to right 
strating from 0. Each building i (except the first and the last) has exactly two 
adjacent buildings with indices i-1 and i+1. The first and the last buildings have 
just a single adjacent building.

Some of the buildings contain bombs. When bomb explodes in some building it 
destroys it and all adjacent to it buildings.

You are given the string S of length N, where Si is 1 if the i-th building 
contains bomb, 0 otherwise. Find for the Little Elephant the number of buildings 
that will not be destroyed after all bombs explode. Please note that all bombs 
explode simultaneously.

Ankit152:
please assign this to me in C++.

ritish099:
Sir @arnavb , check it out

PR #5139, opened 2020-04-04
Codechef LEBOMBS solution in C

ritish099:
**Fixes issue:#5138**

**Changes:**

Added Codechef LEBOMBS solution in C
Added README for Codechef LEBOMBS problem


ritish099:
Sir @arnavb , check it out!

PR #5178, opened 2020-04-05
largest element in array using C++

sifteeR:
**Fixes issue:**
As my first contribution, I have created a cpp file in max_of_array folder which provides a simple algorithm to find largest element in the array

**Changes:**
Made a new folder max_of_array in which I added a .cpp file

### DIFF
--- a/code/unclassified/src/Max_of_array/Max_of_array.cpp
+++ b/code/unclassified/src/Max_of_array/Max_of_array.cpp
@@ -1,0 +1,27 @@
+ #include <iostream>
+using namespace std;
+
+int main() {
+	//the largest number in an array;
+	//declaring an array
+	int n;
+	cout<<"enter the size of the array"<<endl; //getting the number of elements of the array
+	cin>>n;//taking input
+	int arr[n]; //creating array arr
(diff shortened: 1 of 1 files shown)


code/unclassified/src/Max_of_array/Max_of_array.cpp
```diff
@@ -0,0 +1,27 @@
+ #include <iostream>
+using namespace std;
+
+int main() {
+	//the largest number in an array;
+	//declaring an array
+	int n;
+	cout<<"enter the size of the array"<<endl; //getting the number of elements of the array
+	cin>>n;//taking input
+	int arr[n]; //creating array arr
```
RithvikKasarla, line 10: This line is not standard compliant. You can use a dynamic array or std::vector.

code/unclassified/src/Max_of_array/Max_of_array.cpp
```diff
@@ -0,0 +1,27 @@
+ #include <iostream>
+using namespace std;
```
RithvikKasarla, line 2: This line is bad practice. 
https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice

Issue #5206, opened 2020-04-06
Codechef ANUWTA solution in C


ritish099:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

There are N+1 lights. Lights are placed at (0, 0), (1, 0), (2, 0) ... (N, 0). 
Initially all the lights are on. You want to turn off all of them one after one. 
You want to follow a special pattern in turning off the lights.

You will start at (0, 0). First, you walk to the right most light that is on, turn 
it off. Then you walk to the left most light that is on, turn it off. Then again 
to the right most light that is on and so on. You will stop after turning off all 
lights. You want to know how much distance you walked in the process. Note that 
distance between (a,0) and (b,0) is |a-b|.

ritish099:
Sir @arnavb , check it out

PR #5207, opened 2020-04-06
Codechef ANUWTA solution in C

ritish099:
**Fixes issue:#5206**

**Changes:**

Added Codechef ANUWTA solution in C
Added README for Codechef ANUWTA Problem


ritish099:
Sir @arnavb , Check it out !

Issue #5259, opened 2020-04-07
Codechef DRAGNXOR solution in C


ritish099:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

Did you ever hear about 'Dragon Food' ? Its used to refer to the chocolates bought 
for your loved ones :). Po offers dragon food to master Shifu, who is a famous 
cook in the valley of food. In return, Shifu hands over the dragon scroll to Po, 
which is said to hold the ingredients of the secret recipe. To open the dragon 
scroll, one has to solve the following puzzle.

1. Consider a N-bit integer A. We call an integer A' as shuffle-A, if A' can be 
obtained by shuffling the bits of A in its binary representation. For eg. if N = 5 
and A = 6 = (00110)2, A' can be any 5-bit integer having exactly two 1s in it i.e.
, any of (00011)2, (00101)2, (00110)2, (01010)2, ...., (11000)2.

2. Given two N-bit integers A and B, find the maximum possible value of (A' xor 
B') where A' is a shuffle-A, B' is a shuffle-B and xor is the bit-wise xor 
operator.

Given N, A and B, please help Po in opening the dragon scroll.

Notes 1. xor operator takes two bit strings of equal length and performs the 
logical XOR operation on each pair of corresponding bits. The result in each 
position is 1 if only the first bit is 1 OR only the second bit is 1, but will be 
0 if both are 1 or both are 0. For eg: 5 (0101) xor 3(0011) = 6(0110). In most 
languages it is represented using ^ symbol. 5 ^ 3 = 6.
2. If the integer actually needs less than N bits to represent in binary, append 
sufficient number of leading 0 bits. For eg. as shown in the problem statement for 
N = 5, A = 6 = (00110)2

PR #5260, opened 2020-04-07
Codechef DRAGXOR solution in C

ritish099:
**Fixes issue:#5259**

**Changes:**

Added Codechef DRAGXOR solution in C
Added README for Codechef DRAGXOR problem


ritish099:
Sir @arnavb , Check it out

Issue #5416, opened 2020-04-11
HackerRank Balanced Brackets solution in Java


ramneek008:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

A bracket is considered to be any one of the following characters: (, ), {, }, [, or ].

Two brackets are considered to be a matched pair if the an opening bracket (i.e., (, [, or {) occurs to the left of a closing bracket (i.e., ), ], or }) of the exact same type. There are three types of matched pairs of brackets: [], {}, and ().

A matching pair of brackets is not balanced if the set of brackets it encloses are not matched. For example, {[(])} is not balanced because the contents in between { and } are not balanced. The pair of square brackets encloses a single, unbalanced opening bracket, (, and the pair of parentheses encloses a single, unbalanced closing square bracket, ].

By this logic, we say a sequence of brackets is balanced if the following conditions are met:
1. It contains no unmatched brackets.
2. The subset of brackets enclosed within the confines of a matched pair of brackets is also a matched pair of brackets.
Given n strings of brackets, determine whether each sequence of brackets is balanced. If a string is balanced, return YES. Otherwise, return NO.

PR #5417, opened 2020-04-11
Balanced brackets solution in Java

ramneek008:
**Fixes issue:** #5416 

**Changes:**

Added the solution of Balanced Brackets in Java


Maran23 (commented):
This is not following the Java style guide written in the corresponding README.md.

Maran23 (changes requested):
This is not following the Java style guide written in the corresponding README.md.

code/online_challenges/src/hackerrank/Balanced_Brackets/balanced_brackets.java
```diff
@@ -0,0 +1,83 @@
...
+        {
+            return "YES";
+        }
+        else
+        {
+            return "NO";
+        }
+    }
+    
+    private static final Scanner scanner = new Scanner(System.in);
```
Maran23, line 64: Scanner doesn't need to be a member variable

Maran23 (changes requested):

Issue #5470, opened 2020-04-16
Hackerrank Minimum Distance solution in C


ritish099:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

We define the distance between two array values as the number of indices between 
the two values. Given a, find the minimum distance between any pair of equal 
elements in the array. If no such value exists, print -1.

For example, if a = [3, 2, 1, 2, 3], there are two matching pairs of values: 3 and 
2. The indices of the 3's are i = 0 and i = 4, so their distance is d[i,j] = |j - 
i| = 4. The indices of the 2's are i = 1 and , so their distance is d[i,j] = |j - 
i| = 2.

PR #5471, opened 2020-04-16
Hackerrank Minimum distance solution in C

ritish099:
**Fixes issue:#5470**

**Changes:**

Added Hackerrank Minimum distance solution in C
Added README for Hackerrank Minimum Distance Problem


ritish099:
Sir @arnavb , check it out !

Issue #5481, opened 2020-04-17
Hackerrank Encryption solution in C


ritish099:
**This is a(n):**

- [ ] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

PR #5482, opened 2020-04-17
Hackerrank Encryption solution in C

ritish099:
**Fixes issue:#5481**

**Changes:**

Added Hackerrank Encryption solution in C


ritish099:
Sir @arnavb , Check it out

PR #5542, opened 2020-04-29
pr for making array elements equal

NikitaMasand:
**Changes:**
Added code for the article published at iq.opengenus

**Location**
code/mathematical_algorithms/src/minimum_operations_elements_equal/

**ArticleLink**
https://iq.opengenus.org/make-elements-equal/

**Description**
To find the minimum number of operations to make all the array elements equal. The operation includes incrementing all but one element of the array by 1 that is incrementing N-1 elements out of N elements. The approach uses some trick to make it efficient as explained in the above article with the thought process.


AdiChat:
Looks great 😍 

positive_opengenus

PR #5550, opened 2020-05-02
pr for candy distribution

NikitaMasand:
Changes:
Added code for the article published at iq.opengenus

Location
code/algorithm_applications/src/binary_search/distributing_candies/

ArticleLink
https://iq.opengenus.org/distribute-candies/

Description
Given n candy boxes, each box contains different number of candies. We need to distribute the candies amongst k friends such that:
All friends get equal number of candies
All the candies which a particular friend gets must be from a single box only.
A detailed approach on how to solve the problem using the concept of binary search for optimization is explained in the above article.


AdiChat:
Looks great 😍 

positive_opengenus

Issue #5548, opened 2020-05-02
Prime factors in ruby


atarax665:
**This is a(n):**

- [ ] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [*] Proposal to the Repository

**Details:**

Adding implementation to print all the prime factors of a number in ruby.

PR #5551, opened 2020-05-03
Prime factors implementation in ruby

atarax665:
**Fixes issue:**
#5548

**Changes:**

* Added ruby code to print prime factors of a given number.(prime_factors.rb)
* Added another program named prime_factors_sum.rb to calculate sum of all the prime factors of a given number.


atarax665:
@AdiChat  Please do review

Issue #5557, opened 2020-05-04
Print Reversed Linked List in C++


rudrakshi99:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**
There is a program to print  a Linked List in reverse order in C++

rudrakshi99:
@arnavb Sir please check.

bilalib:
Hi, can this issue be assigned to me? What file should this be implemented in?

arjungarg07:
Hey! Can i work on this issue?

Shikha291:
@rudrakshi99 I would like to work on this

PR #5558, opened 2020-05-04
Create print_reverse.cpp

rudrakshi99:
**Fixed issue**:  #5557

### DIFF
--- a/code/data_structures/src/list/singly_linked_list/operations/print_reverse/print_reverse.cpp
+++ b/code/data_structures/src/list/singly_linked_list/operations/print_reverse/print_reverse.cpp
@@ -1,0 +1,57 @@
+#include<bits/stdc++.h>
+
+using namespace std;
+
+struct node {
+    int data;
+    struct node *next;
+};
+
+/* Function to push nodes in a linked list. */
+void push(struct node **head_ref, int data) {
+    struct node *node;
+    node = (struct node*)malloc(sizeof(struct node));
+    node->data = data;
+    node->next = (*head_ref);
+    (*head_ref) = node;
+}
+ 
+/* Function to reverse the nodes in a linked list. */
+void reverse(struct node **head_ref) {
+    struct node *temp = NULL;
+    struct node *prev = NULL;
+    struct node *current = (*head_ref);
+    while(current != NULL) {
+        temp = current->next;
+        current->next = prev;
+        prev = current;
+        current = temp;
+    }
+    (*head_ref) = prev;
+}
+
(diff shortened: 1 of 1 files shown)


rudrakshi99:
@Da-kota now, you will please review it.

rudrakshi99:
Yes, by using delete operator which deallocates the memory.
like:
```c++
     main() {
     .........
         delete head;
      }

```
will I add this in my code?

Da-kota:
How many nodes would that free up?

rudrakshi99:
It will delete the entire linked link at the end.

Da-kota:
That is not true.  Just like when you would have allocated it in the old C way with ```malloc```,  freeing a pointer to node would also free up just one node.  So ```delete``` only frees one node.  So that makes for an interesting challenge.

rudrakshi99:
Delete will free the memory of that node which is pointing to the head of a linked list else I can also delete using this code:
```c++
void deleteList(Node*& head_ref)  
{  
 
Node* current = head_ref;  
Node* next;  
  
while (current !=nullptr)  
{  
    next = current->next;  
    free(current);  
    current = next;  
}  
      head_ref = nullptr;  
}  

main()
{
    ............
    deleteList(head);  
}
```

Da-kota:
- So deleting the head would create a problem:  you loose access to the list because there the pointer to the next node was stored.  You could never use or correctly delete the list anymore. 

- Do you want to code in C++ or C?  Because the source you were using from [geeks](https://www.geeksforgeeks.org/write-a-function-to-delete-a-linked-list/)  is of very poor quality.   It is written in C style instead of clean C++ style.
In good C++ , e.g., you will not use: ```NULL```,  ```malloc```/```free```  (check line 33 in last commit)

- Seems like it's almost done!  Congrats!

rudrakshi99:
ok 😀.Now will u merge it?

Da-kota:
Where's the commit?  :-)

rudrakshi99:
I have mentioned the issue #5557

Da-kota:
All nice but there seems to be to 2 commits.  Where is no 3?

rudrakshi99:
First will I add the code for deleting the node?

Da-kota:
What is your suggestion?  Merge incomplete/incorrect code?

rudrakshi99:
Please merge this one.

Da-kota:
Line 44 is not ready.

Da-kota:
Great work. Congrats, and enjoy!

rudrakshi99:
Thank you😀

PR #5623, opened 2020-05-19
Deletion of a node in a linked list(except the last node) without head pointer code in cpp

aryamasinha:
**Changes** : Added code of deletion without  a head pointer in cpp


aryamasinha:
@AdiChat could you review this?

PR #5728, opened 2020-05-30
Added README for article

Vaidehi0421:
Article on handling exceptions in java https://iq.opengenus.org/handling-exceptions-in-java/

### DIFF
--- a/code/languages/Java/Handlingexceptions/README.md
+++ b/code/languages/Java/Handlingexceptions/README.md
@@ -1,0 +1,7 @@
+#Handling exceptions in java
+
+## [Handling exceptions in java](iq.opengenus.org/handling-exceptions-in-java/)
+
+ An exception is an undesirable or unforeseen occasion, which happens during the execution of a program and disturbs the typical progression of the program's guidelines. Exception Handling is a way to deal with this possibility.
+
+This article discusses the different types of exceptions and how to deal with them.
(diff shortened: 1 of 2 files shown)


code/languages/Java/Handlingexceptions/README.md
```diff
@@ -0,0 +1,7 @@
+#Handling exceptions in java
```
iamrajiv, line 1: @Vaidehi0421 it should be `# Handling exceptions in java`.Also, why to add `Handling exceptions in java` heading two times?

Vaidehi0421: @iamrajiv No need to add the link?

iamrajiv: @Vaidehi0421 I think there is no problem if the document has a link in the heading or you can make like this i.e. adding the link to the bottom of document something like this

Article [Exceptions in Java](https://www.geeksforgeeks.org/exceptions-in-java/) discusses the different types of exceptions and how to deal with them.

iamrajiv: @Vaidehi0421 I think there is no problem if the document has a link in the heading or you can make like this i.e. adding the link to the bottom of document something like this
Article [Exceptions in Java](https://www.geeksforgeeks.org/exceptions-in-java/) discusses the different types of exceptions and how to deal with them.

PR #5777, opened 2020-05-31
Move zeroes

15aarushiGhadiya:
**Fixes issue:**

**Changes:**
Added code for moving zeroes in end of the array.

### DIFF
--- a/cosmos-master/code/unclassified/src/moving_zeroes_to_end_of_array/move_zeroes_to_end_of_array.txt
+++ b/cosmos-master/code/unclassified/src/moving_zeroes_to_end_of_array/move_zeroes_to_end_of_array.txt
@@ -1,0 +1,32 @@
+class MoveZeroes {
+    public static void moveZeroes(int[] nums) {
+        int i=0;
+        int k=0;// will keep track of non zero elements first and then zeroes elements
+        while(i<nums.length)
+        {
+            if(nums[i] !=0)
+            {
+                nums[k] = nums[i];// allowing only non zero elements
+                k++;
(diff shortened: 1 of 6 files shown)


cosmos-master/code/unclassified/src/moving_zeroes_to_end_of_array/move_zeroes_to_end_of_array.txt
```diff
@@ -0,0 +1,32 @@
+class MoveZeroes {
```
iamrajiv, line 1: @15aarushiGhadiya change the code like this
```java
class MoveZeroes {
   public static void moveZeroes(int[] nums) {
      int i = 0;
      int k = 0; // will keep track of non zero elements first and then zeroes elements
      while (i < nums.length) {
         if (nums[i] != 0) {
            nums[k] = nums[i]; // allowing only non zero elements
            k++;
         }
         i++;
      }
      // when the whole array traversed no. of non zeroes = k and no. of zeroes =
      // n-1-k, where n is length of array.
      while (k != nums.length) // putting zeroes at remaining indices.
      {
         nums[k] = 0;
         k++;
      }
   }

   public static void main(String args[]) {
      Scanner sc = new Scanner(System.in);
      int n = sc.nextInt();
      int[] nums = new int[n];
      for (int i = 0; i < n; i++) {
         nums[i] = sc.nextInt();
      }
      moveZeroes(nums);
      String stringArr = Arrays.toString(nums);
      System.out.println(stringArr);
   }
}
```

PR #5778, opened 2020-05-31
Update global_metadata.json

rudrakshi99:
Link to the article: https://iq.opengenus.org/begin-and-end-array-cpp/


AdiChat:
Looks great 😍 

positive_opengenus

PR #5839, opened 2020-06-01
added efficient 3 partition method for quick sort

KartikeySharma:
Worse than Average Case

For quick sort, we could imagine a worse than average case where we get unlucky and:
- for odd levels we choose the worst possible pivot i.e. all elements are to the left or right of the pivot
- for even levels, we choose pivots where 3/4 of the elements are on one side and 1/4 on the other side
- so roughly, after every 2 levels, our biggest subproblem is 3/4 of the size it was 2 levels before
- so now it takes 2 * log_(4/3) (n) levels until the size of the problem is 1
(Why?
n * (3/4)^x = 1, tells us how many times, x, that we have to multiply n by 3/4 to get 1
Solving for x gives us x - log_(4/3) (n), but we have to double it since we only multiply
by 3/4 on every second level)

The average case is better than this worse than the average case.

So, the average case is between the best case of log_2(n) levels and the worst than average case of 2 * log_(4/3) (n) levels
-----------------------------------------------------------------------------------------------------------
Using random pivoting we improve the expected or average time complexity to O (N log N). The Worst-Case complexity is still O ( N^2 ).

**Fixes issue:**

**Changes:**

### DIFF
--- a/code/divide_conquer/src/quick_sort/RandomizedQuickSort.cpp
+++ b/code/divide_conquer/src/quick_sort/RandomizedQuickSort.cpp
@@ -1,0 +1,61 @@
+#include <iostream>
+#include <vector>
+#include <cstdlib>
+
+using std::vector;
+using std::swap;
+
+vector<int> partition3(vector<int> &a, int l, int r) 
+{
+	int x = a[l];
+    int m1 = l;
+	int m2 = r;
+	int i = l;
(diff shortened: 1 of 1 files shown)


KartikeySharma:
I have provided a much more efficient solution for quick sort with 3 partitions. Please review it.

PR #5842, opened 2020-06-03
Prakharsharma2206 patch 2

Prakharsharma2206:
Changed the IsSafe function for a better and opttimised solution

**Fixes issue:**
The IsSafe code was way too lengthy than requiered , so shortened it.

**Changes:**
The complete code of IsSafe function was changed keeping in mind the Style Guide provided.


atarax665:
Please change the PR heading

Issue #5855, opened 2020-06-14
Add area of polygon in C#


irakli:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**
Add area of polygon in C#

nihithreddy:
can i work on this?

SharvariTatachar:
Done! https://github.com/OpenGenus/cosmos/pull/6830

PR #5856, opened 2020-06-14
Add area of polygon in C#

irakli:
**Fixes issue:**
#5855 

**Changes:**
Added area of polygon in C#


AhmedElsisy:
actually, it's for triangles only (polygon with three points).
what about polygon with n points?

Issue #5884, opened 2020-08-24
Add Garsia - Wachs Algorithm in c++


kejriwalShivam:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

Add Garsia-Wachs Algorithm in c++

kejriwalShivam:
> **This is a(n):**
> 
> * [x]  New algorithm
> * [ ]  Update to an existing algorithm
> * [ ]  Error
> * [ ]  Proposal to the Repository
> 
> **Details:**
> 
> Add Garsia-Wachs Algorithm in c++

russellE46:
I can add this

rainmaker6:
Pr number #5935 
Please review the implementation of Garsia - Wachs algorithm

PR #5935, opened 2020-10-02
Added implementation of garsia_wachs algorithm in C++

rainmaker6:
FIXED ISSUE #5884 

Added implementation of Garsia - Wachs algorithm in C++

### DIFF
--- a/garsia_wachs.cpp
+++ b/garsia_wachs.cpp
@@ -1,9 +1,10 @@
 /*
 garsia_wachs algorithm implementation in C++ 14/11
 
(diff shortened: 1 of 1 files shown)


kejriwalShivam (changes requested):
The code is not properly formatted and also the algorithm is not well explained.

danwolb02 (changes requested):
You should probably have a look at the [Cosmos C++ Style guide](https://github.com/OpenGenus/cosmos/blob/master/guides/coding_style/c%2B%2B/README.md), as this code is not adapted to it.
The two most notable aspects are:
- **Indentation**: "Indent 4 spaces at a time"
- **Naming**: "Names should be descriptive and avoid abbreviations"

PR #5943, opened 2020-10-02
Update cocktail_sort.py

entr0pie:
Cocktail sort re-written in a few lines, which makes more easy to understand
> An Hi from Brazil! ~ Caio


walidbarakat (approved):
looks fine!

Issue #5924, opened 2020-10-02
Add State Design Pattern In Java


ighosh98:
**Details:**

**Details:**

**This is a(n):**
- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**
Add State Design Pattern in Java, to Design Patterns.

ighosh98:
Can I begin work on this issue?

AdiChat:
@ighosh98 Yes, you can get started working on this and open [a pull request](https://iq.opengenus.org/create-a-pull-request-at-github/) for review 👍

ighosh98:
Please review #5945

PR #5945, opened 2020-10-03
State Pattern Standard Implementation added

ighosh98:
**Fixes issue:**

#5924 is fixed in Java

**Changes:**

State Pattern implemented in Java. In this commit I have implemented two states: Start State and STop State. This model can be extended to a multi-state solution


Devansh-Maurya (changes requested):
The implementation does not live up to the definition of the state pattern. Your `State` interface has a function named `goNext()` but it is not changing the state of the context. The Client should not need to change the state of the `Context` by itself. If you are doing this, you are losing the power of the state pattern. You should delegate the responsibility to change states to the States themselves.

Also, the `Context` should not implement the `State` interface as it is not a State. It is the object that has State.

Please refer to this resource (or any other resource that you found helpful) for more information on the State pattern: https://sourcemaking.com/design_patterns/state

Devansh-Maurya (changes requested):
1. Please keep classes in separate files. 
2. Group similar classes in the same package. 
3. Structure packages according to the UML diagram of the pattern.

ighosh98:
Thanks. I'll work on it.

Issue #5965, opened 2020-10-04
Typescript palindrome checker not working


joaopedrocampos:
**This is a(n):**
- [ ] New algorithm
- [ ] Update to an existing algorithm
- [x] Error
- [ ] Proposal to the Repository

**Details:**
The Typescript palindrome checker is not working, it's always returning true to any given word.

PR #5966, opened 2020-10-04
Fix Typescript palindrome checker

joaopedrocampos:
**Fixes issue:**
#5965

**Changes:**
Fixes broken Typescript palindrome checker by splitting the string parameter into an array before reversing it.


arnavb (approved):
Looks good! Thanks for contributing!

PR #5987, opened 2020-10-07
Added first_set_bit_using_bitmanipulation.cpp

vishalg8454:
This file adds second way of finding the right-most set bit in a number to the repository.

**Changes:**

Adds one more way of finding the right-most set bit of a number.


vishalg8454:
@AdiChat Please review my first ever PR :)

AdiChat (commented):
Add correct file extension and place within a directory

PR #6005, opened 2020-10-13
feat: add non recursive top down merge sort implementation

ShivenTripathi:
Sorts input array in non decreasing order
Uses tree like post order traversal
Template Stack and custom range pair datastructure defined

**Fixes issue:**

**Changes:**


walidbarakat (changes requested):
i think the code better to explain itself, please make sure if we need these to much comments. and add at the head of the file comment "/* Part of Cosmos by OpenGenus Foundation */“ As mentioned in [guidelines](https://github.com/OpenGenus/cosmos/wiki/contribute)

ShivenTripathi:
Hey, I am sorry I missed out on responding earlier.
I have some major academic commitments in the next 2 weeks, would resolve these comments soon after that

PR #6011, opened 2020-10-18
Create interval_tree.cpp

Sanjay0-4:
**Fixes issue:** implemention of interval tree in cpp
 

**Changes:**

### DIFF
--- a/code/languages/cpp/interval_tree.cpp
+++ b/code/languages/cpp/interval_tree.cpp
@@ -1,0 +1,131 @@
+Algo: 
+Interval overlappingIntervalSearch(root, x)
+1) If x overlaps with root's interval, return the root's interval.
+
+2) If left child of root is not empty and the max  in left child 
+is greater than x's low value, recur for left child
+
+3) Else recur for right child.
+
+
+#include <iostream> 
+using namespace std; 
+  
+// Structure to represent an interval 
+struct Interval 
+{ 
+    int low, high; 
+}; 
+  
+// Structure to represent a node in Interval Search Tree 
+struct ITNode 
+{ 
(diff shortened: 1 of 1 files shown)


code/languages/cpp/interval_tree.cpp
```diff
@@ -0,0 +1,131 @@
+Algo: 
+Interval overlappingIntervalSearch(root, x)
+1) If x overlaps with root's interval, return the root's interval.
+
+2) If left child of root is not empty and the max  in left child 
+is greater than x's low value, recur for left child
+
+3) Else recur for right child.
+
```
walidbarakat, line 9: shouldn't be commented?

walidbarakat (changes requested):
i see you write a lot of comments, but maybe, if you interested, can just try to make more expressive variables to make your code readable with no need to comments.

code/languages/cpp/interval_tree.cpp
Sanjay0-4: done

walidbarakat:
looks ok, just add the slogan at the beginning of the file, sorry but it's required according to the contribution guide.

code/languages/cpp/interval_tree.cpp
```diff
@@ -0,0 +1,101 @@
...
```
walidbarakat, line 1: just add the comment as mentioned in [contribution guide](https://github.com/OpenGenus/cosmos/wiki/contribute)

walidbarakat (changes requested):

AdiChat (commented):
In addition to the changes proposed by @walidbarakat, we need to move the code to the correct location/ folder.

Issue #6016, opened 2020-10-20
Add code for remove kth node from end in linked list


shashank-100:
**This is a(n):**

- [ ] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

snehaa1989:
@shashank-100  please assign this to me

Shikha291:
@shashank-100 kindly assign this to me

shreya-shah-28:
i want to work on this issue please assign this to me

PR #6017, opened 2020-10-20
Added Code remove_kth_node_from_end.cpp

shashank-100:
**Fixes issue:**
#6016 

**Changes:**


bhavanasa (changes requested):
Where is the actual node being deleted?
How about the case of n==k?

bhavanasa (commented):
Instead of calculating the length of the list, can you use fast and slow pointers and solve?

PR #6032, opened 2020-10-25
Warmups

cmankos:
**Fixes issue:**

**Changes:**

Rest of Rosalind's introductory problems


code/online_challenges/src/rosalind/python_village/conditions_and_loops.py
```diff
@@ -0,0 +1,12 @@
+# Sum all the odd numbers between a and b, inclusive
+def conditionsAndLoops(a, b):
+
```
walidbarakat, line 3: redundant line.

code/online_challenges/src/rosalind/python_village/conditions_and_loops.py
```diff
@@ -0,0 +1,12 @@
+# Sum all the odd numbers between a and b, inclusive
+def conditionsAndLoops(a, b):
```
walidbarakat, line 2: you could use python built-in loop feature **for i in range(a, b+1)** instead of declaring 'i' as variable and avoiding +1 operation for every even number.

code/online_challenges/src/rosalind/python_village/working_with_files.py
```diff
@@ -0,0 +1,11 @@
+# Print even number lines
+def workingWithFiles():
+    rosalind = open("C:/Users/Chris/Downloads/rosalind_ini5.txt")
```
walidbarakat, line 3: you can't put a constant path here, it works only on your machine.

code/mathematical_algorithms/src/collatz_conjecture_sequence/collatz_conjecture_sequence.py
```diff
@@ -0,0 +1,11 @@
+def collatzSequence(n):
+    while n != 1:
+        if n%2 == 0:
+            n = int(n/2)
+        else:
+            n = int((3 * n) + 1)
```
walidbarakat, line 6: you could cast to integer in the beginning of the function, if you want to make sure that **n** is integer.

walidbarakat (changes requested):
please avoid unnecessary blank lines in the code, and add the line \**Part of Cosmos by OpenGenus Foundation**/ in head of each file.

PR #6033, opened 2020-10-25
Algorithmic heights

cmankos:
**Fixes issue:**

**Changes:**

Algorithmic heights (rosalind.info) memoized fibonacci

### DIFF
--- a/code/mathematical_algorithms/src/collatz_conjecture_sequence/collatz_conjecture_sequence.py
+++ b/code/mathematical_algorithms/src/collatz_conjecture_sequence/collatz_conjecture_sequence.py
@@ -1,0 +1,11 @@
+def collatzSequence(n):
+    while n != 1:
+        if n%2 == 0:
+            n = int(n/2)
+        else:
+            n = int((3 * n) + 1)
+        print(n)
+
+
+n = int(input())
+collatzSequence(n)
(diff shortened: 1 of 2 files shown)


code/mathematical_algorithms/src/collatz_conjecture_sequence/collatz_conjecture_sequence.py
```diff
@@ -0,0 +1,11 @@
+def collatzSequence(n):
+    while n != 1:
+        if n%2 == 0:
+            n = int(n/2)
+        else:
+            n = int((3 * n) + 1)
+        print(n)
+
```
walidbarakat, line 8: unnecessary blank line.

code/online_challenges/src/rosalind/algorithmic_heights/fibonacci.py
```diff
@@ -0,0 +1,21 @@
+# Find the nth Fibonacci number
+def fib(n):
+
+    fibDict = dict()
```
walidbarakat, line 4: why we need to use dict() here?
- you can find the number using recursive technique as you suggested but directly with no need to load dictionary and make a sub-function.
- maybe you need to think about how to handle a negative input. but note that [fibonacci sequence](https://en.wikipedia.org/wiki/Fibonacci_number) can be extended to include the negative index "negafibonacci".
- and i think in python you don't need to use new lines that much, it's readable by syntax.

and a reminder, please make two different PRs if you commit in two different dirs.

walidbarakat (changes requested):
it's good to take a look at "negafibonacci" on wikipedia.

PR #6020, opened 2020-10-22
Added betweenness centrality approximation algorithm in python

AlexanderDavid:
**Fixes issue:**
#6019

**Changes:**
Add code for betweenness centrality

Issue #6019, opened 2020-10-22
Add Betweenness Centrality Algorithm in Python


AlexanderDavid:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

The betweenness centrality of a graph determines singular node importance based upon the geodesic paths that exist within the graph. This specific implementation allows for a k-node approximation that speeds up the runtime at the cost of accuracy of the metric. This can be used to determine important nodes within a graph.

PR #6035, opened 2020-10-26
Added betweenness centrality approximation algorithm in python

AlexanderDavid:
**Fixes issue:**

#6020, #6019

**Changes:**

Added betweenness centrality approximation graph algorithm


AdiChat (commented):
We need to remove the executable file.

PR #6037, opened 2020-10-27
Create ValidateIp.java

hvsalesforce:
**Fixes issue:**

Added ip validator written in Java

**Changes:**

Added ip validator for java, was missing.

### DIFF
--- a/code/networking/src/validate_ip/ValidateIp.java
+++ b/code/networking/src/validate_ip/ValidateIp.java
@@ -1,0 +1,35 @@
+import java.util.Arrays;
+import java.util.regex.Pattern;
+
+public class ValidateIp {
+    public static void main(String[] args) {
+        assert (isValid("172.16.254.1"));
+        assert (!isValid("172.16.254.01"));
+        assert (isValid("2001:0db8:85a3:0000:0000:8a2e:0370:7334"));
+        assert (isValid("2001:db8:85a3:0:0:8A2E:0370:7334"));
+        assert (!isValid("02001:0db8:85a3:0000:0000:8a2e:0370:7334"));
+    }
+
+    private static boolean isValid(String addr) {
+        if (addr.contains(".")) {
+            return isValidv4(addr);
+        }
+        if (addr.contains(":")) {
+            return isValidv6(addr);
+        }
+        return false;
+    }
+
+    private static boolean isValidv4(String addr) {
+        Pattern pattern = Pattern.compile("(?:(?:(?:[1-9][0-9]{0,2})|0)\\.){3}(?:(?:[1-9][0-9]{0,2})|0)");
+        if (!pattern.matcher(addr).matches()) {
+            return false;
+        }
+        return Arrays.stream(addr.split("\\.")).noneMatch(s -> Integer.parseInt(s) > 255);
+    }
+
+    private static boolean isValidv6(String addr) {
+        Pattern pattern = Pattern.compile("(?:(?:(?:[0-9a-fA-F]{1,4})):){7}(?:(?:[0-9a-fA-F]{1,4}))");
+        return pattern.matcher(addr).matches();
+    }
+}
--- a/code/networking/src/validate_ip/ValidateIp.java
+++ b/code/networking/src/validate_ip/ValidateIp.java
@@ -1,13 +1,20 @@
+/* Part of Cosmos by OpenGenus Foundation */
+
 import java.util.Arrays;
 import java.util.regex.Pattern;
 
 public class ValidateIp {
+    private static final Pattern ipv4pattern =
+        Pattern.compile("(?:(?:(?:[1-9][0-9]{0,2})|0)\\.){3}(?:(?:[1-9][0-9]{0,2})|0)");
+    private static final Pattern ipv6pattern = Pattern.compile("(?:(?:(?:[0-9a-fA-F]{1,4})):){7}(?:(?:[0-9a-fA-F]{1,4}))");
+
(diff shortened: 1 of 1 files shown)


walidbarakat (changes requested):
as mentioned in wiki : 
- naming convention for files should be in lower case, so file name should be validate_ip.java or - for more accuracy - validate_ip_address.java.
- and as well the line "/* Part of Cosmos by OpenGenus Foundation */" it can be the first line of the file.

code/networking/src/validate_ip/ValidateIp.java
```diff
@@ -0,0 +1,35 @@
+import java.util.Arrays;
+import java.util.regex.Pattern;
+
+public class ValidateIp {
+    public static void main(String[] args) {
+        assert (isValid("172.16.254.1"));
+        assert (!isValid("172.16.254.01"));
+        assert (isValid("2001:0db8:85a3:0000:0000:8a2e:0370:7334"));
+        assert (isValid("2001:db8:85a3:0:0:8A2E:0370:7334"));
+        assert (!isValid("02001:0db8:85a3:0000:0000:8a2e:0370:7334"));
```
walidbarakat, line 10: conventionally, it's better to make a separate testing file.

code/networking/src/validate_ip/ValidateIp.java
```diff
@@ -0,0 +1,35 @@
...
+            return isValidv4(addr);
+        }
+        if (addr.contains(":")) {
+            return isValidv6(addr);
+        }
+        return false;
+    }
+
+    private static boolean isValidv4(String addr) {
+        Pattern pattern = Pattern.compile("(?:(?:(?:[1-9][0-9]{0,2})|0)\\.){3}(?:(?:[1-9][0-9]{0,2})|0)");
```
walidbarakat, line 24: i tested this pattern against some regex test website, it seems something wrong with it, and i see we need to handle the case with zero shortcut ipv4 like : 127.1 or 127.0.1, i may suggest using something like this **(?:(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\.){1,3}(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])**

code/networking/src/validate_ip/ValidateIp.java
```diff
@@ -0,0 +1,35 @@
...
+        if (!pattern.matcher(addr).matches()) {
+            return false;
+        }
+        return Arrays.stream(addr.split("\\.")).noneMatch(s -> Integer.parseInt(s) > 255);
+    }
+
+    private static boolean isValidv6(String addr) {
+        Pattern pattern = Pattern.compile("(?:(?:(?:[0-9a-fA-F]{1,4})):){7}(?:(?:[0-9a-fA-F]{1,4}))");
```
walidbarakat, line 32: this pattern looks fine with the examples you get, but as mentioned above isn't handling the case with trimmed zeroes, in ipv6 case it'll be two colons ::, i found this [question](https://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses) on stackoverflow is really useful.

walidbarakat (changes requested):
i have two small questions:
- wouldn't be better if we avoid re-compiling the pattern each time we call the validation function?
- wouldn't be better as well to declare the two patterns as constants with expressive variable names?

code/networking/src/validate_ip/ValidateIp.java
hvsalesforce: Agreed. This is hardly a unit test though, so creating a test file seemed unnecessary.
https://github.com/hvsalesforce/cosmos/blob/patch-1/code/networking/src/validate_ip/validate_ip.c Followed in the assertion footsteps of the C implementation.

code/networking/src/validate_ip/ValidateIp.java
hvsalesforce: Are 127.1 and 127.0.1 legal?
If so, we might need to expand the readme too.

code/networking/src/validate_ip/ValidateIp.java
hvsalesforce: From the README.md,
> However, we don't replace a consecutive group of zero value with a single empty group using two consecutive colons (::) to pursue simplicity. For example, 2001:0db8:85a3::8A2E:0370:7334 is an invalid IPv6 address.
I think it's being handled.

code/networking/src/validate_ip/ValidateIp.java
walidbarakat: yes, they are, i'm not fully aware of the nature of this project, but i think if want to make it a wide used library i guess i agree with you to expand the readme and handle such cases.

code/networking/src/validate_ip/ValidateIp.java
walidbarakat: maybe, i need to check it again, but please don't forget about the file name and check the patterns if they working fine.

hvsalesforce: The patterns are working for the couple of examples in the main().
Java code needs the class name to be the same as the file name, and I've never once seen a class name in snakecase.
Still, I can change it for the sake of consistency.

hvsalesforce:
I agree reg having the pattern compile only once, that was an oversight on my part

walidbarakat (approved):

AdiChat (approved):

Issue #6049, opened 2020-11-10
Sleep Sort C++ Implementation


sanchnat:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [x] Proposal to the Repository

**Details:**

[Here](https://www.geeksforgeeks.org/sleep-sort-king-laziness-sorting-sleeping/) is a link to a description about sleep sort.

sanchnat:
Hello again! I would also like to be assigned this issue.

walidbarakat:
> Hello again! I would also like to be assigned this issue.

i think you can just submit the PR for this issue, unfortunately maintainers of this repo aren't that active.

PR #6053, opened 2020-12-11
Added Sleep Sort in C++

StutiV:
**Fixes issue:**
#6049 

**Changes:**
Added a new algo for Sleep Sort

### DIFF
--- a/code/sorting/src/sleep_sort/sleep_sort.cpp
+++ b/code/sorting/src/sleep_sort/sleep_sort.cpp
@@ -1,29 +1,44 @@
-/* Part of Cosmos by OpenGenus Foundation */
-#include <iostream>
-#include <chrono>
-#include <vector>
-#include <thread>
+// C implementation of Sleep Sort 
+#include <stdio.h> 
+#include <windows.h> 
+#include <process.h> 
 
-using namespace std;
+void routine(void *a) 
+{ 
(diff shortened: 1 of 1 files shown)


code/sorting/src/sleep_sort/sleep_sort.cpp
```diff
@@ -1,29 +1,44 @@
-/* Part of Cosmos by OpenGenus Foundation */
```
walidbarakat, line 1: why you deleted this line?

code/sorting/src/sleep_sort/sleep_sort.cpp
```diff
@@ -1,29 +1,44 @@
...
+	// An array of threads, one for each of the elements 
+	// in the input array 
+	HANDLE threads[n]; 
+
+	// Create the threads for each of the input array elements 
+	for (int i = 0; i < n; i++) 
+		threads[i] = (HANDLE)_beginthread(&routine, 0, &arr[i]); 
+
+	// Process these threads 
+	WaitForMultipleObjects(n, threads, TRUE, INFINITE);
```
walidbarakat, line 30: this a windows API, so what if the code was compiled for another platform?

StutiV: My bad, I'll fix these issues.

PR #6062, opened 2021-03-12
feat: Add Backtracking Minimax implementation in C++

Panquesito7:
The file was taken and modified from https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/backtracking/minimax.cpp.
I improved that algorithm previously (see https://github.com/TheAlgorithms/C-Plus-Plus/pull/916).

**Changes:**
- Adds [Minimax](https://en.wikipedia.org/wiki/Minimax) algorithm implementation in Backtracking.


walidbarakat (approved):
seems fine!

Issue #4957, opened 2020-03-30
Explanation file of implemented code 


hardikhere:
**This is a(n):**

- [ ] New algorithm
- [x] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

We can make a file may be READ.md file in every algorithm.
reading code without explanation is difficult for learners so I am proposing an explanation of implemented programs using gifs, text, or any pictorial method. Reading merely code doesn't help much but explanation does help.
I would like to work on it.

hardikhere:
Please assign me if you want an explanation part of this library. Also, suggest in which form..txt or read.md in every algorithm.

hardikhere:
if anyone wants to contribute like this https://github.com/hardikhere/cosmos/blob/master/code/backtracking/src/knight_tour/Knight_tour_algo.md

Wittty-Panda:
@hardikhere is this issue still open? if yes, can you please assign it to me?

kittusinghkumar:
Hey, I would like to contribute for the same. Kindly assign me.

ergovind08:
kindly assign me

Giorgio-Matessi:
I would love an assignment.

batmandude101:
I would like to work on this

rashmip2102:
Please assign this to me

Issue #4721, opened 2020-03-24
PSA: Plagiarism in Pull Requests


arnavb:
As many of you have noticed, I have come back to maintain this repository after a few months.

While I am reviewing pull requests, I've been noticing many examples of people (I won't say names) plagiarising code from other sources, such as other algorithms repositories or websites.

Obviously, this behavior is intolerable. Copying other's work is a form of disrespect in my eyes and as we can see in our [contributing guidelines](https://github.com/OpenGenus/cosmos/blob/master/.github/CONTRIBUTING.md) and [code of conduct](https://github.com/OpenGenus/cosmos/blob/master/.github/CODE_OF_CONDUCT.md) disrespect to others in this community is unacceptable. As a community built on trust, something like plagiarism makes me lose all trust in you as a contributor. I'm sure we all want this repository of information to grow. 

So all in all: _Please do not plagiarise_. I can't check every website and repository to see if it matches your code, so this will be a system built on honor. If I do see plagiarism, do not expect me to readily merge your pull request (as of now, anything in past is in the past).

The reason I've made this issue is so that it gets a lot of visibility from our contributors (hopefully). 

I would appreciate any comments below as to what action should be taken in order to combat plagiarism.

@OpenGenus/cosmos-maintainers @AdiChat @iendeavor I'd like to know your thoughts on this as well. Perhaps we can add a modification to the code of conduct that warns people against copying other's code?

parjanyaacoder:
Well I think contributors can take only that code from a particular repository which they have contributed in that repository.

arnavb:
@parjanyaacoder You are right, if the code copied from another repository is written by the same person then it should be allowed in.

Feliren88:
What if the contributor provides reference? I think as long as there is citation, it should be fine right? As long as the contributor doesn't claim other people's work as him/her

PR #6064, opened 2021-05-07
Add docs

siddheshJungade:
**Fixes issue:**

#4957 

**Changes:**

add explanation of skip list


arnavb:
Please do not copy explanations from other websites. Write your own instead. #4721 
https://www.javatpoint.com/skip-list-in-data-structure

PR #6096, opened 2021-10-02
Create README.md

Eureka-Viraj:
Here is the README file for better conceptual  understanding of the Binary Search.
Added everything about binary search and it's algorithm!
**Fixes issue:**

**Changes:**


AdiChat (changes requested):
The text is duplicate / available in external resource. You can add you own work only.

Issue #6026, opened 2020-10-23
Add Gregpry Leibniz Series approximation of Pi


AlexanderDavid:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

Add the Gregpry Leibniz Series approximation of Pi. This is the powerseries of arctan specialized to x=1 and approximates the numerical value of pi when given enough iterations.

DeepakReddyVelagala:
Hi, can I work on this and is C++ allowed

rktejesh:
Can I further contribute to your code? I also worked on this issue but didnt realise there was already a pr.

PR #6098, opened 2021-10-02
Added Gregory Leibniz Approximation Series of pi

rktejesh:
**Fixes Issue:**
#6026

**Changes:**
Added Gregory Leibniz Approximation Series of pi under mathematical algorithms for c++ and python.


walidbarakat (changes requested):
can you please divide these files to four PR, each file for a PR?

PR #6104, opened 2021-10-02
 added Minimum Skips to Arrive at Meeting On Time in java and python language

rudraneel18:
…lang

**Fixes issue:**

**Changes:**


AdiChat (commented):
You need to add a main() function so that your code is runnable. You shall provide a sample input as well.

Issue #6076, opened 2021-09-30
Minimum Skips to Arrive at Meeting On Time (DP)


Jahnvi13:
**This is a:**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**
This Dynamic Programming problem is a common interview question and while implementing, it uses concept of precision errors.
Can I work on it in C++

Jahnvi13:
@AdiChat Can you please add hacktoberfest label to this issue? This is the first time I am participating and I want to make a meaningful contribution. Thanks.

AdiChat:
Sure, I have added Hacktoberfest label 🙂

rudraneel18:
Is anyone assigned yet for this task? If not then @AdiChat could you please assign this task to me?

Jahnvi13:
Ya I am doing in cpp. You can take some other language

rudraneel18:
Ok then I'll be working on Python and Java

rudraneel18:
@AdiChat can you assign both of us this task ?

rudraneel18:
#6104 added Minimum Skips to Arrive at Meeting On Time in java and python

rudraneel18:
@AdiChat please review #6107

rudraneel18:
@AdiChat please review #6124

PR #6107, opened 2021-10-02
minimum skips to arrive at meeting on time(DP) solution added in java and python languages

rudraneel18:
**Fixes issue:**
#6076 

**Changes:**
Added the minimum skip to arrive at the meeting on time problem DP solution in java and python languages


AdiChat (commented):
Add the code to call the respective functions with sample input.

rudraneel18:
added it in #6124 please review it

PR #6111, opened 2021-10-02
changes made in explainBlockchain.txt file

brat-blip:
**Fixes issue:**
**explain about the Blockchain**

**Changes:**
add brief intro as well as  add the mechanism as how it works


AdiChat (commented):
Text should be written by you originally and not copied from external sources.

brat-blip:
@AdiChat  I will work on that i didn't know exactly what to do as i am beginner so sorry for inconvenience

joaopedrocampos:
Also prefer for .md instead of .txt files, but we already have an `explainBlockchain.md`, so it would be better, if you see any point of improvement on it, to change there 🙏

PR #6112, opened 2021-10-02
Added a new function

thecoderash:
Added a function to randomly create a linked list based on the input of the user on the number of elements required.

**The function random_list creates a new list for the user to test the other functions on the new list faster.**

random_list_output

The code prompts the user to enter the number of elements required in the new randomly generated list.
Uses srand() and rand() to create random elements.

The function looks like:
image
And the other changes that were made to the code were in the switch case where a new case of our function was made.

### DIFF
--- a/code/data_structures/src/list/circular_linked_list/circular_linked_list.c
+++ b/code/data_structures/src/list/circular_linked_list/circular_linked_list.c
@@ -1,5 +1,7 @@
-#include<stdio.h>
-#include<stdlib.h>
+#include <stdio.h>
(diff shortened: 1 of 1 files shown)


AdiChat (commented):
You need to mention the issues in existing code which you are fixing.

thecoderash:
I'm trying to update the code by adding a new function that generates a sample input so that the user can work on more efficiently.
The newly generated list can also be used to find bugs in the code when there are future updates for this particular code.

Issue #6146, opened 2021-10-04
Count Sort, First Missing Positive


parvsharmaa:
**This is a(n):**

- [x] New algorithm
- [x] Proposal to the Repository

**Details:**

I would like to add a Count Sort algorithm for First Missing Positive Integer which will include Problem Statement along with the examples, Constraints and Complexity Analysis with the Optimized Solution.

#hacktoberfest2021

parvsharmaa:
@AdiChat can i proceed with it ?

AdiChat:
@parvsharmaa Sure, you may work on it.

parvsharmaa:
@AdiChat please review pr #6156

PR #6156, opened 2021-10-05
Added First Missing Positive Integer Algorithm

parvsharmaa:
**Changes:**

Added Split Array Largest Sum Algorithm in the category Counting Sort which includes Problem Statement, Examples, Constraints along with the Approach and Solution .

Issue #6146


AdiChat (commented):
You need to place your code in the correct directory.

parvsharmaa:
@AdiChat please tell the correct directory for this pr.

PR #6166, opened 2021-10-06
Create fibonacci.c

shelly-chouhan:
**Fixes issue:**

**Changes:**

### DIFF
--- a/code/dynamic_programming/src/fibonacci/fibonacci.c
+++ b/code/dynamic_programming/src/fibonacci/fibonacci.c
@@ -1,0 +1,25 @@
+#include<stdio.h>
+ 
+int fib(int n)
+{
+  int f[n+2];   // 1 extra to handle case, n = 0
+  int i;
+ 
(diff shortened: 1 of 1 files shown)


code/dynamic_programming/src/fibonacci/fibonacci.c
```diff
@@ -0,0 +1,25 @@
+#include<stdio.h>
+ 
+int fib(int n)
+{
+  int f[n+2];   // 1 extra to handle case, n = 0
```
walidbarakat, line 5: why this needed? you will have a reduntant 4 bytes every time **n** value isn't 0 or 1.

so i think to make your algorithm more memroy effecient, you can just handle the case when **n** is less than 2 by a simple if condition which returns **n** if it's true. and to have a good **fibonacci** algorithm you can as well check -first of all- if **n** is less than 0, and in this case you can just return an error integer, -1 for example.

PR #6169, opened 2021-10-06
Added the code for Redundant connection

Hrit20:
**Fixes issue:**

**Changes:**


AdiChat (commented):
You need to place your code in the correct directory.

PR #6232, opened 2021-10-16
Add Number of Dice Rolls With Target Sum.java

lyfer233:
**Fixes issue:**

Implementing "Number of Dice Rolls With Target Sum" with Java, its algorithm is dp.

**Changes:**

added the `NumberOfDiceRollsWithTargetSum.java`


lyfer233:
I don't similar to use git. So if I make some mistakes, I didn't mean to mess them up.

Issue #6212, opened 2021-10-12
Maximum Subarray


Op-panda:
**This is a:**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

Given an integer array nums, find the contiguous subarray (containing at least one number) which has the largest sum and return its sum.

A subarray is a contiguous part of an array.

DavidMaged171:
Could you assign me to this problem I can solve it using c++

PR #6242, opened 2021-10-18
Maximum Subarray (#6212)

Op-panda:
**Fixes issue:**
#6212 

**Changes:**
Added algorithm to find maximum subarray sum


Op-panda:
so who does merge @walidbarakat ?

PR #6312, opened 2021-10-28
Added Exponential_Search.cpp

visha09:
Added the program for the Exponential Search in c++

### DIFF
--- a/code/languages/cpp/Exponential_Search.cpp
+++ b/code/languages/cpp/Exponential_Search.cpp
@@ -1,0 +1,66 @@
+// C++ program to find an element x in a
+// sorted array using Exponential search.
+#include <bits/stdc++.h>
+using namespace std;
+ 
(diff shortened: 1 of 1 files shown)


code/languages/cpp/Exponential_Search.cpp
```diff
@@ -0,0 +1,66 @@
+// C++ program to find an element x in a
+// sorted array using Exponential search.
+#include <bits/stdc++.h>
+using namespace std;
+ 
+int binarySearch(int arr[], int, int, int);
```
walidbarakat, line 6: why?

walidbarakat (changes requested):
please don't use comments to make your code explained, if you think your code isn't self-expressing one, so try to think about how to make it clean.

AdiChat (commented):
Code for Exponential Search should be in: /code/search/src/exponential_search

PR #6315, opened 2021-10-29
Update test_sample.cpp

Knight-coder:
**Fixes issue:**

**Changes:**


AdiChat (commented):
You may need to explain idea behind your change.

PR #6320, opened 2021-10-30
adding 2sum

Vinayakjaiswal07:
**Fixes issue:**
adding 2sum in python

**Changes:**

adding 2sum in python


AdiChat (commented):
You need to add your code in the file 2sum.py in correct directory and not in .tmp file

PR #6331, opened 2021-10-31
Lorenz Attractor

AYANCHAKRABORTY0712:
**Fixes issue:**

**Changes:**


walidbarakat (commented):
please consider not using too long comments!

AdiChat (commented):
This is the correct directory: cosmos/code/mathematical_algorithms/src/lorenz_attractor/

It has a Python implementation.

PR #6337, opened 2021-10-31
Added Gray-code bit-manipulation code

tanishchugh01:
<a href="https://leetcode.com/problems/gray-code/">Gray-code</a> is a medium-level question with 54.2% acceptance based on topic bit-maipulation.

The solution is submitted at 4ms runtime.

### DIFF
--- a/code/bit_manipulation/src/gray_code/grayCode.cpp
+++ b/code/bit_manipulation/src/gray_code/grayCode.cpp
@@ -1,0 +1,21 @@
+
+
+class Solution {
+public:
+    vector<int> grayCode(int n) {
+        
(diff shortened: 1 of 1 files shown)


code/bit_manipulation/src/gray_code/grayCode.cpp
```diff
@@ -0,0 +1,21 @@
+
```
AdiChat, line 1: Make sure your code run as expected.

tanishchugh01: image
The code does run as expected in 4ms runtime.

PR #6340, opened 2021-11-01
Lorenz attractor in python added

AYANCHAKRABORTY0712:
**Fixes issue:**

**Changes:**


walidbarakat (changes requested):
can you please consider if using all of these comments necessary?

AdiChat (approved):

PR #6342, opened 2021-11-04
Adding Two more Sorts

Anshir08:
**Fixes issue:**

**Changes:**


walidbarakat (changes requested):
Hi, can you pleasd consider submitting each file in a different PR? and please consider as well not adding not needed comments.

PR #6343, opened 2021-11-07
Dp

kirtisingh3008:
**Fixes issue:**

I have added Count all subsequences having product less than K  to dynamic problem folder in cpp language.
**Changes:**

### DIFF
--- a/code/dynamic_programming/src/Count_Subsequence_Having_Product_Less_Than_K/Solution.cpp
+++ b/code/dynamic_programming/src/Count_Subsequence_Having_Product_Less_Than_K/Solution.cpp
@@ -1,0 +1,67 @@
+//Given a positive array, find the number of subsequences having product smaller than K.
+// LINK - https://www.geeksforgeeks.org/count-subsequences-product-less-k/#
+#include<bits/stdc++.h>
+using namespace std;
+int dp[10005][10005];
+
+int helper(vector<int> a, int n, int k , int product)
+{
+    // base case
+    if(n==0)
+    {
+        return 0;
+    }
+    if(product > k)
+    {
+        return 0;
+    }
+    if(dp[n][product] !=-1)
+    {
+        return dp[n][product];
+    }
+
+    int include = product * a[n-1];
+
+    int exclude =product;
+    int count =0;
+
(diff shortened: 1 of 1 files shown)


code/dynamic_programming/src/Count_Subsequence_Having_Product_Less_Than_K/Solution.cpp
```diff
@@ -0,0 +1,67 @@
+//Given a positive array, find the number of subsequences having product smaller than K.
```
walidbarakat, line 1: not needed

code/dynamic_programming/src/Count_Subsequence_Having_Product_Less_Than_K/Solution.cpp
```diff
@@ -0,0 +1,67 @@
...
+// LINK - https://www.geeksforgeeks.org/count-subsequences-product-less-k/#
+#include<bits/stdc++.h>
```
walidbarakat, line 3: not a good practice to use such header, can you please use only the headers needed?

code/dynamic_programming/src/Count_Subsequence_Having_Product_Less_Than_K/Solution.cpp
```diff
@@ -0,0 +1,67 @@
...
+        a.push_back(x);
+    }
+
+    int k;
+    cout<<"Enter the value of k"<<endl;
+    cin>>k;
+
+    cout<<solve(a,n,k)<<endl;
+    return 0;
+}
```
walidbarakat, line 67: let the last line be a blank one always

walidbarakat (changes requested):
Hi, can you please consider the comments and as well please give a meaningful name to the PR

code/dynamic_programming/src/Count_Subsequence_Having_Product_Less_Than_K/Solution.cpp
kirtisingh3008: okay sure , sir will make the changes .

PR #6345, opened 2021-11-15
added Kadane implementation for C++

IsomerX:
Add the Kadane  algorithm implementation for C++.

### DIFF
--- a/code/data_structures/src/maxSubArray(KadaneAlgorithm)/KadaneAlgorithm.cpp
+++ b/code/data_structures/src/maxSubArray(KadaneAlgorithm)/KadaneAlgorithm.cpp
@@ -1,4 +1,5 @@
-#include <bits/stdc++.h>
+#include <iostream>
+#include <vector>
 
 using namespace std;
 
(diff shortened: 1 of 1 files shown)


code/data_structures/src/maxSubArray(KadaneAlgorithm)/KadaneAlgorithm.cpp
```diff
@@ -0,0 +1,30 @@
+#include <bits/stdc++.h>
```
walidbarakat, line 1: it's a bad practice to use such a header, it can be useful only in comptetive programming

walidbarakat (changes requested):

code/data_structures/src/maxSubArray(KadaneAlgorithm)/KadaneAlgorithm.cpp
IsomerX: I understand, I have changed the header files into the ones required.

walidbarakat (approved):
looks fine!

AdiChat (approved):

PR #6362, opened 2021-12-27
Create optimized segment tree implementation

namanjain020:
**Fixes issue:**

**Changes:**
Adding a file for the code of segment tree which always require only 2N space always whether the size of the array is the power of 2 or not. The code has a bottom-up approach for sum queries and updating values so that we can achieve it using iteration and the analysis is much easier.


SpoonBuoy (approved):
The code is correct and the comments describing the code are very explanatory. Passed all my test cases locally.

AdiChat (approved):

Issue #3706, opened 2018-10-02
Add code for Gaussian Naive Bayes algorithm


AdiChat:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

The task is to add code for **Gaussian Naive Bayes algorithm** in any language of your choice (`C`, `C++`, `Java`, `Python`, `Swift`, `JavaScript`, `GoLang`, `Matlab` and others. 

The path of this code will be: `code/artificial_intelligence/src/gaussian_naive_bayes/`

:speaker: You can, also, write about it at OpenGenus IQ and become a part of **OpenGenus Publisher Team**. [See this](https://discourse.opengenus.org/t/become-a-writer-at-opengenus-iq/843) and let us know your interest :smile: 

If you face any problem, let us know. We are here to help. 

[Join our community at our Discourse forum](https://discourse.opengenus.org/) and stay connected. To get your pull request reviewed instantly, you can directly message us at our forum. 

Enjoy :tada:  :smile: 

To take a look at other possible contributions at OpenGenus, [check out our guides](https://discourse.opengenus.org/c/guide).

Check us out at: [IQ](https://iq.opengenus.org/), [Discuss](https://discourse.opengenus.org/), [Twitter](https://twitter.com/OpenGenus),

cyberdrk:
Hi @AdiChat, Please assign this issue to me! (python implementation from scratch)

GarvTambi:
hi, @AdiChat @Lechenco @souvik3333 @vikram2000b @arnavb I am interested to add code in Gaussian Naive Bayes, So, should I start/pull my code.

Nivig1310:
Hello, is there anything for me to contribute in Python?

Rsakshi11:
hey,, @AdiChat @Lechenco @souvik3333 @vikram2000b @arnavb I am interested to add code in Gaussian Naive Bayes

nrhitik:
I would like to work on this issue. Could you please assign me this issue.

tadeolozano:
Hi, i would like to contribute on this

PR #6370, opened 2022-01-27
Adding Gaussian Naive Bayes Algorithm in C++

m-lin1:
**Fixes issue:**

Addresses Issue #3706, adding in a C++ version of the Gaussian Naive Bayes algorithm

**Changes:**

Addition of Gaussian Naive Bayes algorithm for C++ and an additional test file, inputs.csv, that was used in the algorithm. The code should be a basis for the algorithm, but it has some limitations i.e. what kind of input it can take. I believe it should still have the essentials of the algorithm though. The code only outputs the accuracy of the predictions made using Gaussian Naive Bayes.


aniketmdinde:
Can you please assign me this issue. I would like to work on it.

PR #6423, opened 2022-10-01
Create Rat_in_a_maze_problem.cpp

usernamehacktoberfest:
**Fixes issue:**

**Changes:**


AdiChat (changes requested):
C++ code for this problem is available at: code/backtracking/src/rat_in_a_maze/rat_in_a_maze.cpp

You may submit for another problem or make improvements over the existing implementation.

PR #6431, opened 2022-10-01
Pratyaksha code

Pratyaksha462:
**Fixes issue:**

**Changes:**


Pratyaksha462 (commented):
This is the folder that is created to give the full info on the types of STLs in the C++ Language.

PR #6444, opened 2022-10-02
SVM in C++

7Arms:
**Fixes issue:**

**Changes:**

I attempted to create a basis for a SVM class in C++.
Please let me know if I can make any edits that would make my code fit better with the entire codebase, or if there is any other changes I should make.


AdiChat (commented):
Remove the binary and setting files.

Keep svm.cpp only.

PR #6469, opened 2022-10-03
Create topological_sort.js

AlphaVS-76:
**Changes:**

- Added JavaScript version of topological sort

**Notes:**
I would be glad if you could label this as hacktoberfest2022 accepted :)


AlphaVS-76:
Resolves Issue #6470

PR #6519, opened 2022-10-06
Dinic Algorithm

prashantkr29:
Added Dinic algorithm implementation in python. under the  Dinic maximum flow folder in Graph Algorithm

 Initialize residual graph G as given graph.
 Do BFS of G to construct a level graph (or assign levels to vertices) and also check if more flow is possible

Time Complexity - O(EV^2)
space Complexity - O(E+V)


prashantkr29:
@AdiChat  This is the implementation of Dinic Algorithm for maximum flow. Please have a Look at the PR

PR #6531, opened 2022-10-07
added Maximum size rectangle binary sub-matrix with all 1s DP problem

sourav-naskar:
added Maximum size rectangle binary sub-matrix with all 1s problem in code\dynamic_programming\src\Maximum_size_rectangle_binary_sub-matrix_with_all_1s folder. I Would be very glad if I get "hacktoberfest-accepted" label.

Thank you,
Sourav


sourav-naskar:
I Would be very glad if I get "hacktoberfest-accepted" label @AdiChat

Issue #6583, opened 2022-10-10
Adding Unique Paths problem | Dynamic Programming


Princeton21:
**This is a(n):**

- [ ] New algorithm
- [x] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**
There is a robot on an m x n grid. The robot is initially located at the top-left corner (i.e., grid[0][0]). The robot tries to move to the bottom-right corner (i.e., grid[m - 1][n - 1]). The robot can only move either down or right at any point in time.

Given the two integers m and n, return the number of possible unique paths that the robot can take to reach the bottom-right corner.

* I will be adding a dynamic programming problem with different approaches.
* I will be adding comments and discussing the time complexity of the approaches.
* I have read the contribution guidelines and the style guides for the languages.
* I will also be adding the test file for the problem.

Princeton21:
@AdiChat Kindly assign me this issue.

MehakGupta1103:
please assign this issue to me @AdiChat , as i have solution ready with me in 3 differernt ways. 

1. dp memoization
2. dp recursive
3. math - nCr formula based

PR #6610, opened 2022-10-13
Added Unique Paths DP problem solution

Princeton21:
**Fixes issue:**
Closes issue #6583 

**Changes:**

* Added the important DP problem.
* It included multiple methods of solving the problem along with the various time complexities


Princeton21:
@AdiChat should I add tests for this problem in the test folder

AdiChat (approved):

PR #6628, opened 2022-10-16
Create Sorting_algos

leoperd:
**Fixes issue:**

**Changes:**


leoperd:
I have written 3  sorting algorithms with switch case option.

AdiChat (commented):
Add correct file extension and place in correct directory.

PR #6637, opened 2022-10-18
Create Print Diagonally.cpp

Sneham102:
**Fixes issue:**

**Changes:**


AdiChat (commented):
Place in correct directory.

PR #6640, opened 2022-10-19
Create Paths in Matrix Whose Sum Is Divisible by K

jatinbansal2003:
please accept for hacktoberfest 2022


AdiChat (commented):
Place in correct directory and use correct file extension.

PR #6683, opened 2022-10-29
Create collision_between_rectangles

0shuvo0:
**Changes:**
Added an algorithm to check collision between 2 rectangles


AdiChat (commented):
Make sure your code runs and add file extension.

PR #6687, opened 2022-10-30
Create Maximum sum increasing.cpp subsequence

arjun-rai912:
**Fixes issue:**

**Changes:**

### DIFF
--- a/code/Maximum sum increasing.cpp subsequence
+++ b/code/Maximum sum increasing.cpp subsequence
@@ -1,0 +1,47 @@
+/* Dynamic Programming implementation
+of Maximum Sum Increasing Subsequence
+(MSIS) problem */
+#include <bits/stdc++.h>
+using namespace std;
+
+/* maxSumIS() returns the maximum
+sum of increasing subsequence
+in arr[] of size n */
+int maxSumIS(int arr[], int n)
+{
+	int i, j, max = 0;
+	int msis[n];
+
+	/* Initialize msis values
+	for all indexes */
+	for ( i = 0; i < n; i++ )
+		msis[i] = arr[i];
+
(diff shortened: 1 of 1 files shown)


AdiChat (commented):
Place in correct location with file extension.

code/Maximum sum increasing.cpp subsequence
```diff
@@ -0,0 +1,47 @@
...
+sum of increasing subsequence
+in arr[] of size n */
+int maxSumIS(int arr[], int n)
+{
+	int i, j, max = 0;
+	int msis[n];
+
+	/* Initialize msis values
+	for all indexes */
+	for ( i = 0; i < n; i++ )
```
Md-Shamim-Ahmmed, line 17: hey can we use this line of code here 
```copy (arr, arr + n, msis);``` 
instead of  
```
for ( i = 0; i < n; i++ )
    msis[i] = arr[i];
```
Which reduce the lines of code?

File: code/Maximum sum increasing.cpp subsequence [hunk 2]
```diff
@@ -0,0 +1,47 @@
...
+	for ( i = 1; i < n; i++ )
+		for ( j = 0; j < i; j++ )
+			if (arr[i] > arr[j] &&
+				msis[i] < msis[j] + arr[i])
+				msis[i] = msis[j] + arr[i];
+
+	/* Pick maximum of
+	all msis values */
+	for ( i = 0; i < n; i++ )
+		if ( max < msis[i] )
```
Md-Shamim-Ahmmed, line 31: it is same as ```max(max, msis[i]);```
so why not use this simple function of code to reduce lines of code?

PR #6714, opened 2023-10-01
Added "Find shortest safe route in a path with landmines" Question

MAX-786:
**Fixes issue:**

Added "Find shortest safe route in a path with landmines" question from gfg and its solution.

**Changes:**

### DIFF
--- a/code/backtracking/src/find_safe_path/find_safe_path.java
+++ b/code/backtracking/src/find_safe_path/find_safe_path.java
@@ -1,0 +1,198 @@
+// Java program to find shortest safe Route
+// in the matrix with landmines
+import java.util.Arrays;
+
+class GFG{
+
+static final int R = 12;
+static final int C = 10;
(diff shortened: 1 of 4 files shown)


code/backtracking/src/find_safe_path/find_safe_path.java
```diff
@@ -0,0 +1,198 @@
...
+		{ 0, 1, 1, 1, 1, 0, 1, 1, 1, 1 },
+		{ 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 },
+		{ 1, 1, 1, 0, 1, 1, 1, 1, 1, 1 } };
+
+	// Find shortest path
+	findShortestPath(mat);
+}
+}
+
+// This code is contributed by sanjeev2552
```
AdiChat, line 198: You cannot contribute code implemented by others.

Issue #6664, opened 2022-10-25
Added new css padding example code and image.[hactoberfest]


Shivani1917:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**
Example code for padding css. 
Image is attached for the preview.

![code html](https://user-images.githubusercontent.com/79411680/197822849-2a904db1-b785-46bf-b27b-5ec0d2cfab18.png)

resolves #6663

Subhadip956425:
In this example, the CSS padding property is used to add space around the content of a div element. The padding property can take four values: top, right, bottom, and left. In this case, the padding is set to 20 pixels for all four sides for the container div.

The unique-padding div has a different padding value of 10px 20px 30px 40px. This means that the top padding is 10 pixels, the right padding is 20 pixels, the bottom padding is 30 pixels, and the left padding is 40 pixels.
![Screenshot 2023-10-03 204415](https://github.com/OpenGenus/cosmos/assets/122253446/4b26fe5f-7283-4616-8a0b-29f9d1a27ce9)

PR #6724, opened 2023-10-03
Create Added _new _css _padding.html

Subhadip956425:
solved issue #6664 

In this example, the CSS padding property is used to add space around the content of a div element. The padding property can take four values: top, right, bottom, and left. In this case, the padding is set to 20 pixels for all four sides for the container div.

The unique-padding div has a different padding value of 10px 20px 30px 40px. This means that the top padding is 10 pixels, the right padding is 20 pixels, the bottom padding is 30 pixels, and the left padding is 40 pixels.

**Fixes issue:**

**Changes:**

### DIFF
--- a/Added _new _css _padding.html
+++ b/Added _new _css _padding.html
@@ -1,0 +1,28 @@
+html
+<!DOCTYPE html>
+<html>
+<head>
+<style>
+.container {
+ background-color: lightblue;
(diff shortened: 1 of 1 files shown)


AdiChat (commented):
Add code in correct directory.

PR #6730, opened 2023-10-05
Update linked_list_palindrome.cpp

jvishad:
**Fixes issue:**
Reduced the additional space complexity of the code.

**Changes:**
Changed the vector approach to the linked list approach which doesn't use additional vector which saves our extra space.


code/data_structures/src/Linked_List/linked_list_palindrome.cpp
```diff
@@ -1,24 +1,32 @@
...
-        int e = arr.size() -1;
-        while(s<=e){
-            if(arr[s] != arr[e]){
-                return 0;
-            }
-            s++;
-            e--;
-        }
-        return 1;
+bool isPalindrome(Node* head) {
```
AdiChat, line 1: Do mention if you are making a specific improvement.

PR #6748, opened 2023-10-12
Added 2 Sum Problem

ashmit2704:
Updated new problem to the sub repo of leetcode in online challenges


AdiChat (commented):
Add complete working code.

PR #6754, opened 2023-10-18
Create Added _new _css _padding.html

Subhadip956425:
fix issue #6724

**Fixes issue:**
#6724 

**Changes:**


AdiChat (commented):
Place code in correct directory with correct filename.

Issue #6785, opened 2024-08-07
Lowest Common Ancestor Algorithms


bersubi:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**

The lowest common ancestor of two nodes of a rooted tree is the lowest node
whose subtree contains both the nodes. A typical problem is to efficiently process
queries that ask to find the lowest common ancestor of two nodes. I want to implement two solutions for that problem. One with binary lifting and one with sparse table in C++.

PR #6786, opened 2024-08-08
Added C++ code for lowest common ancestor problem solution using binary lifting

berkay-top:
**Fixes issue:**

#6785

**Changes:**

I have added Lowest Common Ancestor problem solution with binary lifting in C++.

### DIFF
--- a/code/graph_algorithms/src/lowest_common_ancestor/lowest_common_ancestor.cpp
+++ b/code/graph_algorithms/src/lowest_common_ancestor/lowest_common_ancestor.cpp
@@ -1,0 +1,82 @@
+#include <iostream>
+#include <vector>
+#include <array>
+using namespace std;
+
+// Part of Cosmos by OpenGenus Foundation
+
+// Problem Statement
+/*
+ * Given a tree with n nodes
+ * Your task is to process q queries of the following form:
+ * who is the lowest common ancestor of nodes a and b in the tree?
+ *
+ * The lowest common ancestor of two nodes of a rooted tree is the lowest node
+ * whose subtree contains both the nodes.
+ */
+
+void dfs(int k, int p, vector<int> &depth, vector<vector<int>> &adj, vector<array<int, 20>> &ancestor)
+{
+    ancestor[k][0] = p;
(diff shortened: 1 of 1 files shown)


code/graph_algorithms/src/lowest_common_ancestor/lowest_common_ancestor.cpp
```diff
@@ -0,0 +1,82 @@
...
+ * Your task is to process q queries of the following form:
+ * who is the lowest common ancestor of nodes a and b in the tree?
+ *
+ * The lowest common ancestor of two nodes of a rooted tree is the lowest node
+ * whose subtree contains both the nodes.
+ */
+
+void dfs(int k, int p, vector<int> &depth, vector<vector<int>> &adj, vector<array<int, 20>> &ancestor)
+{
+    ancestor[k][0] = p;
```
gargantuadev, line 20: Maybe I'm wrong, but I think we are not considering here the case in which k is the root node.

I would do:

`if (p == 0) {
        depth[k] = 0; // Root node
        ancestor[k][0] = k; // Root node's ancestor is itself
    } else {
        depth[k] = depth[p] + 1;
        ancestor[k][0] = p;
    }`
    

What do you think?

gargantuadev (changes requested):
I wrote the comment about the changes, tell me what you think about it!

code/graph_algorithms/src/lowest_common_ancestor/lowest_common_ancestor.cpp
berkay-top: I don't think it is necessary because the default values for arrays are already zero.

AdiChat (approved):

code/graph_algorithms/src/lowest_common_ancestor/lowest_common_ancestor.cpp
gargantuadev: Yes completely agree, I missed the implicit initialization, nice job by the way!

gargantuadev (commented):
Approved on my side!

Issue #6824, opened 2024-11-23
Rock-paper-scissors algorithm results in lower case input string error


Poly-Etilen:
**This is a(n):**
- [ ] New algorithm
- [x] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**
When the rock-paper-scissors algorithm was executed, an output of "Player 2 wins" occurred when input was received and tied to a lowercase string. It found that the code did not work well when receiving a string input in lowercase in the existing algorithm. Therefore, a code to convert the string to capital letters was added when the string was input in lowercase. In addition, the code was modified to receive the output of "Invalid choice, please choose ROCK, PAPER, or SCISSORS" when another string was input.

![코드 출력 에러](https://github.com/user-attachments/assets/bcbdcde4-f1ea-4416-91e7-e8daa50527d8)

Poly-Etilen:
Please check the code changes and leave a review

suvendukungfu:
please assign this issue

PR #6825, opened 2024-11-23
changed rock_paper_scissors

Poly-Etilen:
**Fixes issue:**

**Changes:**

When the rock-paper-scissors algorithm was executed, an output of "Player 2 wins" occurred when input was received and tied to a lowercase string. It found that the code did not work well when receiving a string input in lowercase in the existing algorithm. Therefore, a code to convert the string to capital letters was added when the string was input in lowercase. In addition, the code was modified to receive the output of "Invalid choice, please choose ROCK, PAPER, or SCISSORS" when another string was input.


Poly-Etilen:
Please check "Rock-paper-scissors algorithm results in lower case input string error #6824"

Issue #6826, opened 2024-11-26
Up-Down game with Python.


guswnsim:
**This is a(n):**

- [x] New algorithm
- [ ] Update to an existing algorithm
- [ ] Error
- [ ] Proposal to the Repository

**Details:**
I tried implementing an up-down game with Python.
Up-down games are games that guess what number is randomly determined within a range designated by the user.
The user enters the minimum and maximum values to specify the range. One random number is determined within the range designated by the user by the program. The user enters the number estimated to be the correct answer once. It is determined that the input number is larger or smaller than the correct answer so that the correct answer can be inferred. When the user enters the correct answer, the game is over.

suvendukungfu:
please assign this issue

PR #6827, opened 2024-11-26
Add UP-DOWN-GAME

guswnsim:
##Please refer to Up-Down game with Python. #6826, which is on the issue list.

##Add algorithm

When you run an up-down game, you'll see the game start screen like the first picture you see
image
Enter the minimum value
image
If you enter the maximum value
image
Numbers are randomly selected within the range I specified and the game begins.
image
If the random value is larger than the value I entered, a picture of the thumb pointing up appears and outputs that a small value has been entered in the output window.
image
If the random value is smaller than the value I entered, a picture of the thumb pointing down appears and a large value is entered in the output window.
image
Due to the nature of the game, as the range increases, the playing time increases, so if it comes within the error range of 10, it is set to give a hint, and if the correct answer is completed after a few attempts, the game is over.
image

### DIFF
--- a/UP-DOWN-GAME.py
+++ b/UP-DOWN-GAME.py
@@ -1,0 +1,131 @@
+import time
+import random
+import pygame
+import sys
+from pygame.locals import *
+
+# Pygame 초기화
+pygame.init()
+
+# 창 크기 및 엄지손가락 이미지 로드
+screen = pygame.display.set_mode((800, 600))
+pygame.display.set_caption("UP&DOWN GAME")
+thumb_up = pygame.image.load(r"C:\Users\OK\Desktop\현준\엄지척.PNG")  # 엄지손가락을 향해 치켜세우는 이미지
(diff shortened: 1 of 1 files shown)


guswnsim:
> ##Please refer to Up-Down game with Python. #6826, which is on the issue list.
> 
> ##Add algorithm
> 
> When you run an up-down game, you'll see the game start screen like the first picture you see image Enter the minimum value image If you enter the maximum value image Numbers are randomly selected within the range I specified and the game begins. image If the random value is larger than the value I entered, a picture of the thumb pointing up appears and outputs that a small value has been entered in the output window. image If the random value is smaller than the value I entered, a picture of the thumb pointing down appears and a large value is entered in the output window. image Due to the nature of the game, as the range increases, the playing time increases, so if it comes within the error range of 10, it is set to give a hint, and if the correct answer is completed after a few attempts, the game is over. image

